PR #1, opened 2016-04-26
added command to enable the newly required module

sammarcus:


maxiloc:
Sorry for the long delay.
Thanks for contributing

Issue #3, opened 2016-06-08
Producthelper class does not exists


redboxmarcins:
Hi 
 Thanks for providing extension but we found it is not working due to invalid class name in many of the files.

Tested on PHP 7.0.5

Producthelper should be replaced by ProductHelper

Files:

Helper/ConfigHelper.php
Helper/Data.php
Model/Indexer/Category.php
Model/Indexer/Product.php
Model/Source/AbstractTable.php
Model/Source/Facets.php

PR #4, opened 2016-06-08
Producthelper class does not exists #3

redboxmarcins:
Fixed class name camelCasing


maxiloc:
Thanks for contributing

PR #9, opened 2016-06-28
Enable mod_rewrite to unite CLI and web generated URLs. 

JanPetr:
Alse remove SIDs from URLs.

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -542,7 +542,7 @@ class ProductHelper extends BaseHelper
         $customData = array(
             'objectID'          => $product->getId(),
             'name'              => $product->getName(),
-            'url'               => $product->getProductUrl(),
+            'url'               => $product->getProductUrl(false),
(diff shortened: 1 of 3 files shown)


Helper/Entity/ProductHelper.php
```diff
@@ -542,7 +542,7 @@ public function getObject(\Magento\Catalog\Model\Product $product)
         $customData = array(
             'objectID'          => $product->getId(),
             'name'              => $product->getName(),
-            'url'               => $product->getProductUrl(),
+            'url'               => $product->getProductUrl(false),
```
maxiloc: What does the false do ?

JanPetr: Removes SID from URL. Method signature:
```php
/**
     * Retrieve Product URL
     *
     * @param  bool $useSid
     * @return string
     */
    public function getProductUrl($useSid = null) { ...
```

maxiloc:
👍

PR #88, opened 2016-10-17
fixed old doc link

sammarcus:


JanPetr:
Good catch, thank you @sammarcus !

PR #97, opened 2016-11-01
Fix config helper to use ProductMetadataInterface to get the correct dependecy

tkotosz:
If `Magento\Framework\App\ProductMetadata` used directly then the `$this->productMetadata->getEdition();` will always return `Community` (see https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/App/ProductMetadata.php#L24).

If `Magento\Framework\App\ProductMetadataInterface` used instead then:
- in Community Edition the object manager will return the `Magento\Framework\App\ProductMetadata` and the `getEdition` will return `Community`
- in Enterprise Edition the object manager will return the `Magento\Enterprise\Model\ProductMetadata` (because of this preference: `<preference for="Magento\Framework\App\ProductMetadataInterface" type="Magento\Enterprise\Model\ProductMetadata"/>`) and the `getEdition` method of that class will return `Enterprise`

### DIFF
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -89,7 +89,7 @@ class ConfigHelper
                                 DirCurrency $dirCurrency,
                                 DirectoryList $directoryList,
                                 Magento\Framework\Module\ResourceInterface $moduleResource,
-                                Magento\Framework\App\ProductMetadata $productMetadata)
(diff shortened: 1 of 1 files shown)


JanPetr:
Thanks for the patch @tkotosz 🎉
Happy to merge, thank you!

tkotosz:
wow, that was quick, thanks for the quick merge :+1:

Issue #99, opened 2016-11-04
"Show categories that are not included in the navigation menu" not taking into account during reindex


rafal-kos:
Looks like option "Show categories that are not included in the navigation menu" is not taking into account during reindexation process of categories. 

In method _getCategoryCollectionQuery_ there is filter on _incude_in_menu_ which always is set to 1

```
$categories
            ->addPathFilter($storeRootCategoryPath)
            ->addNameToResult()
            ->addUrlRewriteToResult()
            ->addIsActiveFilter()
            ->setStoreId($storeId)
            ->addAttributeToFilter('include_in_menu', '1')
            ->addAttributeToSelect(array_merge(['name'], $additionalAttr))
            ->addFieldToFilter('level', ['gt' => 1]);
```

Proposed solution :

```
$categories
            ->addPathFilter($storeRootCategoryPath)
            ->addNameToResult()
            ->addUrlRewriteToResult()
            ->addIsActiveFilter()
            ->setStoreId($storeId)
            ->addAttributeToSelect(array_merge(['name'], $additionalAttr))
            ->addFieldToFilter('level', ['gt' => 1]);

        if (!$this->config->showCatsNotIncludedInNavigation()) {
            $categories->addAttributeToFilter('include_in_menu', 1);
        }
```

JanPetr:
Hello @rafal-kos,
thanks for the report. We will fix it soon.

PR #100, opened 2016-11-04
Fix for issue #99 

rafal-kos:
#99 : Handle properly option "Show categories that are not included in the navigation menu" for category reindexation


JanPetr:
Thanks for the fix @rafal-kos!

PR #106, opened 2016-11-15
Always process at least one queue job

JanPetr:
Fixes the issue, when queue settings are set high, jobs are created and then put back to low. Currently it won't process any jobs and jobs are stuck in DB table.

This PR will ensure that at least one job will be processed every time queue runner runs no matter what data size of the job is.


maxiloc:
👍

PR #134, opened 2016-12-26
Enable SEO request by default

### LINKED ISSUE #128: Disabling instantsearch do not fallback on algolia backend search by default
In order for this to work you need to enable SEO request which is disabled by default.

JanPetr:
Fixes #128


maxiloc:
👍

PR #135, opened 2016-12-26
Index correct CMS pages for specific store views

### LINKED ISSUE #119: CMS pages being indexed across all stores
We have a multi site installation and we have just noticed that when a customer searches for returns they are returned with three "Returns" links. I have double checked the Magento 2 admin and each Returns page is only set to one store view. Looking at the index for each store view on Algolia dashboard it seems there are multiple pages being indexed across all stores. Some have two, some have three pages.

screen shot 2016-11-29 at 18 04 02

JanPetr:
Fixes #119


maxiloc:
👍

PR #143, opened 2017-01-10
Fix ratings styles - show the products' rating on the IS page

### LINKED ISSUE #129: Rating summary missing in results
The review rating summary is missing from the category and search result pages. I have added the `rating_summary` to the attributes configuration in Magento and I have reindexed the data but this is not shown in the frontend. The `rating_summary` has been indexed as seen here - 

screen shot 2016-12-16 at 08 23 43

And I can see where it is suppose to be used in the frontend `hit` template - 

screen shot 2016-12-16 at 08 23 58

I am not sure if this is an issue in the retrieval of the object or in the JavaScript itself. It's worth noting that I don't see any products in your demo site that has the `rating_summary` stars. I do see it on your Magento 1 demo though.

Is this a known issue?

JanPetr:
Fixes #129

### DIFF
--- a/view/frontend/templates/instant/hit.phtml
+++ b/view/frontend/templates/instant/hit.phtml
@@ -23,11 +23,15 @@ $priceKey = $block->getPriceKey();
                         </h3>
                         <div class="ratings">
                             <div class="result-sub-content">
-                                <div class="ratings">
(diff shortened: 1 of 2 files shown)


maxiloc:
👍

PR #145, opened 2017-01-12
Index correct prices with applied catalog price rules

### LINKED ISSUE #122: Prices affected by catalog rule are not working in result page
Prices affected by catalog rule are not working in result page, can you please fix this issue?

JanPetr:
Fixes #122

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -117,7 +117,7 @@ class ProductHelper extends BaseHelper
         }
 
         /* @noinspection PhpUnnecessaryFullyQualifiedNameInspection */
-        $products = $products->addFinalPrice()
+        $products = $products
             ->addAttributeToSelect('special_from_date')
             ->addAttributeToSelect('special_to_date')
             ->addAttributeToSelect('visibility')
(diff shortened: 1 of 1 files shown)


maxiloc:
Let's test that without addFinalPrice which does the join with the price index table

JanPetr:
Having `addFinalPrice()` doesn't make any difference regarding either functionality or performance - so I omitted it.

The new way of getting prices is little bit slower, but that's the only way I found on how to get the correct price with applied catalog price rules.

Thoughts @maxiloc?

maxiloc:
Good for me

maxiloc:
Just need to check if we can do the same on M1 because it would allow us to remove the join issue

JanPetr:
I'm on it now

phupt:
Hi @JanPetr 

When i update the code and reindex, all product price is zero. Do you know why?

Regards

JanPetr:
Hello @phupt,
to be honest I don't know the exact reason.
My guess it's because of [this change](https://github.com/algolia/algoliasearch-magento-2/pull/145/files#diff-059444697db366f04fa1b6f3b6578addL120).
Could you put back the `->addFinalPrice()` and reindex again?

Thanks!

phupt:
Hi @JanPetr 

Yes, i put pack the ->addFinalPrice() again and it works. Thanks for your help.

Regards

JanPetr:
Hey @phupt,

thanks for getting back to me and for the feedback. We'll add it back.

Sorry for the inconvenience

PR #159, opened 2017-02-23
fix(amd): fix algolia admin bundle

### LINKED ISSUE #158: Move algoliaAdminBundle to requirejs
Magento 2 uses requirejs so we need to use that for our bundle too.

This means everywhere you want to use `algoliaAdminBundle`, we need to do:

```js
requirejs(['algoliaAdminBundle'], function(algoliaBundle) {
...
});
```

I am updating the extension bundle to have it ready for this to work and will open a pull request for the other part. Right now it fails when inserted because of: http://requirejs.org/docs/errors.html#mismatch and https://webpack.js.org/configuration/output/#output-umdnameddefine.

Linked to https://github.com/algolia/algoliasearch-magento-2/issues/147.

I do not get exactly why we did not get this error before or how other clients have done it, may be missing something.

vvo:
fixes #158


vvo:
might have to update the dockerfile dev file, not sure but contains algoliaAdminBundle. token

vvo:
cc @JanPetr

JanPetr:
Thanks @vvo, you're the best! ❤️

PR #172, opened 2017-03-07
When customer groups are enabled, send missing replicas and indexers for non-price sort attributes

unicoder88:
When I enable customer groups, sorting doesn't work for all sort attributes other then price. If I delete all indexes in Algolia and reindex Magento, I see that indexes for them are even not created. This is because of missing 'else' cases.


JanPetr:
Hello @unicoder88,

thanks a bunch for your contribution. I refactored the code a little bit more and merged the replicas' logic into a ConfigHelper in #217.
The PR is merged, so when you use `master` branch, the fix should be there. 

I'm closing this PR, but it was very helpful for me. If you have any other issues or ideas, feel free to open a PR or issue. I'll be happy to see it. 💯

PR #174, opened 2017-03-10
Fix secure URLs

### LINKED ISSUE #149: The website is set to use Secure URLs, but the extension indexes URLs with http:// and no https://

JanPetr:
Fixes #149


dhduc:
Hi @JanPetr 
Which version of this patch is applied?

JanPetr:
Should be in 1.0.9 @dhduc

PR #208, opened 2017-04-05
Queue processing optimalization

JanPetr:


Model/Queue.php
```diff
@@ -66,56 +87,11 @@ public function runCron()
...
 
-        // Reserve all new jobs since last run
-        $this->db->query("UPDATE {$this->db->quoteIdentifier($this->table, true)} SET pid = " . $pid . ' WHERE job_id >= ' . $first_id . " AND job_id <= $last_id");
-
-        foreach ($jobs as &$job) {
-            $job['data'] = json_decode($job['data'], true);
-        }
-
-        $jobs = $this->sortAndMergeJob($jobs);
+        $jobs = $this->getJobs($maxJobs, $pid);
```
julienbourdeau: Maybe we should get the job outside this loop (loop because run is called recursively). Jobs added in the queue while running a full reindex won't interfere. What do you think?

JanPetr: The reason it's here and why it's called recursively is to have all jobs done (even those which are added during the `run`)

Model/Queue.php
```diff
@@ -128,82 +104,199 @@ public function run($limit)
                 call_user_func_array([$model, $method], $data);
             } catch (\Exception $e) {
                 // Increment retries and log error information
```
JanPetr: Bad comment, retries are not used anymore

PR #274, opened 2017-06-29
Fix configurable price that breaks special price

unicoder88:
For configurable product, $product->getPrice() returns 0, and so any special prices are not added because of the check
```$specialPrice < $customData[$field][$currencyCode]['default']```

Further in the code configurable price falls back to $min price
```
if ($customData[$field][$currencyCode]['default'] == 0) {
    $customData[$field][$currencyCode]['default'] = $min;
```
but special price is already missed. It looks like more relevant approach is to use priceInfo structure.

Please review. Looking forward to your comments,
Roman

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -398,7 +398,7 @@ class ProductHelper extends BaseHelper
 
                 $specialPrices = [];
                 $specialPrices[] = (double) $this->rule->getRulePrice(new \DateTime(), $store->getWebsiteId(), 0, $product->getId()); // The price with applied catalog rules
-                $specialPrices[] = (double) $product->getFinalPrice(); // The product's special price
+                $specialPrices[] = (double) $product->getPriceInfo()->getPrice('final_price')->getValue(); // The product's special price
 
                 $specialPrices = array_filter($specialPrices, function($price) {
                     return $price > 0;
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -386,7 +386,7 @@ class ProductHelper extends BaseHelper
             foreach ($currencies as $currencyCode) {
                 $customData[$field][$currencyCode] = [];
 
-                $price = $product->getPrice();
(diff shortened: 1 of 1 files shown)


JanPetr:
Hello @unicoder88,

sorry for the late answer. Could you please describe the use-case which causes you issues and indexes wrong price?
Thanks a lot!

unicoder88:
Hi!

Sorry I've had a vacation. When making a description for the case, I came across a second issue with final price, and so fixed it.

To describe issue In two words, when we applied catalog rule to configurable product, it wasn't visible in Algolia.

Please review a video I recorded:
https://vimeo.com/228184768
Password: algolia

Thank you

JanPetr:
Hey Roman!

Thanks for the video, it's awesome!
I tested it and it indeed solved the issue. 👍 
But on the other hand it breaks multi store / currency setup :(

My multi currency setup:
screenshot 2017-08-03 14 12 04

screenshot 2017-08-03 14 14 14

And when I re-index on `master` branch I get:
screenshot 2017-08-03 13 41 53

And your fix:
screenshot 2017-08-03 13 39 21

Would you mind taking a look at this one? I'd really appreciate!! 💯 Thank you for your contributions!

yosefkurniawan:
hi, have this issue been fixed on the latest released?
i am facing the same issue.

JanPetr:
Hello @yosefkurniawan,
the issue should be solved on master branch. Can you update?

yosefkurniawan:
If the product has special price, the price format that will be displayed is like this: `$70 - $89`,
is it correct?
Also, does the changes not fix the special price on tier price as well?
when tier price is set with the special price on certain customer group, the algolia will keep displaying the normal price on the autocomplete menu.

JanPetr:
Hello @yosefkurniawan,

yes, the `70-89` is correct as it's configurable product with different prices.
The extension doesn't count with tier pricing at all, so this should be correct as well.

JanPetr:
I'm closing the PR now the issue was already solved. If you encounter any errors, feel free to re-open.

Issue #285, opened 2017-07-18
EMPTY_QUEUE=1 leads to infinite loop


zts:
The [reference documentation](https://community.algolia.com/magento/doc/m2/indexing/#general-information) mentions setting the environment variable `EMTPY_QUEUE=1` to persuade the indexer to process the entire queue in one go.

This works as documented with the Magento 1 extension, but causes an infinite loop (and eventual segfault) on Magento 2.

The problem was introduced somewhere between 1.0.8 and 1.1.0.

PR #286, opened 2017-07-18
Prevent infinite recursion when EMPTY_QUEUE=1

zts:
When the full queue is being processed, this function recurses after
handling a set of jobs - so we need to quit when we run out of jobs to
process.

This resolves GH-285

For comparison, the M1 module implements this as: https://github.com/algolia/algoliasearch-magento/blame/a0200a39403dd13e0f4820e325e772f985bcccbc/app/code/community/Algolia/Algoliasearch/Model/Queue.php#L77-L79

### DIFF
--- a/Model/Queue.php
+++ b/Model/Queue.php
@@ -93,6 +93,10 @@ class Queue
 
         $jobs = $this->getJobs($maxJobs, $pid);
 
+        if (empty($jobs)) {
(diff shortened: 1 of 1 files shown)


JanPetr:
Great catch @zts,
thanks a lot for your contribution 💯

Issue #288, opened 2017-07-21
Algolia is called on non-product page


julienbourdeau:
If you go to the register page for instance, you'll see that Algolia is still called to load products, even it they're not used. You could come up with a solution to avoid this.

![privacy_policy](https://user-images.githubusercontent.com/1525636/28515783-29e85ff8-705f-11e7-96db-b7d3dc56e605.png)

The DOM is also updated and the content is hidden by CSS.

![privacy_policy](https://user-images.githubusercontent.com/1525636/28518520-72bf0bc4-7068-11e7-987e-4951c3433596.png)

danielozano:
Do you mean that the search adapter tries to "search" on algolia? We had issues with that because there are other extensions using the Search Adapter, so we made more specific the use of algolia's search engine inside the adapter class.

If we are talking about the same problem (I'm not sure) I can share the override we made to avoid this issue.

julienbourdeau:
@danielozano I added a screenshot. I meant that the javascript calls algolia even if we don't need product. Although, your problem seems interesting.

Would you mind submitting a pull request so we can discuss it?

PR #301, opened 2017-08-09
restrict use of search adapter

danielozano:
Hi! This is not a PR to merge, just to discuss as @julienbourdeau requested me here #288

In one of our Magento 2 projects we had the case of a listing product page created by a 3rd party module (I don't want to promote any company but if you want more information you can ask me) not showing all the "expected results".

So going deeper on the code we saw that it's incompatible with algolia search. Algolia overrides the search adapter, but the module uses the algolia query in more cases than expected.

With this changes, we only use the algolia search in the cases where it's needed.

PD: It's not the best code, I think that you shouldn't merge this request because its a development hotfix. I just want to present you my situation, because I think that algolia query is being used in more situations than it should.

Thank you!

### DIFF
--- a/Adapter/Algolia.php
+++ b/Adapter/Algolia.php
@@ -117,16 +117,9 @@ class Algolia implements AdapterInterface
         $documents = [];
         $table = null;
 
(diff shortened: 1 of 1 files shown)


JanPetr:
Thanks for your work @danielozano!
I took your work, polished it and merged in #357. Thanks a lot for bringing this up! 💯

PR #311, opened 2017-09-12
Do not call API on non product page

### LINKED ISSUE #288: Algolia is called on non-product page

julienbourdeau:
Fix #288 

Today, with Instant Search activated,  Algolia is called and the DOM is updated on every page. This PR will only call algolia if you type something is the search bar or if you're on a result page (like category).

It could be nice to add some tests but I haven't really figured out how.

### DIFF
--- a/view/frontend/web/instantsearch.js
+++ b/view/frontend/web/instantsearch.js
@@ -56,6 +56,16 @@ requirejs(['algoliaBundle'], function(algoliaBundle) {
 			urlSync: {
 				useHash: true,
 				trackedParameters: ['query', 'page', 'attribute:*', 'index']
+			},
+			searchFunction: function(helper) {
+				if (helper.state.query === '' && !algoliaConfig.isSearchPage) {
+					$('.algolia-instant-replaced-content').show();
+					$('.algolia-instant-selector-results').hide();
+				} else {
+					helper.search();
+					$('.algolia-instant-replaced-content').hide();
+					$('.algolia-instant-selector-results').show();
+				}
 			}
 		};
 		


quentinfarizon:
Works for me ! Thanks a lot @julienbourdeau 
Without the fix, on my homepage, at page load before searching, the whole instant search block was pre-filled with images and filters, ...
With this fix, at page load it's empty, and hasn't generated queries nor downloaded images.

view/frontend/web/instantsearch.js
```diff
@@ -56,6 +56,16 @@ requirejs(['algoliaBundle'], function(algoliaBundle) {
 			urlSync: {
 				useHash: true,
 				trackedParameters: ['query', 'page', 'attribute:*', 'index']
+			},
+			searchFunction: function(helper) {
+				if (helper.state.query === '' && !algoliaConfig.isSearchPage) {
```
JanPetr: How it works for category pages powered by Algolia? When it loads it has an empty query and is not a search page - wouldn't it break it?

julienbourdeau: I think category page are considered search page. I worked when I tested but I will verify, just to make sure I didn't miss something.

JanPetr: Any news on that @julienbourdeau ?

julienbourdeau: Yes I confirm that `algoliaConfig.isSearchPage` is true for category pages. And only the pages that display results (not the top level page with special template).

JanPetr: Thanks @julienbourdeau!

PR #330, opened 2017-10-10
fixing price calculations

lowczarczuk:
In fact this is another way to fix the problem reported in #274


JanPetr:
Thank you for your contribution @lowczarczuk! I really appreciate it! 💯

PR #359, opened 2017-11-17
Fix special prices

lowczarczuk:
There is a problem when customer group has been specified in a catalog price rule so I fixed it up

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -382,6 +382,10 @@ class ProductHelper extends BaseHelper
 
         $groups = $this->objectManager->create('Magento\Customer\Model\ResourceModel\Group\Collection');
 
+        if (!$areCustomersGroupsEnabled) {
+            $groups->addFieldToFilter('main_table.customer_group_id', 0);
+        }
+
         foreach ($fields as $field => $withTax) {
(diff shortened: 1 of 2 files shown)


JanPetr:
Thank you for the contribution @lowczarczuk!
I'll give it few tests tomorrow and come back to you.

Helper/Entity/ProductHelper.php
```diff
@@ -380,11 +380,7 @@ protected function handlePrice(Product &$product, $subProducts, &$customData)
         $currencies = $store->getAvailableCurrencyCodes();
         $baseCurrencyCode = $store->getBaseCurrencyCode();
 
-        $groups = [];
-
-        if ($areCustomersGroupsEnabled) {
```
JanPetr: Why did you remote this condition? It doesn't need to fetch groups when they are not enabled, no?

lowczarczuk: If I leave this condition I have to duplicate foreach from 407th line in order to avoid problem with undefined `$groups` variable. I've added filter to speed up the code and keep properly defined `$groups` variable

JanPetr:
Thanks @lowczarczuk, makes sense! Everything looks good, thanks for your contribution! 💯

PR #374, opened 2017-12-08
Fix error when in_stock comes from $defaultData

azurre:
Reproduce:
Observer for `algolia_product_index_before`:
```
public function execute(\Magento\Framework\Event\Observer $observer)
    {
        /** @var \Magento\Catalog\Model\Product $product */
        $product = $observer->getProduct();
        $data = $observer->getCustomData();
        $data->setData('in_stock', 1);
        $observer->setCustomData($data);
    }
```


JanPetr:
Thanks @azurre! 💯

PR #375, opened 2017-12-12
Do not cast SKUs, names and descriptions

JanPetr:


MartinWickman:
We are having this particular problem (sku's with starting with zeros getting mangled). When are you planning to include this in the official relaase?

JanPetr:
Hello Martin,

don't have exact ETA, but I'd like to release a new stable version before end of January.

PR #380, opened 2017-12-27
Make sure to delete product in Algolia when it goes out of stock

unicoder88:
Hi again,

I noticed that when product goes out of stock, it's doesn't get updated in Algolia.
Deleting out of stock is handled in \Algolia\AlgoliaSearch\Helper\Data::getProductsRecords , but that code doesn't get executed because we filter such products prematurely.

Please review :)
Thanks


JanPetr:
Hello @unicoder88,

thank you for bringing it up! I fixed it with updated condition in #381.
Great catch, thank you for your contribution! 🚀 💯

PR #401, opened 2018-01-22
Fix buggy onclick event on iOS

JanPetr:


koerakoera:
When will this fix be published?

a-kozlov-360:
Hi, this fix doesn't work. Instead of 'touchmove' called 'touchstart' and following a link.
Tested on Chrome (devtool), iPhone 5s

PR #423, opened 2018-02-14
Update README.md

kauansoft:
Magento 2.2 needs force.

bin/magento setup:upgrade && bin/magento setup:static-content:deploy **-f**


peterjaap:
Magento 2.2 does not need force. Probably you are running your store in developer mode. If you switch to production, the `-f` flag is not necessary.

kauansoft:
Im running as "default mode".

peterjaap:
I don't know about default mode except that you should avoid running in it. Use production on production or developer on development, but never default.

PR #434, opened 2018-03-01
 New front end hook mechanism

JanPetr:


view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,54 @@
+var algolia = {
+	allowedHooks: [
+		'modifyAutocompleteSources',
```
bobylito: I'd use a name that describes a part of the lifecyle of your configuration like `beforeAutocomplete` or `beforeAutocompleteSource`. WDYT?

view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,54 @@
...
+		'modifyAutocompleteOptions',
```
bobylito: I'm not sure we need two steps.

view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,54 @@
...
+		'algoliaHookBeforeInstantsearchInit',
+		'algoliaHookBeforeWidgetInitialization',
+		'algoliaHookBeforeInstantsearchStart',
+		'algoliaHookAfterInstantsearchStart'
```
bobylito: Those are great. Be sure to document all of them with the arguments they get :)

view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,54 @@
...
+	registerHook: function (hookName, callback) {
+		if (this.allowedHooks.indexOf(hookName) === -1) {
+			throw 'Hook "' + hookName + '" cannot be defined. Please use one of ' + this.allowedHooks.join(', ');
+		}
+		
+		if (this.registeredHooks[hookName] === undefined) {
+			this.registeredHooks[hookName] = [];
+		}
+		
+		this.registeredHooks[hookName].push(callback);
```
bobylito: from 16, I'd rewrite it as:

```javascript
		if (this.registeredHooks[hookName] === undefined) {
			this.registeredHooks[hookName] = [callback];
		} else {
		        this.registeredHooks[hookName].push(callback);
		}
```

nitpicking here ;)

view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,54 @@
...
+	},
+	getRegisteredHooks: function(hookName) {
+		if (this.registeredHooks[hookName] === undefined) {
```
bobylito: It's ok to replace `if(val === undefined)` by `if(!val)`

view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,54 @@
...
+		return this.registeredHooks[hookName];
+	},
+	triggerHooks: function () {
+		var hookName = arguments[0];
+		delete arguments[0];
+		
+		var clearedArguments = this.clearArguments(arguments);
+		
+		var registeredHooks = this.getRegisteredHooks(hookName);
+		for (var i = 0; i < registeredHooks.length; i++) {
```
bobylito: I'd use a `forEach` here since you are also using filter later on.

view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,54 @@
...
+			clearedArguments[0] = registeredHooks[i].apply(this, clearedArguments);
+		}
+		
+		return clearedArguments[0];
+	},
+	clearArguments: function(arguments) {
+		var clearArguments = Array.prototype.slice.call(arguments);
```
bobylito: I wouldn't do the transformation from arguments to array here. I'd put it in the calling function and would only use arrays internally.

view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,54 @@
...
+	},
+	getRegisteredHooks: function(hookName) {
+		if (this.registeredHooks[hookName] === undefined) {
+			return [];
+		}
+		
+		return this.registeredHooks[hookName];
+	},
+	triggerHooks: function () {
+		var hookName = arguments[0];
```
bobylito: I'd do the transformation from arguments to array here. And I would use a splice to remove the first element https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

view/frontend/web/internals/common.js
bobylito: I would check if the `hookname` is defined here, as well.

view/frontend/web/internals/common.js
```diff
@@ -1,3 +1,47 @@
...
+		}
+		
+		return this.registeredHooks[hookName];
+	},
+	triggerHooks: function () {
+		var hookName = arguments[0],
+			clearedArguments = Array.prototype.slice.call(arguments, 1);
+		
+		this.getRegisteredHooks(hookName).forEach(function(callback) {
+			clearedArguments[0] = callback.apply(this, clearedArguments);
```
bobylito: can you explain the usage of clearedArguments here? What's the contract of the hook?

bobylito: Why does the hook need to have access to the this?

PR #443, opened 2018-03-14
Subscribe to stock item change and reindex product

unicoder88:
**Summary**

When customers purchase last product item in inventory on frontend, the product becomes out of stock. However, Algolia fails to see this change and doesn't reindex.

My suggestion is to subscribe to stock_item change (both on save and by schedule). This seems to remedy situation.

The only drawback seems to be that sometimes product might reindex twice, after saving product and then after saving stock item. But this should be the case only when algolia index is in save mode and queue mode is disabled. Otherwise duplicate product IDs should be skipped.

Please review :)

### DIFF
--- a/Model/Indexer/StockItemObserver.php
+++ b/Model/Indexer/StockItemObserver.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Model\Indexer;
+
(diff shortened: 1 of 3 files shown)


JanPetr:
Thanks @unicoder88! Looks good and works well! 🚀

PR #447, opened 2018-03-19
Remove disabled products from AlgoliaSearch

dmitriy-convert:
Hello! :)

I had some problems with remove disabled products from AlgoliaSearch.
After long research I thought that problem in function `\Algolia\AlgoliaSearch\Helper\Entity\ProductHelper::getProductCollectionQuery();`

Only enabled products are selected in this function:
`->addAttributeToFilter('status', \Magento\Catalog\Model\Product\Attribute\Source\Status::STATUS_ENABLED);`

But further in the code there is filtration:
`\Algolia\AlgoliaSearch\Helper\Data::getProductsRecords()`
Condition: `$product->getStatus() == Status::STATUS_DISABLED` (string 430)
Why this check, if there are no disabled products?

And I added string `->addAttributeToSelect('status')` in function `getProductCollectionQuery()`, because default Magento method `getStatus()` returned true all time, if status value non exists in the product. Example Magento code from `\Magento\Catalog\Model\Product`:
```
public function getStatus()
{
    $status = $this->_getData(self::STATUS);
    return $status !== null ? $status : \Magento\Catalog\Model\Product\Attribute\Source\Status::STATUS_ENABLED;
}
```

Please consider my offer. Thanks!


JanPetr:
Thank you, @dmitriy-convert! Looks good now, I'll merge 💯 Thanks a lot for your help 🚀

dmitriy-convert:
No problem. Thanks! :)

PR #448, opened 2018-03-22
Make product thumbnail configurable

jahvi:
**Summary**

As it is now the product thumbnail image is hardcoded to 75x75 so if you want to display larger images on the autocomplete list the only choice is to use CSS but this will make images look blurry.

It would make sense to use the `product_thumbnail_image` image size here instead which is already 75x75 by default and, if anyone needs a different custom size, it can be overwritten using the `etc/view.xml` file.

**Result**

Notice image quality is increased after the image size change.

| Before | After |
| --- | --- |
| screen shot 2018-03-22 at 10 42 18 | screen shot 2018-03-22 at 10 40 28 |


JanPetr:
Makes sense, thank you @jahvi 💯 🚀

PR #451, opened 2018-03-26
Resolves issue stemming from skipping too big record attempting to ca…

brucemead:
…st false rather than skipping/continuing.

**Summary**

We've noticed that during reindexing when you skip records which are too big you set `$object` to `false` but then attempt to cast the `$object` which will `foreach (false...` which obviously throws an exception and prevents an entire batch from indexing.

**Result**

I'd hope the result should be clear from the basic code change, but this will skip casting of a record which isn't going to be indexed anyways.


JanPetr:
Great catch, @brucemead! Thanks for that 💯

PR #455, opened 2018-03-28
Added check to see if product is in stock to calculate correct min/ma…

### LINKED ISSUE #442: Out of stock subproduct may cause min price wrong
bug

min price is using out of stock product price
has been set out of stock option to no in algolia settings and inventory settings
use min price of in stock products

Magento 2.2.3

peterjaap:
…x prices, closes #442

**Summary**

**Result**


JanPetr:
The issue was fixed by #462, thanks a lot for the initiative @peterjaap! 👍

PR #460, opened 2018-04-03
Resolve issue Add to cart on instant search page - first time.

hugoruivo:
This resolves an issue with Add to cart on instant search page - first time.
After login the window reloads and the cookie `form_key` gets updated but Algolia keeps old value until another window reload occurs.
This fix just compares Algolia `form_key` value with the cookie one, if they differ it updates Algolia to the cookie one.


JanPetr:
Thanks for the PR @hugoruivo 💯

Issue #426, opened 2018-02-22
Product indexer + queue extremely slow


peterjaap:
When the queue/cron is disabled, the product indexer gets fed the `$productIds` array, which contains **all** product ID's. We have a shop with 100.000+ SKU's of which only about 150 products are actually visible on the frontend.

When I look in Algolia, I only see those ~150 products. So my guess is that the rest is ignored. Running a full reindex (for good reasons ;)) takes about 6 hours. Filtering the `$productIds` array on only products that have visibility set to > 1 should then be taking into the indexer, considerably cutting down the reindexing time.

Am I missing something here in my assumptions? Maybe something to do with the configurable variations?

peterjaap:
Ok that didn't really have any effect. When I log the `$productIds` variable in the [execute](https://github.com/algolia/algoliasearch-magento-2/blob/master/Model/Indexer/Product.php#L46) method, I see it printed out 4 times when I run it through `php bin/magento ind:reindex algolia_runner`. Why is that?

peterjaap:
Ok so the problem isn't with the visibility. The indexer is just generally slow (at least here). If I want to force reindex all my ~150 products, it takes about 6 hours.

But I just found out that when I push all ids and let the indexer chunk it into chunks of 30 items each (or whatever), it's slow. But when I 'chunk' it into chunks of 1 item, each item takes about 10 seconds. So 150 products * 4 storeviews * 10 seconds = 6000 seconds which is about 1.5 hours. Better than 6 hours, but still too slow for what is basically just a simple JSON object....

peterjaap:
Actually, it is also very slow when I insert them with chunk 1 into the database and then let the runner run. I now insert them 1 by 1 into the database and run the queue directly after inserting one so the queue runner only has to process one item. I feel like there is a memory leak or something somewhere.

peterjaap:
In case anyone is interested in my fugly-hack-until-it-works-and-dont-refactor script;

```
<?php

$storeviewIds = [1, 2];
$productIds = [1,2,3,4];

foreach ($storeviewIds as $storeviewId) {
    foreach ($productIds as $productId) {
[...]

PR #463, opened 2018-04-04
  Fix configurable product children attributes indexing

alexey-rakitin:
**Summary**
For configurable product with sub-products, index should contain an array of attribute values, taken from sub-products, for example, color: ['Red', 'Blue', 'Yellow'], size: ['L', 'XL'], but only one attribute value was present in index color: ['Red']  size: ['L'].
As a result, instant search page facets worked incorrectly.
That happened because $values variable in ProductHelper:addNullValue was overwritten on each foreach iteration of sub-products.
I used array_merge to solve that (getValues returns array with numeric keys, that's why array_merge is OK).

**Result**
Product index for configurable product works as intended.

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -678,7 +678,7 @@ class ProductHelper
                 /** @var string|array $valueText */
                 $valueText = $subProduct->getAttributeText($attribute['attribute']);
 
-                $values = $this->getValues($valueText, $subProduct, $attributeResource);
+                $values = array_merge($values, $this->getValues($valueText, $subProduct, $attributeResource));
                 $subProductImages = $this->addSubProductImage($subProductImages, $attribute, $subProduct, $valueText);
             }
         }


JanPetr:
Hello @alexey-rakitin,

thank you for your contribution!
Just 10 minutes before you I pushed this fix with some additional optimization - https://github.com/algolia/algoliasearch-magento-2/pull/462/

Can you give it a try and see if everything works as expected? :) Thanks a bunch!

alexey-rakitin:
Hello @JanPetr,

A little late :)))
Ok, will check.

peterjaap:
This is the same issue I referred to here; https://github.com/algolia/algoliasearch-magento-2/issues/426#issuecomment-376623128

JanPetr:
I merged the previous PR (#462), which contained the same fix. Thanks a lot for your contribution @alexey-rakitin. Feel free to [grab some swag](https://docs.google.com/a/algolia.com/forms/d/e/1FAIpQLScFHcqkzw8QoS-XkU1zZ4eOg77K2t88qIXrqOC-828ahyiCnQ/viewform)! 🚀

Issue #318, opened 2017-09-27
Support Magento 2.2


emiliolodigiani:
trying to install in magento 2.2 composer throws an error:

```
algolia/algoliasearch-magento-2 1.2.1 requires magento/framework ~100.0
```

Magento 2.2 ships with magento/framework v101

JanPetr:
Thanks for the report Emilio! 💯 
Will do tests on 2.2 and update the composer.json. I'll let you know when it's done.

JanPetr:
Hello Emilio, when you install the extension from `dev-master`, it should work as expected.

`composer require algolia/algoliasearch-magento-2:dev-master`

avoelkl:
Just for completeness: I hope the 2.2 support also includes the support for PHP 7.1:

```
composer require algolia/algoliasearch-magento-2
Using version ^1.2 for algolia/algoliasearch-magento-2
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - algolia/algoliasearch-magento-2 1.2.0 requires php ~5.5.0|~5.6.0|~7.0.0 -> your PHP version (7.1.9) does not satisfy that requirement.
    [...]
```

JanPetr:
Hello @avoelkl,

[PHP 7.1 is supported](https://github.com/algolia/algoliasearch-magento-2/commit/4802e189d9d3ce2f94c6b66475dbb4b074d6f717) since version 1.2.1 of the extension.

QuentinFarizonAfrimarket:
Hello @JanPetr 
PHP 7.1 is indeed supported by algolia/algoliasearch-magento-2
But algolia/algoliasearch-magento-2 requires algolia/algoliasearch-client-php": ">=1.27.0 <2.0", that still does not support PHP 7.1 : https://github.com/algolia/algoliasearch-client-php/blob/1.x/composer.json#L22

Is that planned ?

PR #464, opened 2018-04-04
Obliterated usage of ObjectManager in CategoryHelper

peterjaap:
**Summary**

For specific cases, the `algolia_queue_runner` ran out of memory (especially with a large number of jobs or when using `EMPTY_QUEUE`). We diagnosed the problem by using Blackfire and found the objectmanager repeatedly creating objects inside CategoryHelper, hogging up precious memory. By using DI to inject these objects and thereby not re-creating them on every product loop, we fixed the memory leak.

**Result**

No more memory leak.

### DIFF
--- a/Helper/Entity/CategoryHelper.php
+++ b/Helper/Entity/CategoryHelper.php
@@ -97,15 +97,15 @@ class CategoryHelper
         $unretrievableAttributes = [];
 
         foreach ($this->configHelper->getCategoryAdditionalAttributes($storeId) as $attribute) {
-            if ($attribute['searchable'] == '1') {
-                if ($attribute['order'] == 'ordered') {
+            if ($attribute['searchable'] === '1') {
+                if ($attribute['order'] === 'ordered') {
                     $searchableAttributes[] = $attribute['attribute'];
                 } else {
                     $searchableAttributes[] = 'unordered(' . $attribute['attribute'] . ')';
                 }
             }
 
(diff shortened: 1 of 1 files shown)


peterjaap:
@JanPetr is there a specific reason you're using the Object Manager? Because I can't see why?

JanPetr:
Hello @peterjaap and @tdgroot,

sorry I didn't get back to you earlier. After my tests it's working. But I'm afraid of one thing - `$this->collection` now refers to the same instance of `CategoryCollection`. But those collections ([#155](https://github.com/algolia/algoliasearch-magento-2/pull/464/files#diff-4b30278fe7c123401255216cca0e81e4R155), [#318](https://github.com/algolia/algoliasearch-magento-2/pull/464/files#diff-4b30278fe7c123401255216cca0e81e4R318), [R565](https://github.com/algolia/algoliasearch-magento-2/pull/464/files#diff-4b30278fe7c123401255216cca0e81e4R565)) are not the same and use different filters. So eventually depending on order of the calls those collections may give different results.

What do you think about it?

JanPetr:
Hello @peterjaap and @tdgroot, I have found some issues with the changes in this PR so I redid it in #495. Can you take a look there? I'll be happy for any feedback! Thanks a bunch.

I'm closing this PR now.

peterjaap:
Looks good to me!

tdgroot:
Looks good to me, too!

PR #468, opened 2018-04-10
Feat/conversion analytics

JanPetr:
This PR implements [conversion analytics](https://www.algolia.com/doc/guides/analytics/click-analytics/#conversion) based on click analytics.

The conversion is pushed only when a product was found and clicked in search results before an then added to cart. Or if it was added to cart directly from search results.

If it was found, but was added to cart, the conversion shouldn't be pushed to Algolia.

### DIFF
--- a/Block/Algolia.php
+++ b/Block/Algolia.php
@@ -147,6 +147,12 @@ class Algolia extends Template implements CollectionDataSourceInterface
         return $this->registry->registry('current_category');
     }
 
+    /** @return \Magento\Catalog\Model\Product */
+    public function getCurrentProduct()
+    {
+        return $this->registry->registry('product');
+    }
+
     public function getAddToCartParams()
     {
         $url = $this->getAddToCartUrl();
--- a/Block/Algolia.php
+++ b/Block/Algolia.php
@@ -7,6 +7,7 @@ use Algolia\AlgoliaSearch\Helper\ConfigHelper;
 use Algolia\AlgoliaSearch\Helper\Data as CoreHelper;
 use Algolia\AlgoliaSearch\Helper\Entity\CategoryHelper;
 use Algolia\AlgoliaSearch\Helper\Entity\ProductHelper;
+use Magento\Checkout\Helper\Cart;
 use Magento\Framework\App\ActionInterface;
 use Magento\Framework\Data\CollectionDataSourceInterface;
 use Magento\Framework\Data\Form\FormKey;
@@ -22,8 +23,6 @@ class Algolia extends Template implements CollectionDataSourceInterface
 {
     private $config;
     private $catalogSearchHelper;
-    private $storeManager;
-    private $objectManager;
     private $registry;
     private $productHelper;
     private $currency;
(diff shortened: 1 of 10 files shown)


view/frontend/web/click_conversion_analytics.js
```diff
@@ -42,4 +63,57 @@ requirejs(['algoliaBundle', 'algoliaAnalytics'], function(algoliaBundle, algolia
...
+		objectID = objectID.toString();
+		
+		var clickedObjectIds = getObjectIds('clicked');
+		if (!clickedObjectIds[objectID]) {
+			algoliaAnalytics.click({
+				objectID: objectID.toString(),
+				position: parseInt(position)
+			});
+			
+			clickedObjectIds[objectID] = 1;
```
damcou: This kind of array management with numeric indices creates a lot of null values, use a non numeric ("product-[ID]") instead ?

view/frontend/web/click_conversion_analytics.js
```diff
@@ -42,4 +63,57 @@ requirejs(['algoliaBundle', 'algoliaAnalytics'], function(algoliaBundle, algolia
...
+	function trackConversion(objectID) {
+		objectID = objectID.toString();
+		
+		var convertedObjectIds = getObjectIds('converted');
+		if (!convertedObjectIds[objectID]) {
+			algoliaAnalytics.conversion({
+				objectID: objectID
+			});
+			
+			convertedObjectIds[objectID] = 1;
```
damcou: Same as the clickConversion method

view/frontend/web/click_conversion_analytics.js
```diff
@@ -15,14 +18,32 @@ requirejs(['algoliaBundle', 'algoliaAnalytics'], function(algoliaBundle, algolia
...
+			$(document).on('click', algoliaConfig.ccAnalytics.addToCartSelector, function () {
+				var objectId = $(this).data('objectid') || algoliaConfig.productId;
+				
+				setTimeout(function () {
+					trackConversion(objectId);
+				}, 0);
+			});
+		}
+		
+		// "Place order" conversion
```
damcou: Replace this with a onload event on the checkout success page ?

Block/Configuration.php
```diff
@@ -92,6 +94,11 @@ public function getConfiguration()
             }
         }
 
+        $productId = null;
+        if ($config->isClickConversionAnalyticsEnabled() && $request->getControllerName() === 'product') {
```
damcou: Maybe checking also the action ? (if the goal is only to get the ProductId on product view ?)

Block/Algolia.php
```diff
@@ -62,6 +63,7 @@ public function __construct(
         $this->httpContext = $httpContext;
         $this->coreHelper = $coreHelper;
         $this->categoryHelper = $categoryHelper;
+        $this->cart = $cart;
```
damcou: Using "CartHelper" as alias and "$this->cartHelper" as class property may be clearer ?

PR #471, opened 2018-04-13
Prevent backend rendering only for category pages

sanderjongsma:
The prevent backend rendering functionality is only needed for category pages, so when there is no current_category from the registry the observer can return.

**Summary**

I found that the sidebar in the account pages in magento 2 where missing, caused by this bug.

**Result**

Sidebar will only be removed on category pages and if the algolia prevent backend settings apply


sanderjongsma:
@JanPetr can you config if this is a bug in your opinion and when it will be merged if it is.

JanPetr:
Thanks @sanderjongsma! 🚀

PR #473, opened 2018-04-17
Change comparison operator to == instead of === for nonzerodefault price

peterjaap:
We were experiencing a number of products having a price of 0 (zero) in the category pages but not on the product pages. After debugging with xdebug, I've found the comparison operator in PriceManager to be the culprit;

```
if ($customData[$field][$currencyCode]['default'] === 0) {
    $customData = $this->handleZeroDefaultPrice(
        $customData,
        $field,
        $currencyCode,
        $baseCurrencyCode,
        $min,
        $max,
        $store
    );
}
```

This checks for a zero default value by `===`. But PHP's `round()` function returns a float, and not an int. Changing this to `==` loosens the comparison and fixes the issue.

### DIFF
--- a/Helper/Entity/Product/PriceManager.php
+++ b/Helper/Entity/Product/PriceManager.php
@@ -139,7 +139,7 @@ class PriceManager
                         );
                     }
 
-                    if ($customData[$field][$currencyCode]['default'] === 0) {
+                    if ($customData[$field][$currencyCode]['default'] == 0) {
                         $customData = $this->handleZeroDefaultPrice(
                             $customData,
                             $field,
--- a/Helper/Entity/Product/PriceManager.php
+++ b/Helper/Entity/Product/PriceManager.php
@@ -139,7 +139,7 @@ class PriceManager
                         );
                     }
 
-                    if ($customData[$field][$currencyCode]['default'] == 0) {
+                    if (!$customData[$field][$currencyCode]['default']) {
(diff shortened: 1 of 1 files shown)


Helper/Entity/Product/PriceManager.php
```diff
@@ -139,7 +139,7 @@ public function addPriceData($customData, Product $product, $subProducts)
                         );
                     }
 
-                    if ($customData[$field][$currencyCode]['default'] === 0) {
+                    if ($customData[$field][$currencyCode]['default'] == 0) {
```
JanPetr: I'd rather not drop the strict comparison, but have it compared with `0.0` float value.

peterjaap: Why? There is no harm in doing a loose comparison here. A falsey value should always hit the zerodefault method, no matter the reason why.

peterjaap: I'd even go as far as changing the line to

`if (!$customData[$field][$currencyCode]['default']) {`

Since, as I said, *any* falsey value should hit the handleZeroDefaultPrice method.

JanPetr: Makes sense, can you change the `== 0` to `!` statement?

peterjaap: Done :+1:

PR #474, opened 2018-04-17
Also added check for nonCastableAttributes to castRecord method

peterjaap:
We were experiencing that a product with the sku `105.05680` came in Algolia with the sku `105.0568` (so without the trailing zero).

This is exactly the same issue as we had with https://github.com/algolia/algoliasearch-magento-2/pull/375, but that PR is missing the check in `castRecord`, it only has it in `castProductObject`.

### DIFF
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -12,6 +12,9 @@ use Symfony\Component\Console\Output\ConsoleOutput;
 
 class AlgoliaHelper extends AbstractHelper
 {
+    /** @var array */
+    protected $nonCastableAttributes = ['sku', 'name', 'description'];
+
     /** @var Client */
     private $client;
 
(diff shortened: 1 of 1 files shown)


Helper/AlgoliaHelper.php
```diff
@@ -12,6 +12,9 @@
 
 class AlgoliaHelper extends AbstractHelper
 {
+    /** @var array */
+    protected $nonCastableAttributes = ['sku', 'name', 'description'];
```
JanPetr: Can you make it `private` and definite setter / getter method(s)?

peterjaap: Not really. Getter/setter are anti-patterns.

peterjaap: Using accessors only adds bloatware in this case. Not to mention that accessors aren't widely used in Magento and `private`'s are considered evil by a lot of Magento devs. We write a lot of Magento extensions and generally make everything public since there is no good reason to make some protected or private. 

Keeping it protected allows the developer who wants to do a change to the extension upon the same variable instead of having to pluginize accessors to change one small line.

peterjaap: Here's my go-to-link on the subject; http://www.yegor256.com/2014/09/16/getters-and-setters-are-evil.html

peterjaap:
We can close this since this is in the latest release, albeit with a private variable (but not accessors).

PR #478, opened 2018-04-19
Set correct type IDs for image type selector in configuration

peterjaap:
When setting 'small image' as the image type for the images used in Algolia, we noticed base image was used. When changing the value in the database to `product_small_image` instead of `small_image`, it worked like we wanted.

Reference for image type ID's can be found in Magento core; `vendor/magento/theme-frontend-blank/etc/view.xml`


JanPetr:
Thanks Peter! 💯

PR #479, opened 2018-04-25
Fixed exploding of line breaks where \r\n was used instead of \n

peterjaap:
**Summary**
We noticed the prevent backend rendering functionality did not work correctly; the pages for the Googlebot did not include the HTML of Magento's category listing. We found that it was due to a `\r\n` being present in the `algoliasearch_advanced/advanced/backend_rendering_allowed_user_agents`  value. Probably due to somebody on Windows editing that text. The code only split on `\n`, thereby getting `Googlebot\r` and `Bingbot` as a result. `Googlebot\r` does not match in the user agent string `Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)` and there it *will* prevent backend rendering.

**Result**
Backend rendering is *not* prevented for Googlebot.

### DIFF
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -682,7 +682,8 @@ class ConfigHelper
             return true;
         }
 
-        $allowedUserAgents = explode("\n", $allowedUserAgents);
+        $allowedUserAgents = preg_split('/\n|\r\n?/', $allowedUserAgents);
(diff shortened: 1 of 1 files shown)


peterjaap:
As a bonus, this also filters out empty lines, avoiding warnings thrown by `strpos` when it encounters an empty string.

JanPetr:
Great job, Peter! Thank you 🙌

PR #482, opened 2018-05-04
Reindex parent products (configurable, bundle, grouped) when child product is updated

unicoder88:
**Summary**

1. There's a configurable product and assigned simple products
1. Last simple product is sold out and goes out of stock, so does configurable product.
1. Algolia only updates simple product, but not parent.

**Result**

As a result, search results show configurable product as in stock, while it isn't.

I suggest reindexing all parent products when child ones are updated (but not in full reindex mode of course).

Please review :)

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -13,6 +13,8 @@ use Algolia\AlgoliaSearch\Helper\AlgoliaHelper;
 use Algolia\AlgoliaSearch\Helper\ConfigHelper;
 use Algolia\AlgoliaSearch\Helper\Logger;
 use Magento\Catalog\Model\Product\Visibility;
+use Magento\Catalog\Model\Product\Type;
+use Magento\Catalog\Model\Product\Type\AbstractType;
 use Magento\CatalogInventory\Api\StockRegistryInterface;
 use Magento\CatalogInventory\Helper\Stock;
 use Magento\Directory\Model\Currency as CurrencyHelper;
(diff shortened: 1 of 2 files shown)


Model/Indexer/Product.php
```diff
@@ -62,6 +62,10 @@ public function execute($productIds)
             return;
         }
 
+        if ($productIds) {
+            $productIds = array_merge($productIds, $this->productHelper->getParentProductids($productIds));
```
JanPetr: Typo, method name is `getParentProductIds` :)

Helper/Entity/ProductHelper.php
```diff
@@ -459,6 +473,42 @@ private function getSubProducts(Product $product)
...
+    /**
+     * Returns composite product type instances
+     *
+     * @return AbstractType[]
+     * @throws \Magento\Framework\Exception\LocalizedException
+     * @see \Magento\Catalog\Model\Indexer\Product\Flat\AbstractAction::_getProductTypeInstances
+     */
+    private function getCompositeTypes()
+    {
+        if (null === $this->compositeTypes) {
```
JanPetr: Could you swap it to `$this->compositeTypes === null`? We keep this convention all around the extension.

unicoder88: Absolutely, I prefer it as well

Model/Indexer/Product.php
unicoder88: Sharp eye :)

PR #501, opened 2018-05-31
hitsPerPage gets overwritten for first load fixed

satishgadhave:
**Summary**
On any list page first load, URL does not contain hash segments. Those are added after page load by JS.
This means, code to add hierarchicalFacetsRefinements in searchParameters is executed, which is fine.
Problem is it overwrites parameters set above in instantsearchOptions initialization code e.g. hitsPerPage

**Result**
Result is, for first load, request to Algolia does not contain hitsPerPage parameter and default number of records are loaded i.e. 20 even if you have set 100 in the extension backend


JanPetr:
Great catch @satishgadhave ! Thank you for the contribution 🚀

Issue #420, opened 2018-02-08
Customer group catalog price rules not working


SilvanLaroo:
Hi,

We've created specific catalog price rules for our Wholesale customer group. This groups gets 10% discount by default. This discount is shown on the product page, but the old price is shown in the Algolia Search results. 

Can Algolia handle this specific catalog price rules?

michaelwylde:
I am also experiencing what I think is the same issue, logged in customers receive 10% discount through a catalog price rule. The catalog price is displayed on the PDP but not the PLP/search for configurable products (it does work for simple products).

The reason for this is because the handlePrice function in the ProductHelper checks the rulePrice for the configurable product, however Magento only stores the rulePrice against the simple and therefore the group price sent to Algolia does not reflect the rules price.

The code to add ranged price for configurable products does not pick this up as it does not reconsider catalog price rules based on groups.

peterjaap:
We're also experiencing this issue and @michaelwylde's explanation is correct.

peterjaap:
Ok, found it.

Note; this only works when all child products have the same price (so no range).

Replace `handleZeroDefaultPrices` with;

```
    private function handleZeroDefaultPrice(
        $customData,
        $field,
        $currencyCode,
        $baseCurrencyCode,
        $min,
        $max,
        Store $store,
        $subProducts
    ) {
        $customData[$field][$currencyCode]['default'] = $min;

        if ($min !== $max) {
            return $customData;
        }

        if ($currencyCode !== $baseCurrencyCode) {
            $min = $this->priceCurrency->convert($min, $store, $currencyCode);
        }

        $minFormatted = $this->priceCurrency->format(
            $min,
            false,
            PriceCurrencyInterface::DEFAULT_PRECISION,
            $store,
            $currencyCode
        );

        $defaultOriginals = [];
        foreach ($subProducts as $subProduct) {
            $defaultOriginals[] = $subProduct->getPrice();
        }

        $defaultOriginal = min($defaultOriginals);

        $defaultOriginalFormated = $this->priceCurrency->format(
            $defaultOriginal,
            false,
            PriceCurrencyInterface::DEFAULT_PRECISION,
            $store,
            $currencyCode
        );


        $customData[$field][$currencyCode]['default_original'] = $defaultOriginal;
        $customData[$field][$currencyCode]['default_original_formated'] = $defaultOriginalFormated;
        $customData[$field][$currencyCode]['default'] = $min;
[...]

PR #512, opened 2018-06-22
Fix special price for configurables

peterjaap:
**Summary**
See issue #420. The pricing logic for configurables with a catalog price rule are incorrect. In the `handleZeroDefaultPrice` method the original values are not set, thereby not showing the original price on the frontend (and thus showing the special price as the normal price). This applies to configurables with child products that all have the same price as well as for configurables with child products with differing prices (price range, a.k.a. min/max).

This pull request fixes both issues and adds min/max fields that are synced to Algolia. This way, we can leave it up to the implementation specialist to show the range or just the 'From' price (in `instant/hit.phtml`).

**Before (only example for single price for all children)**

image

JSON:

```
  "price": {
    "EUR": {
      "default": 895,
      "default_formated": "€ 895,00",
      "special_from_date": false,
      "special_to_date": false
    }
  },
```

**After**
*Single price for all children*

image

JSON:

```
  "price": {
    "EUR": {
      "default": 863.68,
      "default_formated": "€ 863,68",
      "special_from_date": false,
      "special_to_date": false,
      "default_original": "895.0000",
      "default_original_formated": "€ 895,00"
    }
  },
```

*Differing prices for children (range)*

image

JSON:

```
  "price": {
    "EUR": {
      "default": 800,
      "default_formated": "€ 800,00 - € 863,68",
      "special_from_date": false,
      "special_to_date": false,
      "default_min": 800,
      "default_min_formated": "€ 800,00",
      "default_max": 863.68,
      "default_max_formated": "€ 863,68",
      "default_original": "850.0000",
      "default_original_min": "850.0000",
      "default_original_min_formated": "€ 850,00",
      "default_original_max": "895.0000",
      "default_original_max_formated": "€ 895,00",
      "default_original_formated": "€ 850,00 - € 895,00"
    }
  },
```

### DIFF
--- a/Helper/Entity/Product/PriceManager.php
+++ b/Helper/Entity/Product/PriceManager.php
@@ -139,6 +139,7 @@ class PriceManager
                         );
                     }
 
+
                     if (!$customData[$field][$currencyCode]['default']) {
                         $customData = $this->handleZeroDefaultPrice(
                             $customData,
@@ -147,7 +148,8 @@ class PriceManager
                             $baseCurrencyCode,
                             $min,
                             $max,
-                            $store
+                            $store,
+                            $subProducts
                         );
                     }
 
@@ -503,11 +505,26 @@ class PriceManager
         $baseCurrencyCode,
         $min,
         $max,
-        Store $store
+        Store $store,
+        $subProducts
     ) {
         $customData[$field][$currencyCode]['default'] = $min;
 
+        $defaultOriginals = [];
+        foreach ($subProducts as $subProduct) {
+            $defaultOriginals[] = $subProduct->getPrice();
+        }
+
+        $defaultOriginal = min($defaultOriginals);
+        $defaultOriginalMax = max($defaultOriginals);
+
         if ($min !== $max) {
+            $customData[$field][$currencyCode]['default_min'] = $min;
+            $customData[$field][$currencyCode]['default_max'] = $max;
+            $customData[$field][$currencyCode]['default_original'] = $defaultOriginal;
+            $customData[$field][$currencyCode]['default_original_min'] = $defaultOriginal;
+            $customData[$field][$currencyCode]['default_original_max'] = $defaultOriginalMax;
+            $customData[$field][$currencyCode]['default_original_formated'] = $this->getDashedPriceFormat($defaultOriginal, $defaultOriginalMax, $store, $currencyCode);
             return $customData;
         }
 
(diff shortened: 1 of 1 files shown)


peterjaap:
I feel like this whole pricemanager thing needs a complete rewrite by  now....

JanPetr:
Hello @peterjaap,

thank you so much for your work! ❤️ 

Honestly, you're completely right. I'm aware that pricing calculation is the extension's weak spot and needs to be refactored / rewritten. Your work is a great base for that.
Unfortunately I don't have time to take a proper look on it now, but I'll get to it next week.

If you have any ideas on how to approach it, what should be covered etc. feel free to shoot it here. As comment or commit directly 🙌 

Thank you again!

Helper/Entity/Product/PriceManager.php
```diff
@@ -503,26 +507,108 @@ private function handleZeroDefaultPrice(
...
+        }
+
         $customData[$field][$currencyCode]['default'] = $min;
 
+        $defaultOriginals = [];
+        foreach ($subProducts as $subProduct) {
+            $defaultOriginals[] = $subProduct->getPrice();
+        }
+
+        $defaultOriginal = min($defaultOriginals);
```
JanPetr: This fails when there are no `$subProducts`. And then it has consequences to further code.

Helper/Entity/Product/PriceManager.php
```diff
@@ -503,26 +507,108 @@ private function handleZeroDefaultPrice(
...
+                        $currencyCode
+                    );
+                    $customData[$field][$currencyCode]['group_' . $groupId . '_formated'] = $groupSpecialPriceFormated;
+                }
+            }
+        }
+
         $customData[$field][$currencyCode]['default'] = $min;
 
+        $defaultOriginals = [];
```
JanPetr: How about extracting code from here to line 593 to a separate method as it returns value on line 592? It'll make the code much readable and understandable.

Helper/Entity/Product/PriceManager.php
```diff
@@ -503,26 +507,108 @@ private function handleZeroDefaultPrice(
...
+        foreach ($subProducts as $subProduct) {
+            $defaultOriginals[] = $subProduct->getPrice();
+        }
+
+        $defaultOriginal = min($defaultOriginals);
+        $defaultOriginalMax = max($defaultOriginals);
```
JanPetr: The same as ^

Helper/Entity/Product/PriceManager.php
peterjaap: Right, that makes no sense. Maybe we should add and `if empty($subProducts) { $defaultOriginals[] = <original product price here>; }` ?

Helper/Entity/Product/PriceManager.php
peterjaap: Yes good idea.

Helper/Entity/Product/PriceManager.php
peterjaap: Yes!

peterjaap:
Putting this on hold, this needs to be tested thoroughly. Other pricing issues are popping up when using this code (zero prices and incorrect from/to prices).

PR #532, opened 2018-07-24
Fixed incorrect replacement of "+" and "-" of toggle button of facet panel when displayed on small screen

tamada1:
**Summary**

In https://github.com/algolia/algoliasearch-magento-2/blob/d7482b561eb2f3fe41867c83fef130fd08f8c728/view/frontend/templates/instant/wrapper.phtml ...
```
                <div id="refine-toggle" class="visible-xs visible-sm">
                    + <?php echo $block->escapeHtml(__('Refine')); ?>
                </div>
```
It is line terminator characters that is obtained from the above elements using `$(this).html()[0]`.
If we try to get the string "+" or "-", we need to get rid of the obstacles first.

**Result**

Replacement of "+" and "-" is correct.


JanPetr:
Great catch, @tamada1! Thank you for your contribution 🚀 🙌 
Feel free to send any other PR, we'll be very happy to review and merge! 💪

PR #544, opened 2018-08-11
Check if staging modules are being used

kingshukdeb:
Here you are converting entity_id to row_id if customer is using Magento EE and version is greater than 2.1.0 but lots of customer may not use staging modules which actually changes entity_id to row_id.
Like we are using dumpautoload to remove all staging modules. I have just added a check if Magento_Staging module is in play. Because all staging modules are dependent on this one. So if this one exists then we can convert entity_id to row_id. Else we wont.

**Summary**

**Result**


kingshukdeb:
@JanPetr Hey man can you please check if its fine and merge it.

PR #545, opened 2018-08-14
Check product attribute uses source to prevent spoiling attribute source model

unicoder88:
**Summary**

When indexing additional attributes, in method `addNonNullValue` we render attribute text. However, this method should only be called when attribute uses source model, otherwise default source model `Magento\Eav\Model\Entity\Attribute\Source\Table` is implicitly assigned even for e.g. *text attributes*.

- [getAttributeText](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/Model/Product.php#L1752)
- [getSource](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Eav/Model/Entity/Attribute/AbstractAttribute.php#L596)
- [_getDefaultSourceModel](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/Model/ResourceModel/Eav/Attribute.php#L404)

This results in 3rd party extensions using attributes to treat regular text attributes as dropdowns and further unexpected errors.

I searched for alternative in default [Additional attributes](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/Block/Product/View/Attributes.php#L77) block on product view page.
There they just call `$attribute->getFrontend()->getValue($product)` to render option text. There they check frontend input type `select` or `boolean` before accessing source model. I suggest simpler solution by just calling attribute [usesSource](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Eav/Model/Entity/Attribute/AbstractAttribute.php#L620) check what will do the same job.

I recorded slightly messy video that should help explain what issue I'm having.

https://vimeo.com/284982511
password `algolia`

Looking forward to your comments :)


JanPetr:
Hello @unicoder88,

thank you for the PR and sorry for the late answer.
We'll take a proper look and come back to you.

Thank you for your time! 💯

zyukov:
I also faced such problem. As a  result I have problems with category_product_flat table.

JanPetr:
Hello @unicoder88,

thank you for the great catch and your contribution! Sorry it took a while to merge it, but it'll be part of the next release 🎉 

[Thank you](http://bit.ly/2KHTYV3) again! 🙌

Issue #540, opened 2018-08-09
Wrong CMS page urls indexed when using different domain per store


vanpoppel:
**Do you want to request a *feature* or report a *bug*?**
bug

**Bug: What is the current behavior?**
On a Magento2 install with multiple stores, each with their own domain, all cms page urls refer to the default store (i guess the one that's set as default, store 0). All other types work normally.

Example: if i have `foo.test` as default and `bar.test` for a second store, both fully indexed, the search results for cms pages will in both stores return urls pointing to `foo.test`. 

**Bug: What is the expected behavior?**
CMS pages should refer to pages within the current store scope. So on `bar.test` the results should redirect to the correct url within `bar.test`.

**Bug: What is the proposed solution?**
Fix the indexing. It seems that Algolia\AlgoliaSearch\Helper\Entity\PageHelper::getPages() is returning the wrong URL.

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**
Magento 2.2.5, AlgoliaSearch 1.7.1

JanPetr:
Hello @vanpoppel,

thank you for the report!
I tried it and I'm not able to reproduce it. PageHelper is ready for multi-store and is supposed to give correct domain:
- https://github.com/algolia/algoliasearch-magento-2/blob/master/Helper/Entity/PageHelper.php#L106
- https://github.com/algolia/algoliasearch-magento-2/blob/master/Helper/Entity/PageHelper.php#L129

Could you investigate in the code to see why it gives you wrong URL?

vanpoppel:
Hi @JanPetr ,

Thanks for your reply. I think i have found the issue. Though haven't solved it yet.

This line: https://github.com/algolia/algoliasearch-magento-2/blob/master/Helper/Entity/PageHelper.php#L106 

There's a loop through storeids, within an emulated environment (rebuildStorePageIndex() in Data helper). So in the first run, three url objects are created emulated as store 1. The second and third run, these objects will be re-used, in stead of created for the current emulated env.

PR #551, opened 2018-08-22
Create Url object once per emulated store

vanpoppel:
**Summary**
This function is run per store within an emulated environment. It does not make sense to loop through the stores again. 

This change makes sure there's only one Url object per store which will return the correct Urls.

**Motivation**
Issue with indexing on multistore: https://github.com/algolia/algoliasearch-magento-2/issues/540

**Result**

From a multistore setup with 2 websites and 3 stores, just run these commands:
bin/magento indexer:reindex algolia_pages
bin/magento indexer:reindex algolia_queue_runner
All page urls are now entered into Algolia correctly.

### DIFF
--- a/Helper/Entity/PageHelper.php
+++ b/Helper/Entity/PageHelper.php
@@ -128,26 +128,13 @@ class PageHelper
 
     private function getStoreUrl($storeId)
     {
-        if ($this->storeUrls === null) {
-            $this->storeUrls = [];
-            $storeIds = $this->getStores();
-
-            foreach ($storeIds as $storeId) {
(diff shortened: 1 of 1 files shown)


vanpoppel:
You're welcome @JanPetr 

Can you create a new release, so I can implement the fix properly?

JanPetr:
Hello @vanpoppel,

we have multiple things we want to release now. We'll work on a proper release in the next 2 weeks and your change will be part of it. Thank you for the patience!

peterjaap:
Direct usage of ObjectManager is highly discouraged.

vanpoppel:
I absolutely agree. I wanted to change as little as possible here, as we needed a fix. Working on a better fix.

PR #572, opened 2018-09-24
Fix error on checkout when autocomplete is disabled

damcou:


view/frontend/web/instantsearch.js
```diff
@@ -1,8 +1,8 @@
 requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBundle, priceUtils) {
 	algoliaBundle.$(function ($) {
-		
+
 		/** We have nothing to do here if instantsearch is not enabled **/
-		if (!algoliaConfig.instant.enabled || !(algoliaConfig.isSearchPage || !algoliaConfig.autocomplete.enabled)) {
+		if (!algoliaConfig.instant.enabled || !(algoliaConfig.isSearchPage || !algoliaConfig.autocomplete.enabled) || (!algoliaConfig.autocomplete.enabled && $(algoliaConfig.autocomplete.selector).length == 0)) {
```
JanPetr: The condition is really complicated (it was even before). How about extracting it to standalone and more readable function?

PR #587, opened 2018-10-03
Allow other extension providers to update potentially long attributes

brideo:
Some people require to unset attributes if they are too long rather
than not skipping the full product. So they can easily inject additonal
attribu


JanPetr:
Hello @brideo,

thank you for the PR, it definitely makes sense!
But could you rather write a getter and a setter instead of adding it to a constructor? I don't want to pollute the constructor with internals.

Thank you!

brideo:
Hello @JanPetr,

I just noticed you have this for non-castables now:

```
        // Merge non castable attributes set in config
        $this->nonCastableAttributes = array_merge(
            $this->nonCastableAttributes,
            $this->config->getNonCastableAttributes()
        );
```

Should we follow the same strategy for this so we're consistent?

JanPetr:
Hello @brideo,

sorry for not coming back to you earlier /o\

Yes, keeping the same strategy as for non-castable attributes would be awesome! Will you modify your PR?

brideo:
Hey @JanPetr,

In the latest version it looks like `$potentiallyLongAttributes` is null?

Issue #480, opened 2018-04-25
Prevent Backend Rendering does not work correctly with cache enabled


peterjaap:
**Do you want to request a *feature* or report a *bug*?**
'Bug'

**Bug: What is the current behavior?**
Prequisites;
1. Prevent Backend Rendering is enabled
1. Allow backend rendering user agent list includes 'Googlebot'
1. Cache (full page and/or block) is enabled
1. First request on uncached category page is from non-Googlebot
1. Consecutive request is from Googlebot
1. Problem; result for Googlebot does not include backend rendering

**Bug: What is the expected behavior?**
In step 6, the result should include backend rendering.

**Bug: What is the proposed solution?**
I don't have a proposed solution. 'Workaround' now is to disable Prevent Backend Rendering or to disable full page and block cache.

JanPetr:
Hello @peterjaap,

thank you for the report! Indeed, not an easy issue to fix properly. We will investigate and see what we can come up with. Maybe changing and cache key could help 🤔

peterjaap:
Maybe you can make use of Magento's `X-Vary` header to differentiate based on user agents?

peterjaap:
If you need more info on FPC in Magento 2, Mage2.tv will release a full series on FPC soon; https://www.mage2.tv/content/full-page-cache/

damcou:
Hello @peterjaap ,

I added a commit related to this issue, would it be possible to test the Plugin on your side and give me some feedback about it ? 

https://github.com/algolia/algoliasearch-magento-2/pull/589/commits/5973cc2ffd3f9829babb05308bd8c3ce9257bc62

Thanks in advance ;)

peterjaap:
Cool @damcou thanks! Will try to test next week.

PR #589, opened 2018-10-05
Fix & Enhance backend rendering

damcou:
This feature fixes and enhances the backend rendering (according to the new routing system introduced by Instantsearch V2 : https://community.algolia.com/instantsearch.js/v2/guides/routing.html

### DIFF
--- a/Adapter/Algolia.php
+++ b/Adapter/Algolia.php
@@ -130,9 +130,8 @@ class Algolia implements AdapterInterface
                 $documents = $this->getDocumentsFromAlgolia($algoliaQuery, $storeId);
             }
 
-            $apiDocuments = array_map(array($this, 'getApiDocument'), $documents);
+            $apiDocuments = array_map([$this, 'getApiDocument'], $documents);
             $table = $temporaryStorage->storeApiDocuments($apiDocuments);
-
         } catch (AlgoliaConnectionException $e) {
             $this->nativeQuery($request);
         }
@@ -159,6 +158,7 @@ class Algolia implements AdapterInterface
             'documents' => $documents,
             'aggregations' => $aggregations,
         ];
+
         return $this->responseFactory->create($response);
     }
 
@@ -210,7 +210,7 @@ class Algolia implements AdapterInterface
                 'facetFilters' => [
                     'categoryIds:' . $category->getEntityId(),
                 ],
-                'page' => $page
+                'page' => $page,
             ];
 
             $facetFilters = [];
@@ -243,7 +243,7 @@ class Algolia implements AdapterInterface
 
             // Handle price filtering
             $currencyCode = $this->storeManager->getStore()->getCurrentCurrencyCode();
-            $priceSliders[] = 'price_' .  $currencyCode . '_default';
+            $priceSliders[] = 'price_' . $currencyCode . '_default';
 
             if ($this->config->isCustomerGroupsEnabled($storeId)) {
                 $groupCollection = $this->objectManager
--- a/Adapter/Algolia.php
+++ b/Adapter/Algolia.php
@@ -6,6 +6,7 @@ use Algolia\AlgoliaSearch\Helper\ConfigHelper;
 use Algolia\AlgoliaSearch\Helper\Data as AlgoliaHelper;
 use AlgoliaSearch\AlgoliaConnectionException;
 use Magento\CatalogSearch\Helper\Data;
+use Magento\Customer\Model\Session as CustomerSession;
 use Magento\Framework\App\Request\Http;
 use Magento\Framework\App\ResourceConnection;
 use Magento\Framework\DB\Ddl\Table;
@@ -51,6 +52,9 @@ class Algolia implements AdapterInterface
     /** @var Registry */
     private $registry;
 
+    /** @var CustomerSession */
+    private $customerSession;
+
     /** @var AlgoliaHelper */
     private $algoliaHelper;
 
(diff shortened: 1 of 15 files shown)


peterjaap:
Are you taking FPC into account? See https://github.com/algolia/algoliasearch-magento-2/issues/480

Adapter/Algolia.php
```diff
@@ -90,140 +96,200 @@ public function __construct(
...
+                'facetFilters' => [
+                    'categoryIds:' . $category->getEntityId(),
+                ],
+                'page' => $page,
+            ];
+
+            $facetFilters = [];
+
+            foreach ($this->config->getFacets($storeId) as $facet) {
+                if (!is_null($this->request->getParam($facet['attribute']))) {
```
JanPetr: You can change this condition to:
```
if (is_null($this->request->getParam($facet['attribute']))) {
    continue;
}

// rest of the method
```

It'll lower the depth of the method and will be slightly more readable :)

Adapter/Algolia.php
```diff
@@ -90,140 +96,200 @@ public function __construct(
...
+                    if ($facet['type'] === 'disjunctive') {
+                        $facetValues = explode('~', $this->request->getParam($facet['attribute']));
+                        if (count($facetValues) > 1) {
+                            foreach ($facetValues as $key => $facetValue) {
+                                $facetValues[$key] = $facet['attribute'] . ':' . $facetValue;
+                            }
+                            $facetFilters[] = $facetValues;
+                        }
+                        if (count($facetValues) == 1) {
+                            $facetFilters[] = $facet['attribute'] . ':' . $facetValues[0];
```
JanPetr: It's not that easy with `disjunctive` faceting in PHP. Disjunctive facets need to be sent in different query paramter: https://github.com/algolia/algoliasearch-client-php/blob/master/src/AlgoliaSearch/Index.php#L614

Adapter/Algolia.php
```diff
@@ -90,140 +96,200 @@ public function __construct(
...
-        if (version_compare($this->config->getMagentoVersion(), '2.1.0', '<') === true) {
-            return $this->aggregationBuilder->build($request, $table);
-        }
+        $contextParams = [];
+        $category = $this->registry->registry('current_category');
+        $page = !is_null($this->request->getParam('page')) ?
+            (int) $this->request->getParam('page') - 1 :
+            0;
+
+        if ($category) {
```
JanPetr: Most of the context parameters here are set only on a category page (where `$category` is set), but facets and other parameters can be set on search results page as well.

Adapter/Algolia.php
```diff
@@ -90,140 +96,200 @@ public function __construct(
...
-            return 'getDocument20';
+            // Handle price filtering
+            $currencyCode = $this->storeManager->getStore()->getCurrentCurrencyCode();
+            $priceSliders[] = 'price_' . $currencyCode . '_default';
+
+            if ($this->config->isCustomerGroupsEnabled($storeId)) {
+                $groupCollection = $this->objectManager
+                    ->create('Magento\Customer\Model\ResourceModel\Group\Collection');
+
+                foreach ($groupCollection as $group) {
```
JanPetr: The query doesn't have to iterate over all customer groups. You can get a customer group of the currently logged in customer and filter only on his specific price attribute, no?

Adapter/Algolia.php
```diff
@@ -90,140 +96,200 @@ public function __construct(
...
      */
-    private function isAllowed($storeId)
+    private function getDocumentsFromAlgolia($algoliaQuery, $storeId)
     {
-        return
-            $this->config->getApplicationID($storeId)
-            && $this->config->getAPIKey($storeId)
-            && $this->config->isEnabledFrontEnd($storeId)
-            && $this->config->makeSeoRequest($storeId);
+        $contextParams = [];
```
JanPetr: In Algolia we call it a `searchParameters` rather than `contextParameters`. Can we keep that naming (even for further variables and methods) or did you choose `contextParameters` for a specific reason?

view/frontend/web/instantsearch.js
```diff
@@ -96,6 +95,60 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
 					$('.algolia-instant-replaced-content').hide();
 					$('.algolia-instant-selector-results').show();
 				}
+			},
+			routing : {
+        stateMapping: {
```
JanPetr: Could you fix the indentation? In JS files we are stuck with `tabs` :/ I'm aware of the inconsistency and eventually we'll transform it to spaces, but I'd rather do it in a separate branch to not to pollute a feature branch.

view/frontend/web/instantsearch.js
```diff
@@ -96,6 +95,60 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
 					$('.algolia-instant-replaced-content').hide();
 					$('.algolia-instant-selector-results').show();
 				}
+			},
+			routing : {
```
JanPetr: Could you write a small comment above the `routing:` explaining what URLs the routing is generating?

peterjaap:
This is with my default User-Agent;

image

When I set the User-Agent to Googlebot and I open a category, I get this;

image

So that seems to work just fine! My question is about the display:none. Will Google not consider this data hidden and therefore not index it even though it is in the DOM?

PR #592, opened 2018-10-17
Fix autocomplete bottom links

damcou:
Remove "ors" category links when the categories facet is not active


JanPetr:
@damcou, I added more comments to M1 PR: https://github.com/algolia/algoliasearch-magento/pull/1066 and it applies here as well. Could you fix that in here as well? 🙏

PR #593, opened 2018-10-17
Fix deactivated serialized arrays

damcou:
Fix deactivated serialized arrays : when IS or autocomplete are turned off in the Magento configuration, prevent serialized array fields to be emptied.

### DIFF
--- a/Controller/Adminhtml/System/Config/Save.php
+++ b/Controller/Adminhtml/System/Config/Save.php
@@ -18,7 +18,6 @@ class Save extends \Magento\Config\Controller\Adminhtml\System\Config\Save
     {
         if (isset($groups['instant']['fields']['is_instant_enabled']['value'])
                 && $groups['instant']['fields']['is_instant_enabled']['value'] == '0') {
-
             foreach ($this->instantSerializedValues as $field) {
                 if (isset($groups['instant']['fields'][$field])) {
                     unset($groups['instant']['fields'][$field]);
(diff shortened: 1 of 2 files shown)


Controller/Adminhtml/System/Config/Save.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Controller\Adminhtml\System\Config;
+
+class Save extends \Magento\Config\Controller\Adminhtml\System\Config\Save
+{
+    /**
+     * Get groups for save
+     *
```
JanPetr:
```suggestion
```

Controller/Adminhtml/System/Config/Save.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Controller\Adminhtml\System\Config;
+
+class Save extends \Magento\Config\Controller\Adminhtml\System\Config\Save
+{
+    /**
+     * Get groups for save
```
JanPetr:
```suggestion
```

Controller/Adminhtml/System/Config/Save.php
```diff
@@ -0,0 +1,44 @@
...
+        $groups = parent::_getGroupsForSave();
+
+        return $this->handleDeactivatedSerializedArrays($groups);
+    }
+
+    private function handleDeactivatedSerializedArrays($groups)
+    {
+        if (isset($groups['instant']['fields']['is_instant_enabled']['value'])
+                && $groups['instant']['fields']['is_instant_enabled']['value'] == '0') {
+            if (isset($groups['instant']['fields']['facets'])) {
```
JanPetr: How about specifying names of the fields in a property, eg. `$this->instantSerializedValues = ['facets', 'sorts'];` and then iterate over it and don't write a specific condition for each field. It'll be better to maintain and the code will be shorter and more readable.

The same could apply to autocomplete fields.

PR #604, opened 2018-11-01
Fix error with configuration save while query rules are disabled in Algolia engine

JanPetr:


Helper/Entity/ProductHelper.php
```diff
@@ -372,7 +372,13 @@ public function setSettings($indexName, $indexNameTmp, $storeId, $saveToTmpIndic
         }
 
         if ($saveToTmpIndicesToo === true) {
-            $this->algoliaHelper->copyQueryRules($indexName, $indexNameTmp);
+            try {
+                $this->algoliaHelper->copyQueryRules($indexName, $indexNameTmp);
+            } catch(AlgoliaException $e) {
+                if ($e->getMessage() !== 'Query Rules are not enabled on this application') {
```
0livier: I feel that adding a small context on why do ditch that Exception silently would be nice. (link to public slack convos?)

0livier: Oh you did put it in another file :thinking:

JanPetr: Good catch, @0livier. I added a comment. Works?

PR #616, opened 2018-11-16
New routing system + fixed backend rendering

damcou:
This feature fixes and enhances the backend rendering (according to the new routing system introduced by Instantsearch V2 : https://community.algolia.com/instantsearch.js/v2/guides/routing.html


Adapter/Algolia.php
```diff
@@ -78,7 +93,10 @@ public function __construct(
         ConfigHelper $config,
         Data $catalogSearchHelper,
         StoreManagerInterface $storeManager,
+        Registry $registry,
+        CustomerSession $customerSession,
         AlgoliaHelper $algoliaHelper,
+        ProductHelper $productHelper,
         Http $request,
         DocumentFactory $documentFactory
```
JanPetr: Great job, on the class, @damcou! It looks really good.

One thing - could you try to refactor / split the class to multiple smaller classes as the class has 9 dependencies,
now. This is really not great for a future maintain ability and we should keep the number of class (constructor) dependencies as low as possible. A single class should have [max 4-5 dependencies](https://stackoverflow.com/questions/953407/constructor-injection-how-many-dependencies-is-too-many) as it usually means it breaks [Single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle).
A good way to start splitting is to identify methods which uses a common dependencies and move them to a more self-descriptive smaller class and inject only the newly created class.

I know, the extension is full "large" constructors, but for the future it makes sense to start refactoring it now and we need to start somewhere :) Thank you!

Adapter/Algolia.php
```diff
@@ -90,90 +108,272 @@ public function __construct(
...
+     * @param int $storeId
+     *
+     * @return array
+     */
+    private function getSearchParams($storeId)
+    {
+        $searchParams = [];
+        $searchParams['facetFilters'] = [];
+
+        // Handle pagination
```
JanPetr: Can this block be separated to a method too?

Adapter/Algolia.php
```diff
@@ -90,90 +108,272 @@ public function __construct(
...
+        $page = !is_null($this->request->getParam('page')) ?
+            (int) $this->request->getParam('page') - 1 :
+            0;
+        $searchParams['page'] = $page;
+
+        // Handle category context
+        $category = $this->registry->registry('current_category');
```
JanPetr: Can this block be separated to a method too?

Plugin/RenderingCacheContextPlugin.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Plugin;
+
+use Algolia\AlgoliaSearch\Helper\ConfigHelper;
+use Magento\Framework\App\Http\Context as HttpContext;
+use Magento\Framework\App\Request\Http;
+use Magento\Store\Model\StoreManagerInterface;
+
+class RenderingCacheContextPlugin
```
JanPetr: Nice job, on the plugin! 👏 
I suppose I know what and how it's doing, but could you elaborate more on its' functionality and give examples how it kicks in when a user or robot sees the page? Thank you!

view/frontend/web/instantsearch.js
```diff
@@ -96,6 +92,92 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
 					$('.algolia-instant-replaced-content').hide();
 					$('.algolia-instant-selector-results').show();
 				}
+			},
+			// The url is now rendered as follows : http://website.com?q=searchquery&facet1=value&facet2=value1~value2
+			// "?" and "&" are used to be fetched easily inside Magento for the backend rendering
+			// Multivalued facets use "~" as separator
+			// Targeted index is defined by sortBy parameter
+			routing : {
```
JanPetr: Nice job! 🚀 
Can the `routing` parameter be defined on a different place of the extension e.g, common.js file?
It's only for a readability reasons as the routing definition is pretty long and I think it'll be easier to read it if it's extracted to separated method and only this new method is called here. WDYT?

Adapter/Algolia.php
```diff
@@ -90,90 +108,234 @@ public function __construct(
...
-        }
 
-        if ($useNative) {
-            $nativeQueryData = $this->getNativeQueryData($request);
-            $documents = $nativeQueryData['documents'];
-            $table = $nativeQueryData['table'];
+            $apiDocuments = array_map([$this, 'getApiDocument'], $documents);
+            $table = $temporaryStorage->storeApiDocuments($apiDocuments);
+        } catch (AlgoliaConnectionException $e) {
+            $this->nativeQuery($request);
```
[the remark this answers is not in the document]
damcou: I don't think so ... From what I understand, initially the goal was to use the native behaviour as fallback in case of error (?) : https://github.com/algolia/algoliasearch-magento-2/blob/master/Adapter/Algolia.php#L135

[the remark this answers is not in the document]
JanPetr: I get that, but on [line 131](https://github.com/algolia/algoliasearch-magento-2/pull/616/files/a0a5792df5cf8e094c1acd4b664203a77f21eba3#diff-83e13deff79b331f6a3ca0e7b1a6445aR131) there is `return $this->nativeQuery($request);` instead of just `$this->nativeQuery($request);`.
And the `return` makes sense as you don't want to probably continue in the method as the method is working with Algolia related data.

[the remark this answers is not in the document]
damcou: Oh you're right, sorry ;)

Helper/AdapterHelper.php
```diff
@@ -298,16 +104,16 @@ public function isAllowed()
         $storeId = $this->getStoreId();
 
         return
-            $this->config->getApplicationID($storeId)
-            && $this->config->getAPIKey($storeId)
-            && $this->config->isEnabledFrontEnd($storeId)
-            && $this->config->makeSeoRequest($storeId);
+            $this->algoliaHelper->getConfigHelper()->getApplicationID($storeId)
```
JanPetr: Provision of `configHelper` shouldn't be a responsibility of `algoliaHelper`. `AlgoliaHelper` should handle only Algolia-related stuff, so I don't like that it provisions `configHelper` here. The `__constructor` has 3 arguments now, which is great, but we need to think about readability and what helpers are responsible for what.

I'd much rather inject `ConfigHelper` here or via `Magento\CatalogSearch\Helper\Data` class, which makes more sense as the `CatalogSearch helper` is the class responsible for the correct search in general. WDYT?

Helper/AdapterHelper.php
```diff
@@ -298,16 +104,16 @@ public function isAllowed()
...
+            && $this->algoliaHelper->getConfigHelper()->getAPIKey($storeId)
+            && $this->algoliaHelper->getConfigHelper()->isEnabledFrontEnd($storeId)
+            && $this->algoliaHelper->getConfigHelper()->makeSeoRequest($storeId);
     }
 
     /** @return bool */
     public function isSearch()
     {
-        return $this->request->getFullActionName() === 'catalogsearch_result_index';
+        return $this->filtersHelper->getRequest()->getFullActionName() === 'catalogsearch_result_index';
```
JanPetr: The same as before applies here as well. `filterHelper` should be responsible only to count / prepare filters, but not provision `request`. WDYT of having `CatalogSearch helper` to provision it? Or do you have any other idea?

PR #620, opened 2018-11-21
Admin Analytics Overview

bsuravech:
**Summary**

Replicate the Analytics Overview section of the Algolia Dashboard for Magento 2 using Algolia's Analytics API. 

**Result**
- Create new Analytics Overview menu item
- AnalyticsHelper to handle Analytics API endpoints
- Controllers for overview page and asynchronous form update
- Use store selector and create custom form to handle daterange and index type
- Use Google Charts API v2 for render interactive graph for w/ tooltip breakdown
- Added generic tooltip template based off of Magento's form tooltip


Controller/Adminhtml/Analytics/Update.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    public function execute()
+    {
+        $response = $this->_objectManager->create(\Magento\Framework\DataObject::class);
+        $response->setError(false);
+
+        $this->_getSession()->setAlgoliaAnalyticsFormData($this->getRequest()->getParams());
+
+        $layout = $this->layoutFactory->create();
+        $block = $layout->createBlock('\Algolia\AlgoliaSearch\Block\Adminhtml\Analytics\Index')
```
JanPetr: Can you use `::class` constant here instead of `'\Algolia\AlgoliaSearch\Block\Adminhtml\Analytics\Index'`? :)

Helper/AnalyticsHelper.php
```diff
@@ -0,0 +1,380 @@
...
+        }
+
+        return $this->clientData;
+    }
+
+    public function isAnalyticsApiEnabled()
+    {
+        $clientData = $this->getClientData();
+
+        return $clientData && isset($clientData['analytics_api']) ? $clientData['analytics_api'] : 0;
```
JanPetr: Can you return `bool` value from here?

Helper/AnalyticsHelper.php
```diff
@@ -0,0 +1,380 @@
...
+
+    public function isClickAnalyticsEnabled()
+    {
+        if (!$this->configHelper->isClickConversionAnalyticsEnabled()) {
+            return false;
+        }
+
+        $clientData = $this->getClientData();
+
+        return $clientData && isset($clientData['click_analytics']) ? $clientData['click_analytics'] : 0;
```
JanPetr: Can you return `bool` value from here? Now the method might return `bool` / `int` / `string` and it's a good practice to strictly return only one type out of a method.

Helper/AnalyticsHelper.php
```diff
@@ -0,0 +1,380 @@
...
+     * @param Data $dataHelper
+     * @param ProductHelper $productHelper
+     * @param CategoryHelper $categoryHelper
+     * @param PageHelper $pageHelper
+     * @param Logger $logger
+     */
+    public function __construct(
+        AlgoliaHelper $algoliaHelper,
+        ConfigHelper $configHelper,
+        Data $dataHelper,
```
JanPetr: `DataHelper`, `ProductHelper`, `CategoryHelper`, `PageHelper` are injected only to provision index names. How about extract it to a standalone `AnalyticsIndexNameProvider` (just example name) class and lower number of dependencies here? :)

Helper/AnalyticsHelper.php
```diff
@@ -0,0 +1,380 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Helper;
+
+use Algolia\AlgoliaSearch\Helper\Entity\CategoryHelper;
+use Algolia\AlgoliaSearch\Helper\Entity\PageHelper;
+use Algolia\AlgoliaSearch\Helper\Entity\ProductHelper;
+use AlgoliaSearch\Analytics;
+
+class AnalyticsHelper extends Analytics
```
JanPetr: How about use `algoliaHelper` to "create" and get instance of `AlgoliaSearch\Analytics`?
If I didn't overlook something, this class uses only `request()` method of the `AlgoliaSearch\Analytics` and I feel it's not needed to extend the class completely.

By switching the responsibility of creation of `AlgoliaSearch\Analytics` to `AlgoliaHelper` we keep the right responsibilities in right classes. What do you think of that?
At the same time we'd reduce coupling with Algolia API client in general.

Block/Adminhtml/Analytics/Index.php
```diff
@@ -0,0 +1,422 @@
...
+     * @param PageCollection $pageCollection
+     * @param array $data
+     */
+    public function __construct(
+        Context $context,
+        AnalyticsHelper $analyticsHelper,
+        TimezoneInterface $dateTime,
+        ProductCollection $productCollection,
+        CategoryCollection $categoryCollection,
+        PageCollection $pageCollection,
```
JanPetr: `ProductCollection`, `CategoryCollection`, `PageCollection` are injected to enrich / hydrate popular results in `getPopularResults()` method. How about extracting this logic to a standalone class (like `PopularResultsDataHydrator` :)) where the logic can sit and you could have a single method for each data type (product, category, page)? WDYT?

etc/adminhtml/menu.xml
```diff
@@ -54,5 +54,14 @@
...
+            id="Algolia_AlgoliaSearch::analytics"
+            title="Analytics Overview"
+            module="Algolia_AlgoliaSearch"
+            sortOrder="110"
+            parent="Algolia_AlgoliaSearch::manage"
+            action="algolia_algoliasearch/analytics/index"
+            resource="Algolia_AlgoliaSearch::manage"
+        />
     </menu>
 </config>
```
JanPetr: Could you add "empty" line to each file ending like that? :) It's not super useful, but looks better :)))

Helper/AnalyticsHelper.php
bsuravech: Admittedly, I just tried to create an instance of AlgoliaSearch\Analytics but then was remembered why I extended the API in the first place. The request() method is protected and not accessible outside of the class.

JanPetr: /o\ Makes sense then, I had no idea the Analytics class has this limitation. All good then 👍 :)

PR #629, opened 2018-12-12
Refactor ObjectManager into standalone DataProvider class

JanPetr:


JanPetr:
Actually I don't think `DataProviders` are M2 thing - I just created the folder and class, without sticking into any M2 standard :))

Issue #635, opened 2018-12-13
Facet item template - different id in phtml and js 


vmalyk:
**Do you want to request a *feature* or report a *bug*?**
Bug

**Bug: What is the current behavior?**
Template for frontend frontend/templates/instant/facet.phtml isn't used for facets and impossible to customize it. In template *id="facet-template"*. But in instantsearch.js in [line 482](https://github.com/algolia/algoliasearch-magento-2/blob/master/view/frontend/web/instantsearch.js#L482) used *id="'refinements-lists-item-template'"*

**Bug: What is the expected behavior?**
IDs should be equal and possible to customize template from theme instead rendering by default template from instantsearch.js library

**Bug: What is the proposed solution?**
Change id in  frontend/templates/instant/facet.phtml to *id="'refinements-lists-item-template'"*. It`ll be the same as in [Magento 1 module ](https://github.com/algolia/algoliasearch-magento/blob/master/app/design/frontend/base/default/template/algoliasearch/instantsearch/refinementsItem.phtml)

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**

Magento 2.2.6
Algolia 1.8.3 (looks as it`s relevant for all previous releases)

PR #636, opened 2018-12-14
Fixed id for facet.phtml template. Added classes from widget confgura…

vmalyk:
Pull request for fix issue #635


JanPetr:
Great catch, @vmalyk! 👏 

[Thank you](http://bit.ly/2KHTYV3), for the contribution 💯🚀

PR #645, opened 2018-12-20
Avoid duplicated entries to process

alepane21:
**Summary**

When doing mass import of product changes, lot of duplicated rows are created in the job queue.
I know that there is a merge done before the processing, but still this change reduce the indexing operations in some edge case not fixed from the current merge.

**Result**

With 24 products changed we go from 162 rows to 46.

### DIFF
--- a/Model/Queue.php
+++ b/Model/Queue.php
@@ -64,6 +64,23 @@ class Queue
         $this->maxSingleJobDataSize = $this->configHelper->getNumberOfElementByPage();
     }
 
+    /**
+     * Check if there is already an item in the queue with the same values
+     */
+    public function alreadyInQueueToProcess($className, $method, $encodedData, $data_size = 1) {
+        $data = $this->db->query(
+            $this->db->select()
+                ->from($this->table, ['jobs_count' => 'COUNT(job_id)'])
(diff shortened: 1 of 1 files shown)


alepane21:
If someone is interested, we have this pull request as patch here: https://github.com/Hevelop/magento2-patches/blob/master/Patch-Algolia-Reduce-Operations.patch

JanPetr:
Hello @alepane21,

thank you for the contribution 💯 Unfortunately it is not that simple :(
Imagine situation when you'd have "update", "delete" and "update". Your approach would not enter the second update, because it's already in the queue and after the queue processing, the object would be removed from Algolia, even though it's supposed to be indexed.

Currently it's not issue for "products", as products don't have "delete" job, but categories do.
In products the "moveIndex" job needs to be taken into account.

We plan to narrow and optimize the indexing pipeline in the next months, so if you don't mind, I'll keep your PR in mind and incorporate it into the new pipeline. Would it work for you? 🙏

alepane21:
Oh, I see, yes of course it is ok. Obviously I only tried with products...
But the delete/update problems is only when moving the categories?
If so, I think I will keep the patch in our environments.

Thanks for your help!

PR #665, opened 2019-01-24
Optimize default indexing queue settings

JanPetr:
Each job to run has overhead with fetching data from database. 

Making it run less jobs (10 -> 5) with bigger page size (100 -> 300) reduces the overhead and improves the indexing performance, especially for simple products.


JanPetr:
Definitely, I marked it "for review" too soon.

PR #666, opened 2019-01-29
Subscribe to price index changes - reindex products when catalog rule is applied

unicoder88:
Hi there, how is it going?

Our customers too often complain that when they start new campaign with catalog rules, Algolia prices are not updated, and ask us to reindex all manually. That's it! We need to do something about it :)

One difficulty with catalog rules is that it's sometimes difficult to know when some particulat product is affected by catalog rule. After some research I found that after catalog rules reindex, catalog price index is always called. This means that we can subscribe to `catalog_product_index_price` table and get Algolia to reindex when new rules take effect.

We keep all our indexers in update by schedule mode and Algolia queue runner enabled, so this approach should work.

## Steps to reproduce

1. Let's test on simple products, default Magento 2.2.7. Bag initial price 45 NOK. https://prnt.sc/mdmk3d
1. Set all indexes in update by schedule mode, enable queue runner. https://prnt.sc/mdmkd1
1. Create catalog rule 10% e.g. for category Bags https://prnt.sc/mdmkrn
1. Run cron to update indexes
1. Run Algolia queue runner index

Expected:
1. Discounted price in Algolia

Actual:
1. Changed products not added to `algolia_products_cl` table
1. Job not added to Algolia queue
1. Price in Algolia is outdated https://prnt.sc/mdmls3

When I added new mview subscription and ran setup:upgrade, triggers got created https://prnt.sc/mdmgjy .
Now trying to turn the rule off and on again, running cron:run and reindex queue (a few times) - and new price is there! https://prnt.sc/mdmm0y https://prnt.sc/mdmmaz

P.S. I checked since when catalog rules started automatically influencing price index - since Magento 2.2.5:

* price index modifier https://github.com/magento/magento2/blob/2.2.5/app/code/Magento/CatalogRule/Model/Indexer/ProductPriceIndexModifier.php
* indexer dependency https://github.com/magento/magento2/blob/2.2.5/app/code/Magento/CatalogRule/etc/indexer.xml#L18

Looking forward to your thoughts :)

### DIFF
--- a/etc/mview.xml
+++ b/etc/mview.xml
@@ -35,6 +35,7 @@
             <table name="catalog_product_website" entity_column="product_id" />
             <table name="catalog_category_product" entity_column="product_id" />
             <table name="cataloginventory_stock_item" entity_column="product_id" />
+            <table name="catalog_product_index_price" entity_column="entity_id" />
         </subscriptions>
     </view>
 </config>


unicoder88:
Oh I hope this PR isn't cursed

rliwoch:
Apologies for digging out this PR, may I have a quick question, sorry if what I'm about to ask about is well know but I'm not a magento2 guy, I've only been asked by someone to help investigate an issue. 
It's well understood that algolia_queue cron job has to be scheduled - the manual talks about that. So I have it scheduled every 5 minutes. All algolia cron jobs are setup to run on schedule, the only additional job in cron is a default `magento cron:run` and unfortunately I can't see the queue being populated after my catalog rule change.

Now, I came across this PR and step 2) in the reproduce section may suggest that I should have all algolia's indexers found in `mview.xml` setup in my `crontab`.
Could you please confirm my understanding is correct? 
If yes, what is a suggested interval for these indexers?

Apologies for a naive question but it's not explicitly called out in the manual (I didn't find anything) and I was hoping that algolia module has a listener built in for any save operations and indexer:reindex is only used if a user wishes to rebuild the index completely. 

BTW. We're on 1.10.0 so according to release notes this change should be incorporated there. 

Thank you!

damcou:
Hello @rliwoch ,

Thanks for reaching out.
Could you please tell me if the queue if not populated only after a catalogrule change or after any event (product save, category save, etc ... ) ? And of course, if you activated the indexing queue inside the configuration of the extension ?

If I understood correctly, you set all your indexer to run "on schedule", that mean that if you want to populate the queue with jobs, you have to run bin/magento indexer:reindex [indexer] to do so.
We have observers that are triggers on save events for products and categories, but you have to set the indexer as "on save" to make them work.

Hope that helps.
Best.

rliwoch:
@damcou Damien, first of all thank you for taking the time to answer my question. That helps a lot and doesn't force me to go and read the code!

Yes - of course, the queue is activated in the extension config.
As to the question if it's not updated at all - I think that's the case, I seem to remember that the queue was sitting at 0 messages all the time. I'll confirm that. However I think the second part of your answer sheds light onto what the problem is.

Previously - someone set ALL Algolia's indexers `on Save`, which didn't seem right because, well, queue is on schedule. So, in my battle-cry, without putting too much thought into it - I have changed all algolia's indexers to `on schedule`. Clearly that's the same calibre of a mistake.

If I understand correctly, the observers should work correctly (even for the catalog rule) if:
* algolia queue indexer is `on schedule` 
* all other algolia's indexers (product,categories,pages) are set `on save`

If I don't want something to be automatically put on the queue on the `save` op then I should simply configure the particular indexer to `schedule` and add a crontab entry. Simples!

I'll confirm if that works shortly!

Thank you very much,
Raf

rliwoch:
@damcou I got too excited. Unfortunately
* algolia queue indexer is `on schedule`
* all other algolia's indexers (product,categories,pages) are set `on save`
* crontab `algolia_queue_runner` every 5 minutes

Queue enabled in module config. 

Scenario 1: 
entry conditions as above
1) change product name by adding one character 
2a) magento indexers kick in
2b) messages on algolia queue immediately
3) awaiting algolia queue cron run
4) algolia runner runs 
5) logs on server side visible: 
``` 
10 jobs processed in 3 seconds.
Algolia Search Queue Runner index has been rebuilt successfully in 00:00:03
``` 
6) browsing category page - I can see product name changed.

Scenario 1: 
entry conditions as above
1) Apply catalog rule on a product typ
2a) magento indexers kick in
2b) *nothing* on algolia queue
3) check frontend *product page* (not coming from algolia) - discounted price visible
4) awaiting algolia queue cron run
5) algolia runner runs 
6) logs on server side visible: 
``` 
 0 jobs processed in 0 seconds.
Algolia Search Queue Runner index has been rebuilt successfully in 00:00:01
``` 
6) browsing category page (data from algolia) - I *cannot*  see discounted price


Indexers setup:
image

PR #670, opened 2019-02-05
M2.1 JS Minify Issue

bsuravech:
**Summary**
Support Ticket #117914 

There is an issue in M2.1 when javascript is minified and common.js is throwing a syntax error. This is because M2.1's JShrink\Minifier does not consider backtick template literals a string type and thinks the // in the template literal, still a comment line. This is corrected in M2.2. 

https://github.com/algolia/algoliasearch-magento-2/blob/master/view/frontend/web/internals/common.js#L569-L571

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

**Result**
- Updated return string type to be accepted in 2.1

### DIFF
--- a/view/frontend/web/internals/common.js
+++ b/view/frontend/web/internals/common.js
@@ -563,12 +563,12 @@ requirejs(['algoliaBundle'], function(algoliaBundle) {
 				createURL({ qsModule, routeState, location }) {
 					const { protocol, hostname, port = '', pathname, hash } = location;
 					const queryString = qsModule.stringify(routeState);
-					const portWithPrefix = port === '' ? '' : `:${port}`;
+					const portWithPrefix = port === '' ? '' : ':' + port;
 					// IE <= 11 has no location.origin or buggy. Therefore we don't rely on it
 					if (!routeState || Object.keys(routeState).length === 0)
-                        return protocol + '//' + hostname + portWithPrefix + pathname;
+						return protocol + '//' + hostname + portWithPrefix + pathname;
 					else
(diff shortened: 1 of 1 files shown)


view/frontend/web/internals/common.js
```diff
@@ -566,9 +566,9 @@ requirejs(['algoliaBundle'], function(algoliaBundle) {
 					const portWithPrefix = port === '' ? '' : `:${port}`;
```
JanPetr: Could you fix this "backticked" string as well? :)

view/frontend/web/internals/common.js
```diff
@@ -566,9 +566,9 @@ requirejs(['algoliaBundle'], function(algoliaBundle) {
...
 					// IE <= 11 has no location.origin or buggy. Therefore we don't rely on it
 					if (!routeState || Object.keys(routeState).length === 0)
-						return `${protocol}//${hostname}${portWithPrefix}${pathname}`;
+                        return protocol + '//' + hostname + portWithPrefix + pathname;
```
JanPetr: In JS files we use tabs instead of spaces (I know, there is an inconsistency), could you fix that? 🙏

view/frontend/web/internals/common.js
bsuravech: You got it!

view/frontend/web/internals/common.js
bsuravech: tabs tabs tabs!

Issue #677, opened 2019-02-13
AlgoliaSearch v1.9.1 error on IE11


BanyuUrip:
**Do you want to request a *feature* or report a *bug*?**
_Bug._
Steps to reproduce:

- Open https://magento2.algolia.com/women/tops-women/jackets-women.html on IE 11 

**Bug: What is the current behavior?**

![magento2_algoliasearch_v191_error_ie11](https://user-images.githubusercontent.com/6737571/52693500-79b64180-2f99-11e9-87dd-b234c69eabd8.jpg)

**Bug: What is the expected behavior?**
Product list page showed properly.

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**

Algolia v1.9.1 for Magento 2

JanPetr:
Hello @BanyuUrip, thank you for the report! 
It was fixed in the PR^. Feel free to give it a try 🙌

BanyuUrip:
Hi @JanPetr,

Thank you for the reply.

I have tried the PR but still had a problem with the https://github.com/algolia/algoliasearch-magento-2/pull/678/commits/1da9d8f641225b8287d6abdb59a249916e2bc3ea. When I applied the line code below, the product list page still doesn't work including on Chrome and Firefox.

`<script src="//polyfill.io/v3/polyfill.min.js?features=default%2CArray.prototype.includes%2CPromise" src_type="url" crossorigin/>`

So I try to add "anonymous" to crossorigin attribute, and it works fine now (Chrome, Firefox and IE11).

`<script src="//polyfill.io/v3/polyfill.min.js?features=default%2CArray.prototype.includes%2CPromise" src_type="url" crossorigin="anonymous"/>`

Thanks

Issue #659, opened 2019-01-18
Algolia Instant Search doesn`t work on IE11


vmalyk:
**Bug: What is the current behavior?**

Algolia Instant search not rendered and error shows in console

SCRIPT1003: Expected ':'  common.js(556, 13)

**Bug: What is the expected behavior?**
Algolia Instant search should be work in IE 11, because this version in offical included to Magento 2 supported browser list: https://devdocs.magento.com/guides/v2.3/install-gde/system-requirements_browsers.html 

**Bug: What is the proposed solution?**
Issue is related with routing  for Instant search pages (URLs are SEO-friendly) and it's implemented with ES6 specification and it's not supported in IE11. In this case -  shorter syntax for method definitions on objects initializers. So, need to restore IE11 support in JS files.

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**

- Magento 2.2.3-2.2.6
- AlgoliaSearch 1.9.0

mnjongerius:
Hi @vmalyk , we found the same issue on IE11. I'm willing to look into fixing this myself, can you point me in the right direction?

JanPetr:
Fixed in PR^ 🚀 
[Thank you](http://bit.ly/2KHTYV3) for the report, @vmalyk and @mnjongerius 💯

robbertstevens:
@JanPetr when wil a new version release, so I can update the module?

PR #678, opened 2019-02-13
Update JS for IE11 compatibility 

bsuravech:
**Summary**
IE11 has some issues with ES6 syntaxes for method definition shorthand and destructuring assignments. You can read more about it here:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

Related Issues: #677, #659 

**Result**
Converted syntax into IE11 friendly versions. Manually assigning variables instead of the shorthand.


bsuravech:
@JanPetr Our extension is good under IE11. There is still that Promise error that I think is on instantsearch.js which recommends using a polyfill.io or babel.

choussamaster:
Hello @bsuravech what's the utility of adding the polyfill js file ? he seems empty thank you
https://polyfill.io/v3/polyfill.min.js?features=default%2CArray.prototype.includes%2CPromise

PR #681, opened 2019-02-14
Fix empty categories variable resulting in count on null error

### LINKED ISSUE #673: Reindex "algolia_categories" Error
BUG

If I reindex: `php bin/magento indexer:reindex algolia_categories`

i get `Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/html/magento2/vendor/algolia/algoliasearch-magento-2/Model/Indexer/Category.php on line 91`

image

indexer finishes successful

no idea... :(

Magento 2.3
php7.2
"name": "algolia/algoliasearch-magento-2",
"version": "1.9.1"

mwr:
During a full reindex there are no category-IDs therefore the variable $categories is null.

**Summary**

This pull request initializes an empty array so the rest of the indexer works.

**Result**

Indexer does not throw exceptions and reindexes the categories.

Edit: Fixes #673


Model/Indexer/Category.php
```diff
@@ -82,6 +82,7 @@ public function execute($categoryIds)
                 );
             } else {
                 $this->queue->addToQueue($this->fullAction, 'saveConfigurationToAlgolia', ['store_id' => $storeId], 1);
+                $categoryIds = [];
```
JanPetr: Adding condition on line 66 like:
```php
if ($categoryIds === null) {
    $categoryIds = [];
}
```
and changing the `if ($categoryIds !== null)` conditon would be much more readable. What do you think, @mwr?

mwr: Do you mean before the `if ($categoryIds !== null) ` in line 75?

If it is before the if in line 75 it would change the behaviour since the deleteObjects queue entry would also be added. 
With the solution above if categoryIds === null the deleteObjects will not be queued.

After the if, I see no problem in adding it there.

JanPetr: I know, my idea was to change the condition as well to still work consistently.

But I'm wondering - the `null` is issues only for line 92: `count($categoryIds)`. How about just changing this line to:

```php
is_array($categoryIds) ? count($categoryIds) : 1
```

I think that'd be better for readability as well.

mwr: ok, now I understand what you mean.
Sure that is definitely easier to read and understand.

JanPetr: Cool! I'll wait for your commit then :) Or let me know if you want me to do it :)

mwr: Feel free to do it. I won't get to this until thursday

PR #700, opened 2019-02-26
Add native price facet handling

damcou:


Helper/Adapter/FiltersHelper.php
```diff
@@ -197,11 +197,20 @@ public function getPriceFilters($storeId)
...
             $prices = explode(':', $pricesFilter);
+        }
+
+        // Native Magento facet compatibility
+        if (!$this->config->isInstantEnabled($storeId) && !is_null($this->request->getParam('price'))) {
+            $pricesFilter = $this->request->getParam('price');
+            $prices = explode('-', $pricesFilter);
+        }
 
+        if (count($prices) > 0) {
```
JanPetr: When `count($prices)` is 1, nothing will happen. Is that expected? Is this condition still needed when the condition below is `count($prices) == 2`?

Helper/Adapter/FiltersHelper.php
```diff
@@ -197,11 +197,20 @@ public function getPriceFilters($storeId)
         $paramPriceSliderData = $this->getParamPriceSlider($storeId);
         $priceSlider = $paramPriceSliderData['price_slider'];
         $paramPriceSlider = $paramPriceSliderData['param'];
+        $prices = [];
 
         if (!is_null($this->request->getParam($paramPriceSlider))) {
```
JanPetr: Could you maybe add comment that this condition parses prices for IS.js generated URL? (if it's correct)

Issue #701, opened 2019-02-26
"Add to cart" translation not working in Algolia Instant Search Page


aboutzahurul:
We are unable to translate "Add to cart" text in Algolia Instant Search Page

### Preconditions
1. Magento CE 2.1.9 with sample product, Multi language en_US and ar_SA
2. algolia/algoliasearch-magento-2 1.10.0 installed by composer (you will need algolia account to config the module)
3. On i18n en_US.csv translated to ar_SA.csv

### Expected result
"Add to cart" text should translated as in AR csv on Algolia Instant Search Page

### Actual result
"Add to cart" text didn't translated

### Solution
on file 'vendor/algolia/algoliasearch-magento-2/view/frontend/templates/instant/hit.phtml' line 71 "Add to cart" text didn't added to translation.

replace line 71 code by below code.
```
<span><?php echo __('Add to Cart') ?></span>
```
Note: Its not recommenced to replace original module file cos change may lost in future update. for this you may follow the link "https://www.algolia.com/doc/integration/magento-2/guides/create-a-custom-extension/" to know how to override algolia magento 2 module 

@algolia:  Please fix it on next version.

bsuravech:
Hi @aboutzahurul,

Good catch! We have created a PR #702 to address this issue. You can take a look at it and patch accordingly. Once approved, it will be included for our next release. Thanks!

aboutzahurul:
Hi @bsuravech,

We were also finding difficulties to translate the text "Show more products" when turning on the infinite scroll for Algolia Instant Search Page. After long search we got the text on "https://github.com/algolia/algoliasearch-magento-2/blob/master/Block/Configuration.php" under line number 229. This text also not added on en_US.csv. I just manually added the text on csv and working file. Please regenerate the translation text and solve the issue also on next release. Thanks!

aboutzahurul:
> Hi @aboutzahurul,
> 
> Good catch! We have created a PR #702 to address this issue. You can take a look at it and patch accordingly. Once approved, it will be included for our next release. Thanks!



> We are unable to translate "Add to cart" text in Algolia Instant Search Page
> 
> ### Preconditions
> 1. Magento CE 2.1.9 with sample product, Multi language en_US and ar_SA
> 2. algolia/algoliasearch-magento-2 1.10.0 installed by composer (you will need algolia account to config the module)
> 3. On i18n en_US.csv translated to ar_SA.csv
> 
> ### Expected result
[...]

PR #702, opened 2019-02-26
Translate html text in hit template

bsuravech:
**Summary**
Translate text in template for i18n translation. 

Related Issue: #701 

**Result**
Update the hit template for translation and review other templates for any text that hasn't been translated.

### DIFF
--- a/view/frontend/templates/instant/hit.phtml
+++ b/view/frontend/templates/instant/hit.phtml
@@ -26,7 +26,7 @@ $tierFormatedVar = 'price' . $priceKey . '_tier_formated'
                             <div class="result-sub-content">
                                 <div class="product-reviews-summary short">
                                     <div class="rating-summary">
-                                        <span class="label"><span>Rating:</span></span>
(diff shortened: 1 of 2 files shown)


view/frontend/templates/instant/hit.phtml
```diff
@@ -26,7 +26,7 @@ $tierFormatedVar = 'price' . $priceKey . '_tier_formated'
                             <div class="result-sub-content">
                                 <div class="product-reviews-summary short">
                                     <div class="rating-summary">
-                                        <span class="label"><span>Rating:</span></span>
+                                        <span class="label"><span><?php echo __('Rating') ?>:</span></span>
```
JanPetr: I'm checking the `en_US.csv` file and "Rating" is not translated there. Can you add it? 🙏

bsuravech: I added Rating and As low as! Thanks for the catch!

PR #714, opened 2019-03-13
Sales Data - change getAllIds() to getColumnValues() for optimization

bsuravech:
**Summary**
$collection->getAllIds() unsets the select query and retrieves the entire collection IDs. For the getSalesData() method, we should filter the collection IDs down to the current page and limit in the query. Use getColumnValues() to retrieve all entity_ids of the queried product collection.

**Result**
Optimized select query for getSalesData();

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -707,7 +707,7 @@ class Data
 
         $ordersTableName = $this->resource->getTableName('sales_order_item');
 
-        $ids = $collection->getAllIds();
+        $ids = $collection->getColumnValues('entity_id');
         $ids[] = '0'; // Makes sure the imploded string is not empty
 
         $ids = implode(', ', $ids);


JanPetr:
Thank you, @bsuravech!
I just remembered - does `entity_id` column apply to all Magento versions and editions? I know we had issues with `enitity_id` vs. `row_id` column on Enterprise versions and staging module enabled: https://github.com/algolia/algoliasearch-magento-2/blob/master/Helper/Entity/CategoryHelper.php#L576-L592

bsuravech:
@JanPetr I thought `row_id` was associated to Magento_Staging (Commerce only) and is a temporary ID for previewing staged updates. `sales_order_item` should be saving entity_id of the product for product_id. `entity_id` should still be available as the product ID.

PR #717, opened 2019-03-19
Indexing Queue export button

bsuravech:
**Summary**
For troubleshooting, it might be helpful to have a button to export the indexing queue table so that the client can send over upon request for those who do not have DB access.

**Result**
Added a button to the indexing queue page.


JanPetr:
Hey @bsuravech, thanks for that!

I'm just lacking Magento knowledge here - Magento itself is handling the export and no additional code in the extension has to be added? :)

bsuravech:
@JanPetr I'm familiar with the M1 way of adding an export option but for M2 it seems to be the same logic where it exports the grid collection. You don't need to supply additional code for it. I would try it on your local to see how it works. I was amazed at the single line I added myself :)

PR #721, opened 2019-03-22
Replace all hashes in search query with encoded characters

### LINKED ISSUE #720: Form submit with hash in search query works not correctly
Bug

If there is a hash in the search query, the url is not interpreted correctly by the browser when the form is submitted.

So the search does only search for `shirt` instead of `shirt #281` in my example.

```
https://shop.local/catalogsearch/result/?q=shirt%20#281
```

The hash gets encoded and the search includes the complete search query.

```
https://shop.local/catalogsearch/result/?q=shirt%20#281
```

Replacing all `#` in the search query with `%23`. 

To be one hundred percent sure, we could also `encodeURI()` the complete search query, then all special characters would be encoded.

Algolia 1.10.0
Magento: 2.2.7

marvinhuebner:
**Summary**
Replaces all `#` in search query with `%23` on form submit.

**Result**
If the search query contains a hash, the search query get correctly interpreted.

Resolves #720

### DIFF
--- a/view/frontend/web/internals/common.js
+++ b/view/frontend/web/internals/common.js
@@ -443,6 +443,8 @@ requirejs(['algoliaBundle'], function(algoliaBundle) {
 		$(algoliaConfig.autocomplete.selector).each(function () {
 			$(this).closest('form').submit(function (e) {
 				var query = $(this).find(algoliaConfig.autocomplete.selector).val();
+				
+				query = query.replace(/#/g, '%23');
 
 				if (algoliaConfig.instant.enabled && query === '')
 					query = '__empty__';


view/frontend/web/internals/common.js
```diff
@@ -443,6 +443,8 @@ requirejs(['algoliaBundle'], function(algoliaBundle) {
 		$(algoliaConfig.autocomplete.selector).each(function () {
 			$(this).closest('form').submit(function (e) {
 				var query = $(this).find(algoliaConfig.autocomplete.selector).val();
+				
+				query = query.replace(/#/g, '%23');
```
marvinhuebner: As already proposed in #720, to be one hundred percent sure, we could also `encodeURI()` the complete search query, then all special characters would be encoded.

```
query = encodeURI(query);
query = query.replace(/#/g, '%23');
```

That would have the consequence if you search in german for example for "küchentuch #2349" you would have `?q=k%C3%BCchentuch%20%232349` instead of `?q=küchentuch%20%232349`.

marvinhuebner:
Looks like the failed tests have nothing to do with my adjustments, am I right?

JanPetr:
Thank you for the PR, @marvinhuebner 🙌 
I'll take a look and come back to you :)

> Looks like the failed tests have nothing to do with my adjustments, am I right?

Yes, it's because of "secured" variables in Travis, no worries about that.

JanPetr:
Great catch @marvinhuebner, thanks you for spotting that and pushing a PR 💯 
But indeed, I'd go rather for full `encodeURI()` method to make sure all the browsers are able to handle that. Would you be able to change that? :) Thank you!

marvinhuebner:
Thanks @JanPetr! 

I've updated my PR. I have now used `encodeURIComponent()` instead of `encodeURI()`. This looks even better and I was able to remove the extra replace function for hashes.

PR #722, opened 2019-03-22
Use factory to create new AlgoliaSearch\Client

marcusirgens:
This will allow interceptor generation, and is considered best practice.

**Summary**
Use Factory to get the Client instead "new".

**Result**
You can now make plugins.


JanPetr:
Great catch and job, @marcusirgens! I'm happy to merge 🙌 

[Thank you](http://bit.ly/2KHTYV3)!

haiacom:
This doesn't work on Magento 2.1.x, see https://github.com/algolia/algoliasearch-magento-2/issues/842

Issue #668, opened 2019-01-30
Customer group prices are always "0" for grouped products


eugene-shramko:
Magento 2.3 EE
Algolia 1.9.1

1. Configure Algolia indexes to be updated on save ("System" -> "Index Management").
2. Configure Algolia to index customer group prices (set "Stores" -> "Configuration" -> "Algolia Search" -> "Advanced" -> "Enable Customer Groups" to "Yes").
3. Create one simple product with regular price set to "100".
4. Create another simple product with regular price set to "150".
5. Create a grouped product, select created simple products as the child products.
6. Check created grouped product data within Algolia indices. Check autocomplete results for grouped product.

**Expected result:**
All customer group prices for grouped product within Algolia indices correspond to Magento.
The price for grouped product in autocomplete correspond to Magento.

**Actual result:**
All customer group prices for grouped product within Algolia indices are "0".
The price for grouped product in autocomplete is "0".

**P.S.** Default prices for grouped product are calculated correctly.

JanPetr:
Hello @eugene-shramko,

thank you for the report and I'm sorry for the late answer.
We took the reported issue back internally and we'll investigate soon and send an update here.

If you have more information or details, feel free to share it with us here, it'll help us to fix it. Thank you!

JanPetr:
Fixed, it'll be part of the next release 🎉

PR #726, opened 2019-03-25
Fix customer group prices on grouped products

bsuravech:
**Summary**
[MAG-136] Customer group pricing on grouped products are returning 0 instead of min/max values. 

Related issue: #668 

**Result**
Cast type on group price for type comparison in setFinalGroupPrices();

[MAG-136]: https://algolia.atlassian.net/browse/MAG-136

### DIFF
--- a/Helper/Entity/Product/PriceManager/ProductWithChildren.php
+++ b/Helper/Entity/Product/PriceManager/ProductWithChildren.php
@@ -105,7 +105,7 @@ abstract class ProductWithChildren extends ProductWithoutChildren
         foreach ($this->groups as $group) {
             $groupId = (int) $group->getData('customer_group_id');
 
-            if ((int) $this->customData[$field][$currencyCode]['group_' . $groupId] === 0) {
+            if ($this->customData[$field][$currencyCode]['group_' . $groupId] == 0) {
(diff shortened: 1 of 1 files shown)


Helper/Entity/Product/PriceManager/ProductWithChildren.php
```diff
@@ -105,7 +105,7 @@ protected function setFinalGroupPrices($field, $currencyCode, $min, $max, $dashe
         foreach ($this->groups as $group) {
             $groupId = (int) $group->getData('customer_group_id');
 
-            if ($this->customData[$field][$currencyCode]['group_' . $groupId] === 0) {
+            if ((int) $this->customData[$field][$currencyCode]['group_' . $groupId] === 0) {
```
JanPetr: Is the `(int)` cast correct? What if the price is `float`, lika `35,76`? Won't it change it to `35` for the future use and therefore it'd index wrong value?

bsuravech: Ah! You're right, a float would be more appropriate though to be honest, I'm not sure why we're zero checking here instead of just applying the final group prices.

Issue #615, opened 2018-11-15
Price not shown in locale format


peterjaap:
**Do you want to request a *feature* or report a *bug*?**
Bug

**Bug: What is the current behavior?**
After saving a product, the Algolia index for the German storeview does not contain the German-locale formatted price (`24,95 €`), but the default storeview formatted price (`€24,95`);

![image](https://user-images.githubusercontent.com/431360/48564232-47dc0300-e8f6-11e8-8016-8343355818e8.png)

**Bug: What is the expected behavior?**
It should contain the German-locale formatted price (with the euro sign behind the price instead of in front of it), like this;

![image](https://user-images.githubusercontent.com/431360/48564274-56c2b580-e8f6-11e8-91f0-4fb740b7d6a4.png)

**Bug: What is the proposed solution?**
Not sure. Since [the code](https://github.com/algolia/algoliasearch-magento-2/blob/master/Helper/Entity/Product/PriceManager.php#L109) does try to fetch it through the priceCurrency `format` function;

```
                $customData[$field][$currencyCode]['default_formated'] = $this->priceCurrency->format(
                    $price,
                    false,
                    PriceCurrencyInterface::DEFAULT_PRECISION,
                    $store,
                    $currencyCode
                );
```

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**

Magento 2.2.6, Algolia 1.7.1

peterjaap:
Hang on. Just saw there were updates in the PriceManager since 1.7.1. Closing for now while I upgrade and test again.

peterjaap:
Confirmed fixed in at least 1.8.3;

![image](https://user-images.githubusercontent.com/431360/48565148-23812600-e8f8-11e8-9eb6-7492efe9918e.png)

PR #730, opened 2019-03-25
Added retrieval of store locale through config helper to fetch correc…

peterjaap:
…t locale instead of relying on LocaleResolver in non-frontend contexts

**Summary**

We noticed that when using the Reindex SKU functionality in the backend, the locale format for the price wasn't correct. I've opened a similar issue before (#615) which was fixed eventually but since the introduction of the Reindex SKU option, the incorrect locale currency formats were back.

I expected to see `€ 24,95` but actually got `€24.95` (so without a space and with a dot instead of  comma). This is the `en_US` locale while I expected the `nl_NL` locale (or, depending on storeview, the `de_DE` locale version `24,95 €`).

It turns out that calling `formatPrice()` in the Algolia extension delegates the responsibility of resolving the locale to Magento's `LocaleResolver`. Which in most cases is perfectly fine, except we're trying to get a locale from a product in the backend. So we had the backend in `en_US` locale while the product needed the `nl_NL` locale for the Dutch frontend to be synced to Algolia.

I've therefore refactored the `formatPrice()` method to fetch the locale from the database instead of relying on the `LocaleResolver`. This causes the correct format to show up.

**Result**
The correct price format :tada:


peterjaap:
Maybe also all `priceCurrency->format()` calls in the price manager should be wrapped through this?

JanPetr:
Thank you @peterjaap 💯 Definitely makes sense.

We removed the [dead code from PriceManager](https://github.com/algolia/algoliasearch-magento-2/pull/732), I resolved the conflicts and push a commit to fix coding styles. I hope you don't mind. When the Travis is done with CI, I'm happy to merge. Thank you again! 🙌

PR #734, opened 2019-03-26
Show indexer warning when type mismatch between Algolia Products and Price Indexer

peterjaap:
**Summary**
When the price indexer is set to Update on Schedule, and the Algolia Products indexer set to Update on Save, the prices synced to Algolia may be incorrect. This showed up when we raised the price for a product; the previous price was set as the 'special price' (final price) while the new price was set as the new original price, causing a discount to show up. This appears to be caused by the price index; when the collection with the product is loaded, the `final_price` field is populated from the indexer table. But since this indexer is set to Update on Schedule, the indexer table has the old price.

The solution is to also set the Catalog Price Indexer to Update on Save, causing the price indexer to run and put the correct price in the price indexer table. The Algolia indexer then fetches the correct price.

This PR adds a warning on the dashboard and the indexer listing page if the configuration isn't correct.

**Correct configurations:**
Algolia Products Indexer = Update on Save
Catalog Price Indexer = Update on Save

Algolia Products Indexer = Update on Schedule
Catalog Price Indexer = Update on Schedule

Algolia Products Indexer = Update on Schedule
Catalog Price Indexer = Update on Save

**Incorrect configuration:**
Algolia Products Indexer = Update on Save
Catalog Price Indexer = Update on Schedule

**Result**

image

image


peterjaap:
Sidenote; yes, it is better to use Update on Schedule. Unfortunately, we have some very impatient clients.

JanPetr:
Great catch! Thanks you for the contribution @peterjaap 💯 
I [commit a small CS changes](https://github.com/algolia/algoliasearch-magento-2/pull/734/commits/ce936992d06c1f217f04c4490b7854fa6595c3c2), I hope you don't mind.

I'll be happy to merge, when Travis is finished :)

peterjaap:
:+1:

PR #744, opened 2019-04-01
Fix suggestions functionality when "All Departments" string is translated

JanPetr:


JanPetr:
Good point, @bsuravech. We can take it into account while refactoring the front-end :)

Issue #748, opened 2019-04-09
Getting an 503 while Reindex SKU is running


thanura-bi:
Hi,
When I run Algolia search re-index SKUs, it's throwing a 503 error on below call,
.../algolia_algoliasearch/reindex/save
I can see the actual error in Magento reports,

```
`{"0":"Warning: Illegal offset type in vendor\magento\framework\DataObject.php on line 76","1":
"#0 vendor\magento\framework\DataObject.php(76): Magento\\Framework\\App\\ErrorHandler->handler(2, 'Illegal offset ...', '\app\...', 76, Array)\n
 #1 [internal function]: Magento\\Framework\\DataObject->setData(Object(Magento\\Framework\\Phrase),Object(Magento\\Framework\\Message\\Collection))\n
 #2 vendor\magento\framework\Session\SessionManager.php(157): call_user_func_array(Array, Array)\n
 #3 generated\code\Magento\Framework\Message\Session\Proxy.php(103): Magento\\Framework\\Session\\SessionManager->__call('setData', Array)\n
 #4 vendor\magento\framework\Message\Manager.php(121): Magento\\Framework\\Message\\Session\\Proxy->__call('setData', Array)\n
 #5 vendor\magento\framework\Message\Manager.php(143): Magento\\Framework\\Message\\Manager->getMessages(false, Object(Magento\\Framework\\Phrase))\n
 #6 vendor\magento\framework\Message\Manager.php(315): Magento\\Framework\\Message\\Manager->addMessage(Object(Magento\\Framework\\Message\\Error), Object(Magento\\Framework\\Phrase))\n
 #7 vendor\algolia\algoliasearch-magento-2\Controller\Adminhtml\Reindex\Save.php(169): Magento\\Framework\\Message\\Manager->addErrorMessage(Object(Algolia\\AlgoliaSearch\\Exception\\ProductDisabledException), Object(Magento\\Framework\\Phrase))\n
 #8 vendor\algolia\algoliasearch-magento-2\Controller\Adminhtml\Reindex\Save.php(98): Algolia\\AlgoliaSearch\\Controller\\Adminhtml\\Reindex\\Save->checkAndReindex(Object(Magento\\Catalog\\Model\\Product\\Interceptor), Array)\n
 #9 vendor\magento\framework\App\Action\Action.php(107): Algolia\\AlgoliaSearch\\Controller\\Adminhtml\\Reindex\\Save->execute()\n
 ...........
```

The issue is at,
[...]

PR #749, opened 2019-04-09
remove exception object from parameter for addErrorMessage method

bsuravech:
**Summary**
Update addErrorMessage() in \Algolia\AlgoliaSearch\Controller\Adminhtml\Reindex\Save to remove exception object in parameters. 

Related Issues: #748

**Result**
Updated all addErrorMessage calls in Reindex SKU save controller.


JanPetr:
Hello @bsuravech, is this the case for all Magento versions?

bsuravech:
@JanPetr I just checked 2.1, 2.2, and 2.3. This is applicable for all three general versions. I made the error by just changing the method from addExceptionMessage which required the exception object. I didn't check to remove it.

PR #750, opened 2019-04-09
Bugfix searchable attribute configuration not saving correct scope 

bsuravech:
**Summary**
Helpscout Issue (132804) reported searchable attributes are not respecting scope level and only returning default or current saved. 

**Result**
Added storeID for searchableAttribute so that it pulls correctly the scope.


bsuravech:
I've updated getUnretrieveableAttributes() as well! Good catch!

PR #755, opened 2019-04-17
Fixed active categories for stores with a different root category ID

peterjaap:
**Summary**
We noticed having categories in our category indexes that belonged to a different storeview.

The problem is that the extension expects that all categories are available in the default scope (0).

This isn't always the case since we can set a website to a separate category root with completely different category structure.

The current situation causes ALL categories to show up in ALL stores, a lot of them without a rewrite since the rewrite doesn't exist since the category isn't associated to that store.

This method walks through all active store groups and fetches the category root ID. It then walks through the activeCategories array looking for categories that are assigned to that root ID and then changes the prefix (which is almost certainly `0-`) to the correct prefix, causing these categories only to show up in that store.

**Result**
Correct categories only show up in the correct indexes.

The solution I provided here probably isn't the nicest since it fixes a wrong output instead of generating a correct output in the first place. However, this was the easiest and quickest for me to fix the issue short-term so I could create a module or a patch. I'll leave the refactoring to you guys.

### DIFF
--- a/Helper/Entity/CategoryHelper.php
+++ b/Helper/Entity/CategoryHelper.php
@@ -471,6 +471,7 @@ class CategoryHelper
                 ->from([$connection->getTableName('store'), 'store'], ['store_id', 'store_group.root_category_id'])
                 ->joinInner([$connection->getTableName('store_group'), 'store_group'], 'store.group_id = store_group.group_id')
                 ->where('store.store_id != ?', 0)
+                ->where('store_group.root_category_id != ?', 2)
         );
 
         $fixedCategories = [];
--- a/Helper/Entity/CategoryHelper.php
+++ b/Helper/Entity/CategoryHelper.php
@@ -443,7 +443,54 @@ class CategoryHelper
             $this->activeCategories = $connection->fetchAssoc($select);
         }
 
+        $this->activeCategories = $this->fixStorePrefixForActiveCategories($this->activeCategories);
+
         return $this->activeCategories;
+    }
+
+    /**
+     * The problem is that the extension expects that all categories are available in the default scope (0).
+     *
+     * This isn't always the case since we can set a website to a separate category root with completely different
+     * category structure.
+     *
+     * The current situation causes ALL categories to show up in ALL stores, a lot of them without a rewrite
+     * since the rewrite doesn't exist since the category isn't associated to that store.
+     *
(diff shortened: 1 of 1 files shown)


peterjaap:
I just realized my fix (well, workaround) only works for my scenario where there are multiple stores using the original root category and one other store using a second root category. When you would have additional stores using the second root category, it won't index categories for that store.

So @JanPetr  / @damcou this is the use case you should test against to cover all bases;

+----+------------------+---------------------+------------+------------+------------+
| Website ID | Default Group Id | Name                | Code       | Sort Order | Root Category Id |
+----+------------------+---------------------+------------+------------+------------+
| 0  | 0                | Admin               | admin      | 0          |           |
| 1  | 1                | Chairs English   | base       | 0          | 2          |
| 2  | 2                | Chairs Dutch | chairs_nl | 0          | 2          |
| 3  | 3                | Tables English     | tables_en | 0          | 40          |
| 4  | 4                | Tables Dutch     | tables_nl | 0          | 40          |
+----+------------------+---------------------+------------+------------+------------+

JanPetr:
Hello @peterjaap,

thank you for the PR 💯 
Can you reproduce the solved issue even on `develop` branch? Categories went through big refactoring since the last release and we tried to target the issue as well. Could you give it a try and see you can still reproduce?

Thank you! ❤️

peterjaap:
@JanPetr unfortunately the problem still occurs when using the develop branch. I cleared all my categories indices and ran `php bin/magento ind:reindex algolia_categories && PROCESS_FULL_QUEUE=1 php bin/magento ind:reindex algolia_queue_runner`, and all categories are still showing up in all categories.

JanPetr:
Hello @peterjaap,

sorry for the late answer. I just tried that and I'm not able to reproduce on the latest `develop` branch.

I created another website and another root category. Created new category under the new root category, assigned a product there and reindexed.

The category is indexed correctly in the separate index and not in the default store one. As well, the product is correctly indexed in a separate index only with the category from its store. 

I tested it on M 2.2.6 Enterprise.

Do I miss something?

damcou:
Hello there, as this PR is now very old and the extension has evolved a lot, I'm not sure it's still relevant.
If you want to submit it again, feel free to reopen/recreate it with on an up-to-date branch.

PR #759, opened 2019-04-18
Reindex affected products of saved category

### LINKED ISSUE #758: Category name change and data inconsistencies
Bug

1. Have categories, products and data synced to Algolia. Use Algolia to render Magento category pages
2. Go to Magento admin category and select a category with a number of products
3. Change the category name and save
4. Open category page on front-end of site
5. Page loads with no products or missing for a period of time (until affected products are reindexed)

On step 4 users should be able to load the category and see all products

This is caused due to the way Algolia fetches the products/ data for the category.
The POST that fetches the category data uses the category names in the Form Data request
`facetFilters: [["categories.level1:Kids /// PJ Sets test"]]` and because the name has changed in Magento and thus the POST value for facetFilters changes at the same time, whilst the category names need to be re-index to Algolia, we see inconsistencies.

A potential solution for this would be to filter by Magento category ID, as this is constant and will remove the inconsistency of missing products.
 
Screen Shot 2019-04-18 at 2 10 39 pm

**Feature: What is your use case for such a feature?**

1. Anyone that uses Magento + Algolia and changes the name of any category.
2. Magento Admin needs to change the category name for a new sale
 

**Feature: What is your proposed configuration entry? The new option to add? What is the behavior?**
NA

Magneto Cloud 2.2.6
Algolia latest

JanPetr:
Fixes #758


dzungnt:
I tried to apply that fix to our site and make some testing, however I got the issue that all of products are missing on category frontend after the queue processed. And the products showing back after about 1 hour. I checked the product data and it seem the category_ids field is empty for a while after the queue processed. So I think that could be the issue here.
It takes a long time for the products to be updated with new category info with the patch in place - up to 1 hour for the update during which time all of the categories associated to the product are removed from the product data in Algolia. In other words they are removed from all categories onsite which is not acceptable.

PR #775, opened 2019-05-20
Update Category Plugin 

bsuravech:
Improving PR #759

Following best practice in Magento to avoid using around for plugin by changing to the appropriate before/after method. 

Previous aroundDelete getProductCollection() was not fetching the correct store view product collection for category and in some cases not returning any products. Since we only require product_ids, I used the getProductsPosition() to fetch the product IDs to be passed into the $affectedProductIds variable for indexing.

### DIFF
--- a/Model/Indexer/CategoryObserver.php
+++ b/Model/Indexer/CategoryObserver.php
@@ -11,9 +11,8 @@ class CategoryObserver
 {
     private $indexer;
 
-    public function __construct(
-        IndexerRegistry $indexerRegistry
-    ) {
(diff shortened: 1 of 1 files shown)


dzungnt:
Hi @JanPetr,
I have applied your fix to our site, but when I changed the category name on admin, then all of the products on category frontend dropped after all the algolia queue processed. It seem your changes still not fix that issue. I'm using Algolia version 1.10.0 and Magento EE 2.2.6. Please help me to check!
Thanks.

PR #789, opened 2019-06-07
[add-storeid-to-facets] Add storeId to getFacets when you send change…

Eroslaev:
…s to algolia (for multistore)

Why this need?

When you have on magento multistore and need different design (on first site price slider, on second - don't need) all your changes on algolia.com set from storeId = null (first your store view) and all facets showed on all sites equal. That's not correct, I think


JanPetr:
Great catch, @Eroslaev! Thank you! 🚀

Issue #769, opened 2019-05-02
Html_entity_decode voor de request query


JeroenBoersma:
We received a security report for one of our customers regarding unescaped HTML-tags in the content which can be set via request parameters.

A request to: `/catalogsearch/result/index/?q=hello<script>alert(1)</script>` will add:
`{...snap...,"request":{"query":"hello<script>alert()<\/script>",...snap...}` to the config instead of the `&lt;` `&gt;` replacements one would expect or the url'encoded `%20` versions.

In our case they try to inject `<esi:...>` blocks, where they try to retrieve secure data. Of coarse they where filtered by our hosting, but it raises questions and that's why I reach out here.

Our research got track of a `html_entiy_decode` of the `q` parameter and that way added to the json configuration.

**Related files:**

https://github.com/algolia/algoliasearch-magento-2/blob/198eda5d1f13403b71a560a5b2dc4cbe4af8046e/Block/Configuration.php#L183

https://github.com/algolia/algoliasearch-magento-2/blob/198eda5d1f13403b71a560a5b2dc4cbe4af8046e/view/frontend/templates/internals/configuration.phtml#L20

Here I see the usage in `instantsearch.js`, which will be send to the minified algoliaBundle.js where I lost track of it. Didn't have time at hand to look into the original bundle or refactor the bundle.

https://github.com/algolia/algoliasearch-magento-2/blob/198eda5d1f13403b71a560a5b2dc4cbe4af8046e/view/frontend/web/instantsearch.js#L158

Because I don't know exactly what is going on in the widgets of Algolia I post the question here where the experts are.

**Questions**

- Is there any reason why the `decoding` is done?
- Is this still needed or can we life without it, which would be my recommendation.

damcou:
Hello there, I'm closing the issue as we didn't hear from you for a while.

Regarding the issue raised previously, the related PR is still existing and will be addressed as soon as possible.

In the meantime, you can keep track of it looking at our [M2 community board](https://github.com/algolia/algoliasearch-magento-2/projects/1).

If you are still experiencing further issues, please reach out to us at support@algolia.com with any error logs, screenshots of examples you can provide, as well as your APPID so that we can further assist you. Thanks!

PR #800, opened 2019-06-18
Escape "query" parameter in JSON configuration

JanPetr:
Related Issue(s): #769


view/frontend/web/internals/common.js
```diff
@@ -834,3 +834,20 @@ var AlgoliaBase64 = {
...
 	}
 };
+
+var htmlspecialchars_decode = function(string) {
+	var unescapedString = string,
+		specialchars = [
+			[ '"', '&quot;' ],
+			[ '>', '&gt;' ],
+			[ '<', '&lt;' ],
+			[ '&', '&amp;' ]
```
tkrugg: it is usually expected to include `'`. Look for example in this lodash implementation: https://github.com/lodash/lodash/blob/master/unescape.js#L3-L7

Block/Configuration.php
```diff
@@ -186,7 +186,7 @@ public function getConfiguration()
             'maxValuesPerFacet' => (int) $config->getMaxValuesPerFacet(),
             'autofocus' => true,
             'request' => [
-                'query' => html_entity_decode($query),
+                'query' => htmlspecialchars(html_entity_decode($query)),
```
tkrugg: shouldn't the function here be `htmlspecialchars_decode ` ?

bsuravech: `htmlspecialchars` is correct here as we should be encoding it at this stage and decoding in js.

bsuravech:
@tkrugg I don't know if we want to include lodash here just for this fix. I've made some adjustments so that the method is defined with in the algolia object. Can you review that when you have a moment? I've also added the apostrophe character "`'`"

Can you elaborate on your comment regarding "`htmlspecialchars` looks not defined." ? That is a php function that encodes the query in the configuration json output.

view/frontend/web/internals/common.js
```diff
@@ -45,6 +45,23 @@ var algolia = {
...
+	},
+	htmlspecialcharsDecode: function(string) {
+		var unescapedString = string,
+			specialchars = [
+				[ '"', '&quot;' ],
+				[ '>', '&gt;' ],
+				[ '<', '&lt;' ],
+				[ '&', '&amp;' ],
+				[ "'", '&#39;' ]
+			];
```
broujo: Hey team,

The OWASP recommends to escape forward slash too (https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html). Not sure how necessary it is to actively prevent XSS, but I think it's still worth including.

Block/Configuration.php
tkrugg: yes, it makes sense.

tkrugg:
@bsuravech sorry for the late reply.
> htmlspecialchars looks not defined.
I just meant I didn't find the definition of the function. Can you point it to me?

Regarding lodash, as far as I can tell, it is already part of the scope so you might as well just use it, and their function are always battle tested, but I don't have a strong opinion on this.

bsuravech:
Closing out as it is outdated. If issue returns, we'll address it in the latest versions.

PR #805, opened 2019-06-25
Events tracking

JanPetr:
Adds event tracking via Algolia's proxy.


Helper/ConfigHelper.php
```diff
@@ -458,13 +458,18 @@ public function isAutocompleteDebugEnabled($storeId = null)
         return $this->configInterface->isSetFlag(self::AUTOCOMPLETE_MENU_DEBUG, ScopeInterface::SCOPE_STORE, $storeId);
     }
 
-    public function getSorting($storeId = null)
+    public function getSortingDefaultValue($storeId = null)
```
damcou: Isn't it more a "raw" value than a "default" one ?

Model/Observer/SaveSettings.php
```diff
@@ -16,10 +17,17 @@ class SaveSettings implements ObserverInterface
     /** @var IndicesConfigurator */
     private $indicesConfigurator;
 
-    public function __construct(StoreManagerInterface $storeManager, IndicesConfigurator $indicesConfigurator)
-    {
+    /** @var ConfigurationTracker */
+    private $configurationTracker;
+
+    public function __construct(
```
0livier: Please add a docblock

Helper/ConfigHelper.php
bsuravech: Updated to:

getRawSortingValue
getRawProductCustomRanking

Model/ConfigurationTracker.php
```diff
@@ -0,0 +1,128 @@
...
+            'storeId' => $storeId
+        ]);
+    }
+
+    /**
+     * @param int $storeId
+     *
+     * @return bool
+     */
+    private function isSortingChanged($storeId)
```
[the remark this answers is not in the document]
bsuravech: The parent method of this function has the same parameter that is required. 

' public function trackConfiguration($storeId)`

I can add the option for default but it will always have a $storeId param in this case.

AlexEven:
Ping @dmartindelorme @mathieuolivier  - FYI!

dmartindelorme:
Hey team, @AlexEven pinged @mathieuolivier and me because we're in charge of making sure our standards of Segment implementations are respected. Since we don't have a doc yet on this (sorry for that 🙇) we're doing this project by project for now.

First things first, in production are you going to use this js source https://app.segment.com/algolia-app/sources/magento/overview or this python source https://app.segment.com/algolia-app/sources/python_magento_2_proxy/overview? I'm a bit confused about this.

FYI we usually recommend using only a front-end implementation as much as possible. Except of course if you need to track events or data that is only present in the backend.

dmartindelorme:
Second thing, the usual implementation is to have one source for dev/staging data and one source for prod data. Example, for the Algolia dashboard we have this 
image. Each source has at least one mandatory destination: Redshift dev database for the dev data source and the Redshift prod database for the prod data source.

Do you think you create those 2 sources and then we'll be able to test that dev tracking events are forwarded to the Segment dev source before rolling this out in production?

Again sorry for coming in late in the process 🙇.

0livier:
:rocket:

PR #807, opened 2019-06-26
Resolve issue from Magento 2.3.2 causing category pages to break

marcusirgens:
**Summary**
The parameter list for \Magento\Framework\Search\Adapter\Mysql\ResponseFactory changed in Magento 2.3.2, causing an exception to be thrown on the category pages. This patch should be backwards compatible, and resolve the issue.


marcusirgens:
Closing in favour of #807

PR #814, opened 2019-07-03
Removed objectManager from ProductHelper

peterjaap:
Also removed useless options array from ImageHelper

**Summary**

Using ObjectManager directly is bad mmmmkay

**Result**
No objectmanager used directly :tada:


Helper/Entity/ProductHelper.php
```diff
@@ -20,19 +21,22 @@
 use Magento\Catalog\Model\Product\Type\AbstractType;
 use Magento\Catalog\Model\Product\Visibility;
 use Magento\Catalog\Model\ResourceModel\Eav\Attribute as AttributeResource;
+use Magento\Catalog\Model\ResourceModel\Product\Collection;
```
fooman: I think moving this into the constructor could introduce a subtle change. Previously `objectManager->create()` would provide a new Collection object every time it's run. Moving it into the constructor would make it a shared object. I don't have any Algolia instance to test but from looking at the code using `CollectionFactory` and then `->create()` on it seems safer.

peterjaap: Thanks, you're right! I changed it to use the factory.

Helper/Entity/ProductHelper.php
```diff
@@ -201,10 +218,8 @@ public function getProductCollectionQuery(
         $onlyVisible = true,
         $includeNotVisibleIndividually = false
     ) {
-        /** @var \Magento\Catalog\Model\ResourceModel\Product\Collection $products */
-        $products = $this->objectManager->create('Magento\Catalog\Model\ResourceModel\Product\Collection');
-
-        $products = $products
+        $productCollection = $this->productCollectionFactory->create();
+        $products = $productCollection
```
damcou:
```suggestion
        $products = $productCollection
```

damcou: Hello @peterjaap !
Thanks for this PR and sorry that we didn't have time to handle it before.
Is it possible to submit this change so the codestyle check can pass ?
Thanks in advance ;)

Helper/Entity/ProductHelper.php
```diff
@@ -20,19 +21,22 @@
...
 use Magento\Framework\DataObject;
 use Magento\Framework\Event\ManagerInterface;
-use Magento\Framework\ObjectManagerInterface;
 use Magento\Store\Model\Store;
 use Magento\Store\Model\StoreManagerInterface;
 
 class ProductHelper
 {
+    public $productCollectionFactory;
+    public $groupCollection;
```
damcou: Could you please make these 2 properties private ?

Helper/Image.php
```diff
@@ -10,8 +10,11 @@
 
 class Image extends \Magento\Catalog\Helper\Image
 {
+    /**
+     * @var ConfigHelper
+     */
+    public $configHelper;
```
damcou: Same thing here :) 
Thanks in advance.

Helper/Entity/ProductHelper.php
peterjaap: I'd rather not. Shall we compromise and make them protected? Private leaves me with very few nice options to override these.

Helper/Image.php
peterjaap: I'd rather not. Shall we compromise and make them protected? Private leaves me with very few nice options to override these.

Helper/Entity/ProductHelper.php
peterjaap: (Right now we have a composer patch in place that makes all private properties and functions in your extension public so we can pluginize/override them - this isn't preferred)

Helper/Image.php
damcou: Let's go for protected :)

Helper/Entity/ProductHelper.php
```diff
@@ -86,6 +89,25 @@ class ProductHelper
         'color',
     ];
 
+    /**
+     * ProductHelper constructor.
```
damcou: Hi @peterjaap !
Looks that there's a separation line missing between the description of the constructor and the first param (it breaks the quality-tools check).
Could you please add it and then I think we're good to merge ;)

Helper/Entity/ProductHelper.php
```diff
@@ -86,6 +89,26 @@ class ProductHelper
...
+     *
```
damcou:
```suggestion
     *
```
@peterjaap Super sorry to ask you to change this again, but the tool is pretty strict :(

PR #817, opened 2019-07-08
Added All Website domains for advanced restriction

vmalyk:
**Summary**

Added to Client add websites domains for Algolia Client init
 
**Result**

Added possible to manage restriction by domain (*HTTP Referers* for requests)  in Algolia App - API keys tab and prevent reindex data.


bsuravech:
Hi @vmalyk! Thanks for your PR. Unfortunately, the hostsArray parameter for the API Client is used to add custom hosts/endpoint used by the API. We currently do not have indexing operations restricted by website/domain in the extension but this could be something to take into consideration future feedback. Thanks!

PR #819, opened 2019-07-10
Category Product Updates for Update on Schedule

bsuravech:
**Summary**
Products within a category needs to be updated when a category name, include_in_menu, and/or is_active attributes are updated. When the algolia indexes are set to Update on Schedule, the mview subscription for categories does not push products into the algolia_products_cl table if a products in the category have not been changed. We need to compensate for this missing condition. 

Related Issues: #758

**Result**
Updated the CategoryObserver beforeSave() method to pass in also deleted or added products as well as condition the product collection to only when there are specific modifications. This is to reduce extra operations on indexing. We also added a new method to update the product depending on the mode of the indexer.

### DIFF
--- a/Model/Indexer/CategoryObserver.php
+++ b/Model/Indexer/CategoryObserver.php
@@ -58,7 +58,8 @@ class CategoryObserver
             // To reduce the indexing operation for products, only update if these values have changed
             if ($category->getOrigData('name') !== $category->getData('name')
                 || $category->getOrigData('include_in_menu') !== $category->getData('include_in_menu')
-                || $category->getOrigData('is_active') !== $category->getData('is_active')) {
+                || $category->getOrigData('is_active') !== $category->getData('is_active')
+                || $category->getOrigData('path') !== $category->getData('path')) {
                 /** @var ProductCollection $productCollection */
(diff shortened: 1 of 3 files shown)


Model/Indexer/CategoryObserver.php
```diff
@@ -40,12 +54,25 @@ public function __construct(IndexerRegistry $indexerRegistry, ConfigHelper $conf
     public function beforeSave(CategoryResourceModel $categoryResource, CategoryModel $category)
     {
         $categoryResource->addCommitCallback(function() use ($category) {
-            if (!$this->indexer->isScheduled() || $this->configHelper->isQueueActive()) {
+            $collectionIds = [];
+            // To reduce the indexing operation for products, only update if these values have changed
+            if ($category->getOrigData('name') !== $category->getData('name')
```
maximbaibakov: Hey @bsuravech 

I think, it would make sense to add an extra condition to take into account category path to support moving properly moving category to another category.

Regards,
Maxim

bsuravech: Hey @maximbaibakov! Good catch! Updating now!

Issue #793, opened 2019-06-12
Magento 2 B2B Index Error


glo-arandazzo:
**Bug: What is the current behavior?**
On a Magento 2.2.8 Enterprise B2B site using shared catalogs and permissions, the following error occurs:

```#bin/magento index:reindex algolia_products
Algolia Search Products indexer process unknown error:
Notice: Undefined offset: 1 in /site/vendor/algolia/algoliasearch-magento-2/Model/Observer/CatalogPermissions/ProductCollectionAddPermissions.php on line 66```

**Bug: What is the expected behavior?**
The index should complete without issue. 

Any help would be greatly appreciated.

bsuravech:
Hi @glo-arandazzo,

The PR to correct this issue has been merged into the latest version. Please feel free to upgrade and let us know if that works out for you. Thanks!

PR #820, opened 2019-07-12
Update B2B permissions condition

bsuravech:
**Summary**
Update B2B permissions to include the allow browsing catalog scope. Add count checks before list to prevent notice errors.

Related Issues: #793

### DIFF
--- a/Factory/CatalogPermissionsFactory.php
+++ b/Factory/CatalogPermissionsFactory.php
@@ -9,6 +9,7 @@ use Magento\Framework\ObjectManagerInterface;
 class CatalogPermissionsFactory
 {
     const CATALOG_PERMISSIONS_ENABLED_CONFIG_PATH = 'catalog/magento_catalogpermissions/enabled';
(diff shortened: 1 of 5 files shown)


Factory/CatalogPermissionsFactory.php
```diff
@@ -53,6 +46,11 @@ public function getCatalogPermissionsHelper()
         return $this->objectManager->create('\Magento\CatalogPermissions\Helper\Data');
     }
 
+    public function getCatalogPermissionsConfig()
+    {
+        return $this->objectManager->create('\Magento\CatalogPermissions\App\Config');
```
damcou: Is there any particular reason why the objectManager is used in these 2 factories rather than construct injection ?

bsuravech: Perhaps I'm not using constructors enough to know but I used the objectManager to generate the classes because catalogpermissions is a Commerce feature and isn't available on open source. I don't want to generate errors if the module doesn't exists or isn't enabled. I'm free to discuss if you have an improved idea for this case.

Issue #818, opened 2019-07-08
Error fetching images in 1.11 releases


batmike:
I've tested the 1.11 release series with our Commerce install version 2.3.1 and 2.3.2. Each time, the resulting index in Algolia contains only the placeholder images.

All of our configurable products do not have any images assigned so everything is pulled from the child simple products. However, something is not working right in the 1.11 series as it just gets the placeholders and that's all that ends up showing on the InstantSearch pages and the auto-complete.

Let me know if more information would help in troubleshooting this. Thanks!

bsuravech:
Hi @batmike! Can you confirm that this was working in the 1.10 version? Was there any customizations made on your end for the previous version to pull child simple product images to its parent? Perhaps a backend observer on product indexing? 

The extension does have a feature where you can switch the color variation image from it's simple product. This is a configuration that can be found in Configuration > Algolia Search > Products. Would this help your issue?

batmike:
@bsuravech - It is working as expected in the 1.10 release. I’ve also checked and the setting for replacing the image with the variation is set to Yes and has been in both 1.10 and in our testing of 1.11.

Other than that, we do not have any custom observers or anything else that would be interfering with the processing of the images as the indexes are built.

bsuravech:
Hi @batmike,

I see it now. We have PR#727 had updated the product image helper to pull the correct product image type, however, it seems that $this->getProduct()->getImage() that was removed by this was actually setting the child image as the parent's image. 

https://github.com/algolia/algoliasearch-magento-2/pull/727

We have created a ticket internally to re-add the ability to get the correct image and type from the child product if the parent does not have an image assigned. 

We will update once this issue once it has been addressed. Thank you for reporting!

batmike:
@bsuravech,

Thanks for the information and I'm glad it's was fairly straightforward to track down. I'll watch for a new release and test out the upgrade after that.

bsuravech:
Hi @batmike, 

PR#823 has been included in our latest release v1.12.0. Please feel free to patch or upgrade to the latest version to incorporate these changes. Thanks for reporting!

PR #823, opened 2019-07-17
Update get product images

bsuravech:
**Summary**
Configurable products without an image are not pulling from their child product image as it did before.

Related Issues: #818

**Result**
Add back the initBaseFile() method, however, fetching the appropriate image type from child products if the configurable parent does not have an image set.

### DIFF
--- a/Helper/Image.php
+++ b/Helper/Image.php
@@ -3,6 +3,7 @@
 namespace Algolia\AlgoliaSearch\Helper;
 
 use Magento\Catalog\Model\Product\ImageFactory;
+use Magento\ConfigurableProduct\Model\Product\Type\Configurable as ProductTypeConfigurable;
 use Magento\Framework\App\Helper\Context;
 use Magento\Framework\View\Asset\Repository;
 use Magento\Framework\View\ConfigInterface;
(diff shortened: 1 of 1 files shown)


Helper/Image.php
```diff
@@ -61,6 +62,44 @@ public function getUrl()
...
+
+    /**
+     * Configurable::setImageFromChildProduct() only pulls 'image' type
+     * and not the type set by the imageHelper
+     *
+     * @return string
+     */
+    private function getProductImageType()
+    {
+        if (!$this->getImageFile() && $this->getType() !== 'image') {
```
0livier: To be 100% sure, is it possible to call `$this->getProduct()->getImage();` even if  `!$this->getImageFile() && $this->getType() !== 'image'` ? I've reformulated the following code to highlight what I mean (hopefully).

```suggestion:-0+13
if ($this->getImageFile() || $this->getType() === 'image'
    || $this->getProduct()->getTypeId() !== ProductTypeConfigurable::TYPE_CODE) {
    return $this->getProduct()->getImage();
}

$childProducts = $this->getProduct()->getTypeInstance()->getUsedProducts($this->getProduct());
foreach ($childProducts as $childProduct) {
    if ($childProduct->getData($this->getType())
        && $childProduct->getData($this->getType()) !== 'no_selection') {
        return $childProduct->getData($this->getType());
    }
}

// should it be possible still end up here?
return $this->getProduct()->getImage();
```

0livier: I do believe that exiting/returning as soon as possible is a good practice, even if it means having several returns. Yet, don't take my word for it, ask around for other opinions.

My very first goal with the comment was to highlight that you could end up with calling $this->getProduct()->getImage(); if there's no image and only "no_selection" type in child products. I wanted to ensure it was a possible / safe possibility.

bsuravech: the method `$this->getProduct()->getImage()` will always return a value so it's safe to return as the default. I've updated the code to break it apart. Hopefully it's more readable.

0livier: It is. LGTM :+1:

julianvdrielen:
@0livier Any information on when this will be merged & released?

PR #835, opened 2019-08-15
AttributesToHighlight fix on frontend

vmalyk:
**Summary **
Pull-request contains 2 fixes:
1. If I removed all AttributesToHighlight from index in Algolia Application hit._highlightResult would be undefined and I get Javascript error.
2. After re-save index from Magento - all AttributesToHighlight for index are removed.


unicoder88:
Upvote! :)

PR #838, opened 2019-08-19
chore(InstantSearch V3): migrate javascript

tkrugg:
**Summary**

In a nutshell:

- migrated every widget to its IS v3 equivalent. Some of
notable changes are:
  * instantsearch: options have changed:
    - IS v2 accepts `appId` and `apiKey`, IS v3 accepts `searchClient`
  * searchBox template changed:
    - IS v2 accepts an input field, IS v3 does not: cf: https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/search-box/search-box.js#L124
  * stats: template options have changed:
    - IS v2 has the `templates.body`, IS v3 has `templates.text`
    - IS v2 has the `transformData`, IS v3 does not have it but found a workaround with function syntax
  * sortBy: options have changed:
    - IS v2 widget was called sortBySelector, IS v3 component is sortBy
    - in IS v2 accepts {name: string, label: string}, in IS v3 accepts {value: string, label: string}
  * currentRefinements: various things changed
    - in IS v2 it could "clear all", in IS v3 that's done by clearRefinements
    - `attributes` is now `includedAttributes`
    - header template has been removed, added a panel
  * clearRefinements: introduced
  * queryRuleCustomData: introduced for the banner
    - previously using `hits` widget
  * infiniteHits & hits: options have changed
   - transformData is now transformHits
   - __position computation was removed, as no longer needed since insights integration
  * pagination: options have changed
   - added options `showFirst showLast showNext showPrevious` to preserve current behaviour
   - `label.(previous|next)` is now `templates.(previous|next)`
   - removed `scrollTo: 'body'` as it's not the default behavour
  * hierarchicalMenu: options have changed
    - simplified item template
    - removed cssClasses as they are no longer necessary
    - header template has been removed, added a panel
  * refinementList(s) & RangeSlider: options have changed
    - header template has been removed, added a panel
    - update searchable* related options to preserve the current behaviour
    - removed unsused css classes

Other changes to note:
- reuse the hogan or html templates already being used.
- updates docs links in comments, to point to instantsearch v3 docs

For now, we're leaving the global stuff where they are.

**Result**

N/A


view/frontend/templates/instant/wrapper.phtml
```diff
@@ -23,44 +24,26 @@
...
                             {{/second_bar}}
                         </div>
                     </div>
                 </div>
                 <div class="row">
-                    <div>
+                    <div class="col-md-12">
                         <div itemscope itemtype="http://schema.org/ItemList" class="hits">
-                            <div class="infos">
+                            <div class="algolia-infos">
```
tkrugg: renamed this to stay consistent with algolia-banner.
There is probably more renaming we can do to make it clear which ids and classes we own.
Open to suggestions

view/frontend/templates/instant/wrapper.phtml
```diff
@@ -23,44 +24,26 @@
...
                                     <div class="sort-by-label pull-left">
-                                        <?php echo $block->escapeHtml(__('SORT BY')); ?>
+                                        <?php echo $block->escapeHtml(__('SORT BY')); ?> &nbsp;
                                     </div>
                                     <div class="pull-left" id="algolia-sorts"></div>
                                 </div>
                                 <div class="algolia-clearfix"></div>
                             </div>
                             <div id="algolia-banner"></div>
+                            <div id="instant-empty-results-container"></div>
```
tkrugg: this is the mounting point for the custom Suggestions widget 
Previously it was simply using the mounting point of the hits widget, which I found to be unnecessarily confusing, so I split them.

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
-						root: 'facet hierarchical'
-					}
+					sortBy: ['name:asc']
 				};
-				
+
 				hierarchicalMenuParams.templates.item = '' +
-					'<div class="ais-hierearchical-link-wrapper">' +
-					'<a class="{{cssClasses.link}}" href="{{url}}">{{label}}' +
+					'<a class="{{cssClasses.link}} {{#isRefined}}{{cssClasses.link}}--selected{{/isRefined}}" href="{{url}}">{{label}}' + ' ' +
```
tkrugg: I simply simplified this, I think adding the div was initially for styling, but we're going with a more simple/vanilla style.
In fact I would be even in favour of removing the template altogether. The default one is good enough imho
see https://instantsearchjs.netlify.com/stories/?path=/story/hierarchicalmenu--default

view/frontend/templates/instant/wrapper.phtml
Haroenv: why instant here & not Algolia as prefix?

view/frontend/templates/instant/wrapper.phtml
Haroenv: infos -> info (always singular in English)

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
-					root: 'facet'
-				},
 				templates: {
-					header: '<div class="name">' + algoliaConfig.translations.selectedFilters + '</div>',
-					clearAll: algoliaConfig.translations.clearAll,
 					item: $('#current-refinements-template').html()
 				},
-				attributes: attributes,
-				onlyListedAttributes: true
+				panelOptions: {
```
Haroenv: is this because you wrap everything in a panel now? I'm missing some context here probably :)

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
+					return items.map(function (item) {
+						var attribute = attributes.filter(function (_attribute) {
+							return item.attribute === _attribute.name
+						})[0];
+						if (!attribute) return item;
+						return {...item, label: attribute.label}
+					})
+				}
+			},
+			queryRuleCustomData: {
```
Haroenv: needs a comment

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
+				container: '#algolia-banner',
+				templates: {
+					default: `
```
Haroenv: can you use template strings here? I haven't checked other places, but IIRC there's no babel on this project

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
 				templates: {
 					item: $('#instant-hit-template').html()
 				},
-				transformData: {
-					item: function (hit) {
+				transformItems: function (items) {
+					return items.map(function (item) {
+						var hit = Object.assign({}, item);
 						hit = transformHit(hit, algoliaConfig.priceKey, search.helper);
+						// FIXME: transformHit is a global
```
Haroenv: doesn't seem relevant?

Haroenv: oh yes it is, but should be swapped with line above

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
-					data.first = data.page * data.hitsPerPage + 1;
-					data.last = Math.min(data.page * data.hitsPerPage + data.hitsPerPage, data.nbHits);
-					data.seconds = data.processingTimeMS / 1000;
-					
-					data.translations = window.algoliaConfig.translations;
-					
-					return data;
+					text: function (data) {
+						var hoganTemplate = algoliaBundle.Hogan.compile($('#instant-stats-template').html());
+						var _data = Object.assign({}, data)
```
francoischalifour: `Object.assign` is not supported on IE. Is it transpiled in the environment?

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
-					cssClasses: {
-						list: 'hierarchical',
-						root: 'facet hierarchical'
-					}
+					sortBy: ['name:asc']
 				};
-				
+
 				hierarchicalMenuParams.templates.item = '' +
-					'<div class="ais-hierearchical-link-wrapper">' +
```
Haroenv: note that this was misspelled too

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
+				includedAttributes: attributes.map(function (attribute) {
+					return attribute.name
+				}),
+				transformItems: function (items) {
+					return items.map(function (item) {
+						var attribute = attributes.filter(function (_attribute) {
+							return item.attribute === _attribute.name
+						})[0];
+						if (!attribute) return item;
+						return {...item, label: attribute.label}
```
francoischalifour: Spread operator won't work everywhere if there's no JavaScript tooling.

view/frontend/web/instantsearch.js
```diff
@@ -207,408 +206,454 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
...
-								priceUtils.formatPrice(formattedValue, algoliaConfig.priceFormat);
-						}
-					}
-				}];
+			return ['hierarchicalMenu', hierarchicalMenuParams];
+		}
+	};
+
+	/** Add all facet widgets to instantsearch object **/
+	window.getFacetWidget = function (facet, templates) {
```
Haroenv: name is fairly awkward, and this is global. Where is it used?

view/frontend/web/instantsearch.js
francoischalifour: Yes, if `panelOptions` is present:

https://github.com/algolia/algoliasearch-magento-2/blob/98d9457f8d46f33819d562a0ae4fbf4ba3069771/view/frontend/web/instantsearch.js#L640-L642

view/frontend/templates/instant/wrapper.phtml
tkrugg: I wanted to keep the same naming as the previous ` <div id="instant-search-results-container"></div>`
So I could have but then some will ask the other question: why `algolia` an not `instant` prefix 😄  

Like you, I definitely think naming here deserves a big make over. I left that in the refactoring step which will come later.

view/frontend/web/instantsearch.js
tkrugg: No it's not!, thanks for pointing in out!

view/frontend/web/instantsearch.js
tkrugg: Yes, everything facet widget is wrapped in a panel.
To reuse the current abstraction I has to "bundle" the panel configs along with the widget config.
This are some of the main things I thinking about removing with the refactoring, but for now this migrate the template.header

view/frontend/web/instantsearch.js
tkrugg: thanks!

view/frontend/web/instantsearch.js
tkrugg: haha yes :)

view/frontend/web/instantsearch.js
tkrugg: I didn't come up with the naming, I just moved the function
https://github.com/algolia/algoliasearch-magento-2/pull/838/files#diff-f45183f22871cde50da868482a4768dbL393
I'll remove it as global soon, I just need to make sure I didn't break anything :)

view/frontend/templates/instant/hit.phtml
```diff
@@ -12,7 +12,7 @@ $tierFormatedVar = 'price' . $priceKey . '_tier_formated'
 <script type="text/template" id="instant-hit-template">
     <div class="col-md-4 col-sm-6">
         <div class="result-wrapper" itemprop="itemListElement" itemscope itemtype="http://schema.org/Product">
-            <a itemprop="url" href="{{url}}" class="result" data-objectid="{{objectID}}" data-position="{{__position}}">
+            <a itemprop="url" href="{{url}}" class="result" data-objectid="{{objectID}}" data-position="{{__position}}" data-queryid={{__queryid}}>
```
francoischalifour: Nit: respecting the original letter case would lead to more predictable data attributes:

- Object ID → `data-object-id`
- Query ID → `data-query-id`

PR #841, opened 2019-08-22
Prepare MSI's handling

damcou:
Add some changes to the extension to be more easily overriden by the Algoia MSI external module.


Helper/Entity/ProductHelper.php
```diff
@@ -217,18 +217,11 @@ public function getProductCollectionQuery(
...
         }
 
         /* @noinspection PhpUnnecessaryFullyQualifiedNameInspection */
-        $products = $products->addFinalPrice()
-            ->addAttributeToSelect('special_price')
-            ->addAttributeToSelect('special_from_date')
-            ->addAttributeToSelect('special_to_date')
-            ->addAttributeToSelect('visibility')
-            ->addAttributeToSelect('status');
+        $products = $this->addMandatoryAttributes($products);
```
nunomaduro:
```suggestion
        $this->addMandatoryAttributes($products);
```

Helper/Entity/ProductHelper.php
```diff
@@ -263,6 +256,28 @@ public function getProductCollectionQuery(
...
+    {
+        /** @var \Magento\Catalog\Model\ResourceModel\Product\Collection $products */
+        $products->addFinalPrice()
+            ->addAttributeToSelect('special_price')
+            ->addAttributeToSelect('special_from_date')
+            ->addAttributeToSelect('special_to_date')
+            ->addAttributeToSelect('visibility')
+            ->addAttributeToSelect('status');
+
+        return $products;
```
nunomaduro: I think you need to return nothing here. Seems that we are mutating `$products`.

Helper/Entity/ProductHelper.php
```diff
@@ -728,7 +743,7 @@ private function addImageData(array $customData, Product $product, $additionalAt
         return $customData;
     }
 
-    private function addInStock($defaultData, $customData, Product $product)
+    protected function addInStock($defaultData, $customData, Product $product)
```
nunomaduro: There is a reason why we are altering the visibility of this method?

Helper/Entity/ProductHelper.php
```diff
@@ -41,7 +41,7 @@ class ProductHelper
     private $eventManager;
     private $visibility;
     private $stockHelper;
-    private $stockRegistry;
+    protected $stockRegistry;
```
nunomaduro: There is a reason why we need to alter the visibility here?

damcou: To be overriden by the inventory module, as discussed ;)

Helper/Entity/ProductHelper.php
damcou: To be overriden by the inventory module, as discussed ;)

Helper/Entity/ProductHelper.php
damcou: true ! :)

Helper/Entity/ProductHelper.php
damcou: done

PR #843, opened 2019-09-03
Fixed the dependency list

DavidLambauer:
This module depends on a lot more modules than it is declared in the module.xml. This can cause issues during a fresh installation for sure. I've parsed the code and identified the given list of dependencies.

**Summary**

**Result**


damcou:
Thanks a lot for this PR @DavidLambauer , this will be added in our next release!

PR #848, opened 2019-09-09
Update API client to v2

DevinAlgolia:
**Summary**
This PR updates the Algolia API Client from version 1 to version 2. This allows us to make use of any features that will be added through the client.

**Result**
The extension keeps working as expected.


nunomaduro:
My thoughts about this: in theory, users shouldn't not rely on dependencies of their dependencies. But in practice, I am afraid that magento users are using directly `PHP API Client v1` without specifying it in their `composer.json`.

This means, that if users update their magento dependency and they get `PHP API Client v2` we may break their implementation if they are using the api client directly.

Question for @bsuravech, @0livier and @damcou: Is it possible that magento developers are using the api client directly?

bsuravech:
@nunomaduro It could be possible that this is the case for our more customised clients but not a majority of them.

If the case that they have extended or utilised the php client directly, we should advise in the release notes that this version update will break compatibility. We will need to provide a link to the docs if there are any public/protected methods that have been modified between v1 and v2.

nunomaduro:
@damcou @bsuravech Alright! As discussed with @DevinAlgolia, we just need to make sure that we have a plan to test the modifications performed in this pull request. I have some ideas, we have a meeting `hursday, September 19⋅12:00 – 12:30pm` to discuss this.

DevinAlgolia:
Feedback is addressed, we can soon start the testing process!

DevinAlgolia:
Hi team,

just wanted to let you know that I ran manual tests to see if the extension still works as expected, and everything seems to be good! I did not get any errors, products get (re)-indexed properly 😄

PR #853, opened 2019-09-14
Pass set product website_id to load catalogrule prices with enabled customer groups

unicoder88:
**Summary**

We have a catalog rule that gives customer group "Blogger" discount to virtual products. We tried activating customer groups in Algolia module settings and reindexing, but all prices were exported as regular final price, blogger discounts not visible.

As it turns out, admin even observer applies catalogrule prices given assigned customer_group_id and website_id. However, website_id was not passed and so observer does nothing.

See [ProcessAdminFinalPriceObserver](https://github.com/magento/magento2/blob/2.3.1/app/code/Magento/CatalogRule/Observer/ProcessAdminFinalPriceObserver.php#L82).

Thank you :)

**Result**


damcou:
Good catch @unicoder88 and thanks a lot for your contribution. 
Your fix will be available in our next release ! ;)

PR #860, opened 2019-09-25
Magento Cloud Development setup teardown

bsuravech:
**INTERNAL DEV ONLY** 

Scripts to run for setting up and tearing down algoliasearch in the development Magento Commerce Cloud environment. This sets the environment prefix to the environment variable `MAGENTO_CLOUD_ENVIRONMENT` that is provided by the cloud. The teardown script will remove indexes and its replicas in a double iteration.

### DIFF
--- a/dev/cloud/teardown.php
+++ b/dev/cloud/teardown.php
@@ -1,4 +1,3 @@
-#!/usr/bin/env php
 <?php
 
 use Magento\Framework\App\Bootstrap;
@@ -10,10 +9,12 @@ $obj = $bootstrap->getObjectManager();
 
 $algoliaHelper = $obj->get('\Algolia\AlgoliaSearch\Helper\AlgoliaHelper');
 
-if ($algoliaHelper) {
-
-    $indices = $algoliaHelper->listIndexes();
-
+/**
+ * @param $algoliaHelper Algolia\AlgoliaSearch\Helper\AlgoliaHelper
+ * @param array $indices
+ */
+function deleteIndexes($algoliaHelper, array $indices)
+{
     foreach ($indices['items'] as $index) {
         $name = $index['name'];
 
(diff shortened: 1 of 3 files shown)


dev/cloud/teardown.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+use Magento\Framework\App\Bootstrap;
+require '/app/app/bootstrap.php';
+
+$bootstrap = Bootstrap::create(BP, $_SERVER);
+$obj = $bootstrap->getObjectManager();
```
0livier:
```suggestion
$objectManager = $bootstrap->getObjectManager();
```

dev/cloud/teardown.php
```diff
@@ -0,0 +1,42 @@
...
+
+$algoliaHelper = $obj->get('\Algolia\AlgoliaSearch\Helper\AlgoliaHelper');
```
0livier:
```suggestion
$algoliaHelper = $objectManager->get('\Algolia\AlgoliaSearch\Helper\AlgoliaHelper');
```

dev/cloud/teardown.php
0livier: This code is simple and not that long, so people may not confound the purpose of that variable here (a manager of things for a thing), but I do believe it's a good practice to always remove as much ambiguity as possible, even if the name of the variable is a bit longer.

dev/cloud/teardown.php
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @param $algoliaHelper Algolia\AlgoliaSearch\Helper\AlgoliaHelper
+ * @param array $indices
+ */
+function deleteIndexes($algoliaHelper, array $indices)
```
0livier:
```suggestion
function deleteIndexes($algoliaHelper, array $indices, $magentoCloudIndexPrefix)
```

dev/cloud/teardown.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    foreach ($indices['items'] as $index) {
+        $name = $index['name'];
+
+        if (mb_strpos($name, getenv('MAGENTO_CLOUD_ENVIRONMENT')) === 0) {
```
0livier:
```suggestion
        if (mb_strpos($name, $magentoCloudIndexPrefix) === 0) {
```

dev/cloud/teardown.php
```diff
@@ -0,0 +1,42 @@
...
+                // Might be a replica
+            }
+        }
+    }
+}
+
+if ($algoliaHelper) {
+    $indices = $algoliaHelper->listIndexes();
+    if (count($indices) > 0) {
+        deleteIndexes($algoliaHelper, $indices);
```
0livier:
```suggestion
        deleteIndexes($algoliaHelper, $indices, getenv('MAGENTO_CLOUD_ENVIRONMENT'));
```

dev/cloud/teardown.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    $replicas = $algoliaHelper->listIndexes();
+    if (count($replicas) > 0) {
+        deleteIndexes($algoliaHelper, $replicas);
```
0livier:
```suggestion
        deleteIndexes($algoliaHelper, $replicas, getenv('MAGENTO_CLOUD_ENVIRONMENT'));
```

dev/cloud/teardown.php
0livier: If a function uses a `getenv`, it means it has a [tight coupling](https://williamdurand.fr/2013/07/30/from-stupid-to-solid-code/#tight-coupling) with way the script is launched. I believe it's good practice to avoid that. Trying to apply the same principles in every snippet of PHP would allow us to copy/move code from scripts into our main code base without worrying.

PR #866, opened 2019-10-01
Fix #520 (disabled autocomplete with active facet query rule)

### LINKED ISSUE #520: Results div disappearing with autocomplete menu deactivated and facet query rules used
A bug
Deactivate the autocomplete menu and create a facet query rule on one particular attribute 
Type an exact value of one of the possible value of this attribute
Facet selects automatically (facet query rule) and query is considered as empty

The div containing the results disappear

Div should remain visible

Fix in js that handles this case (when a facet is selected, the div should always be visible)

v 1.7 (same issue with M1 extension)

damcou:


view/frontend/web/instantsearch.js
```diff
@@ -70,7 +70,7 @@ requirejs(['algoliaBundle','Magento_Catalog/js/price-utils'], function(algoliaBu
 		 * Docs: https://community.algolia.com/instantsearch.js/
 		 **/
 		
-		var ruleContexts = ['']; // Empty context to keep BC for already create rules in dashboard
+		var ruleContexts = ['magento_filters', '']; // Empty context to keep BC for already create rules in dashboard
```
bsuravech: Crazy that we didn't add this context before!

PR #868, opened 2019-10-02
Add extension notifier

damcou:


Model/Observer/ExtensionNotification.php
```diff
@@ -0,0 +1,147 @@
...
+    private function setLastCheckedVersion($version)
+    {
+        $this->cacheManager->save($version, 'algoliasearch_notification_lastchecked_version');
+
+        return $this;
+    }
+
+    private function checkExtensionVersion()
+    {
+        $versionFromRepository = $this->getLatestVersionFromRepository()->name;
```
0livier: In case the call is not working (e.g. Github down, or customer has a dump/strict firewall), I believe it should fall back nicely (e.g. try/catch or something).

damcou: I added a try/catch on the new model

Issue #831, opened 2019-08-12
Update operations extremely high


jasuten:
**Bug: What is the current behavior?**

We have around 15k products, around 50 custom attributes, several price groups and catalog price rules. We are getting update record operations over 500k-1m on days that are busy with product and inventory updates.

The search indexes are all set to schedule, we can see that the algolia_products_cl table can have 40 duplicates of a single product every time a product is saved, which is 40 update operations on the same product.

**Bug: What is the expected behavior?**

Update record operations should be a lot lower.

**Bug: What is the proposed solution?**

The Magento mview tables (i.e. algolia_products_cl, catalog_product_flat_cl) do not have a unique key set on the entity_id column. We've made a little extension internally that adds a unique key to the column to stop duplicates. This is more of an issue with Magento but as Algolia uses the default indexing logic it is affected by it.

We also do not show prices on our search results (B2B site) & show out of stock products, but there is no easy way to remove subscriptions in the mview.xml from watching the catalog_product_index_price & cataloginventory_stock_item tables unnecessarily. It would be great to be able to have options to control if this isn't requires to cut down on update operations as well. At the moment we had to make a local copy of the extension and modify it instead of installing through composer.

These two changes for us reduced update operations from 500k to 600 per day.

**Feature: What is your use case for such a feature?**

Reduce monthly expenses and improve site performance.

**Feature: What is your proposed configuration entry? The new option to add? What is the behavior?**

New options to control which tables are utilised for mview subscriptions. Possibly also adding a unique key for just the algolia_products_cl table until Magento fixes their cl tables.

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**

Using Magento 2.3.1, latest Algolia. Tested the same issue occurs on 2.3.2.

damcou:
Hello @jasuten ,

Thanks for reaching out.
Your use case seems to be interesting and we would be very interested to know what are the ecxact changes you made on your local copy of the extension and which files are affected.

It will help us a lot to see if we can do something "global" to add to the extension. 

Let us know.
Best.

jasuten:
Hi @damcou ,

Here is our attempt at fixing the issue for our scenario https://github.com/rival-ecommerce/magento-mviewfix

One drawback is that if a product is updated, then indexed, then updated again soon after the product isn't added to the changelog tables until they get cleared (every hour by Magento). Which isn't an issue for us but could be for others.

damcou:
[...]

PR #870, opened 2019-10-02
Remove catalog index price update by schedule subscription

unicoder88:
Hi there!

It saddens me to inform that I discovered a bug with update by schedule which results to a lot of reindex operations in Algolia.
In PR https://github.com/algolia/algoliasearch-magento-2/pull/666 , we subscribed to `catalog_product_index_price` in order to see changed catalog rule prices.
In _Magento 2.3_ this still works well when it comes to individual product updates, but it goes mad with full `catalog_product_price` reindex.

This is likely the reason behind reported issue https://github.com/algolia/algoliasearch-magento-2/issues/831 .

# Steps to reproduce

1. In Magento 2.3.1, install latest Algolia and switch all indexes to update by schedule mode.
  Initially `algolia_products_cl` is empty.
2. Reindex prices, e.g. `bin/magento indexer:reindex catalog_product_price`
3. Check `algolia_products_cl` and find that all products IDs were added - meaning full Algolia reindex!
4. Truncate or delete records from `algolia_products_cl`, or note last record number.
5. Reindex prices 2-nd time, e.g. `bin/magento indexer:reindex catalog_product_price`
6. Check `algolia_products_cl`, and see that no new records were added.
7. Reindex prices 3-rd time, e.g. `bin/magento indexer:reindex catalog_product_price`
8. Check `algolia_products_cl`, and see that all product IDs were added again.

# What happens:

Reason here is `catalog_product_index_price_replica` table.

In scheduled mode, _triggers_ are originally created on `catalog_product_index_price`,
and `catalog_product_index_price_replica` is empty.

https://prnt.sc/pdzjdh
https://prnt.sc/pdzjl6

In this state (scenario A), full price reindex does following:
1. truncates `catalog_product_index_price_replica`.
2. fills it with fresh index data.
3. switches `catalog_product_index_price_replica` vs `catalog_product_index_price`.
During this switch, triggers follow renamed tables - triggers are now assigned to `catalog_product_index_price_replica` table.

https://prnt.sc/pdzjvw

In scenario A, no triggers are fired, even if some prices actually changed.

With switched table and triggers now assigned to `_replica` table, next full reindex does following:
1. truncates `catalog_product_index_price_replica`. On delete triggers won't fire with TRUNCATE command.
2. fills it with fresh index data. --- problem here!
On insert triggers add every product in catalog to `algolia_products_cl` table. https://prnt.sc/pdzkhz
3. switches `catalog_product_index_price_replica` vs `catalog_product_index_price`.
Triggers follow renamed table and are now assigned to original `catalog_product_index_price` table - scneatio A.

And so with every full price reindex, scenario A - no updates at all - alternates with scneario B - all products are added to changelog.

Full price index update can happen quite often, and some modules even invalidate this index on every product save (Amasty_Shopby)!

I have not come up with a proper solution here yet, so we had to install an alternative module that makes a plugin to catalog rules reindexing and detects product price changes, and then inserts changed product IDs into `algolia_products_cl` table manually. But this is not generic enough.

# Cleaning up

Removing subscription from mview.xml doesn't seem to delete evil triggers, so I had to come up with schema migration. Should any other unfortunate code subscribe to the same index, the triggers will get recreated on next setup:upgrade recurring run.

What do you guys think?


Setup/UpgradeSchema.php
```diff
@@ -563,6 +563,12 @@ public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $con
             $connection->createTable($table);
         }
 
+        if (version_compare($context->getVersion(), '1.11.3', '<')) {
+            $connection->dropTrigger('trg_catalog_product_index_price_after_delete');
+            $connection->dropTrigger('trg_catalog_product_index_price_after_insert');
+            $connection->dropTrigger('trg_catalog_product_index_price_after_update');
+        }
```
bsuravech: Not sure if we should drop the trigger here in this case as there could be other subscriptions on these tables. It would be best to advise our customers to temporarily switch out their Indexer Mode from `Update on Schedule` to `Update on Save` for **algolia_products** and back to remove all triggers and add the ones they need.

Magento definitely did not make this easy but that would be the safest way.

unicoder88: Hi @bsuravech, thanks for checking!

When preparing this PR, first thing I tired was to switch to update on save, then back to update by schedule mode. However, the triggers were not removed for me, which is why I had to make a migration. I'm fairly certain that Magento creates missing triggers automatically every time we run `bin/magento setup:upgrade` - this is part of [recurring setup](https://github.com/magento/magento2/blob/2.3.2/app/code/Magento/Indexer/Setup/RecurringData.php#L47).

I will try again within following days and report.

unicoder88: Good day @bsuravech !

Like I promised, I tried again. Here's a video:

https://vimeo.com/368516103
Password: algolia

If we don't drop triggers, price index subscription remains even when we try switching indexer mode to Update on save and back to Update by schedule. I checked why:

[\Magento\Indexer\Controller\Adminhtml\Indexer\MassOnTheFly::execute](https://github.com/magento/magento2/blob/2.3.3/app/code/Magento/Indexer/Controller/Adminhtml/Indexer/MassOnTheFly.php#L30)
[\Magento\Indexer\Model\Indexer::setScheduled](https://github.com/magento/magento2/blob/2.3.3/app/code/Magento/Indexer/Model/Indexer.php#L293)
[\Magento\Framework\Mview\View::unsubscribe](https://github.com/magento/magento2/blob/2.3.3/lib/internal/Magento/Framework/Mview/View.php#L224)

The code iterates subscriptions, and removes them. However, our subscription **does not exist** any more in mview.xml, and so it's not removed from triggers.

If we drop triggers, they will be recreated on subsequent recurring data install:

[\Magento\Indexer\Setup\RecurringData::install](https://github.com/magento/magento2/blob/2.3.3/app/code/Magento/Indexer/Setup/RecurringData.php#L52)

Demo video with dropping triggers:

https://vimeo.com/368524421
Password: algolia

Please review :)

bsuravech: Ah, I see, the process workflow should be change to update on save, then install the upgrade/patch, and then switch back to update on schedule. 

Seeing as the subscription handles the creation / removal of these mview triggers, would it be better to replicate the old subscription to remove it using the subscription factory? See suggested. 

```suggestion
$subscriptionInstance = $this->subscriptionFactory->create(
    [
        'view' => $this->getMviewView()->load('algolia_products'),
        'tableName' => 'catalog_product_index_price',
        'columnName' => 'entity_id',
        'subscriptionModel' => SubscriptionFactory::INSTANCE_NAME,
    ]
);
$subscriptionInstance->remove();
```
getMviewView() would be an instance of \Magento\Framework\Mview\ViewInterface to load algolia_products View. 

I haven't tested this out myself but remove() method from Subscription would handle any linkedViews. I would feel more comfortable with this method.

unicoder88: Good catch @bsuravech !

I updated the code and tested when updating existing subscriptions and when installing Algolia for the first time.

bsuravech: Amazing @unicoder88 ! Thanks for your effort on this!

peterjaap:
@bsuravech @damcou @unicoder88 although I agree this PR in itself is good, it's not a *fix* in the sense that it fixes the underlying problem, right?

How are products which prices are updated because of catalog price rules now added to the queue for resyncing?

peterjaap:
@bsuravech still unsure about this one - it seems to me there's still no valid solution for a price rule kicking in not updating the products in Algolia, right?

PR #876, opened 2019-10-03
chore: ignores and fixes some phpstan mistakes

nunomaduro:
This pull request fixes and ignores some of the phpstan mistakes so we can start including phpstan in our CI pipeline.


.php_cs
```diff
@@ -1,7 +1,11 @@
 <?php
 
 /** @var PhpCsFixer\Config $config */
-$config = require __DIR__ . '/../../../.php_cs.dist';
+if (isset($_SERVER['argv']) && $_SERVER['argv'][3]) {
```
nunomaduro: Makes phpcs work with symlinks.

PR #878, opened 2019-10-03
Fix errors in code instead ignore in PHPStan

vmalyk:
**Summary**
Related with PR #876. Fixed errors in code instead adding to ignore.


vmalyk:
@damcou, I see failed test and it`s related with file 
https://github.com/algolia/algoliasearch-magento-2/blob/f6e047091e0676578abcc72d1b0776598f73e231/Controller/Adminhtml/Landingpage/Duplicate.php#L78-L80

Do you think it`s actual code or irrelevant?

damcou:
@vmalyk : Very interesting, this seems to be an oversight on our side :( . Would it be possible for you just to remove the "else" statement ? Thanks a lot for catching this.

nunomaduro:
@damcou Removing that `else` may be a regression. Would be cool run a QA that passes over this specific code to see if there is no regressions.

damcou:
@nunomaduro I'll make some tests on my side today about this to be sure it doesn't break anything but I'm pretty sure this is something I missed during the development .

vmalyk:
@damcou PR has been updated and I`be just removed "else' statement.

Controller/Adminhtml/Query/Save.php
```diff
@@ -84,9 +93,6 @@ public function execute()
 
             if (isset($data['banner_image'][0]['name']) && isset($data['banner_image'][0]['tmp_name'])) {
                 $data['banner_image'] = $data['banner_image'][0]['name'];
-                $this->imageUploader = \Magento\Framework\App\ObjectManager::getInstance()->get(
-                    'Algolia\AlgoliaSearch\QueryImageUpload'
-                );
                 $this->imageUploader->moveFileFromTmp($data['banner_image']);
```
damcou: Hello @vmalyk . For some reason, with this change, it breaks on save with this errror : 

`Exception #0 (BadMethodCallException): Missing required argument $baseTmpPath of Algolia\AlgoliaSearch\Model\ImageUploader.`

Seems that the imageUploader has to be defined in the di.xml as it's done for [ImageUpload](https://github.com/algolia/algoliasearch-magento-2/blob/develop/etc/di.xml#L97-L101) .

When I add the same lines below line 101, for  
`<type name="Algolia\AlgoliaSearch\Controller\Adminhtml\Query\Save">`,
the error seems to disappear.

vmalyk: @damcou , yes, it`s right. I added to di.xml injection ImageQueryUploader

nunomaduro:
✨

PR #889, opened 2019-10-11
Add config warnings for ES and MSI

damcou:


vmalyk:
@damcou,  @bsuravech it’s cool to move some parts of comparability with Magento modules to sub-modules. I have some small suggestions about  that PR: 
1.Replace literal class names to :class  - It’s Magento code style guidelines.
2. Add “suggest” section to composer.json as reference for developers.
3. Re-implement comparability with Staging in Magento Commerce Edition and move that to additional sub-module.

Thnx!

ViewModel/Adminhtml/Common.php
```diff
@@ -217,6 +229,44 @@ public function isClickAnalyticsTurnedOnInAdmin()
...
     }
 
+    public function isMysqlUsed()
+    {
+        $isMysql = false;
+        if ($this->configHelper->getBackendSearchEngine() === 'mysql') {
+            $isMysql = true;
+        }
+
+        return $isMysql;
```
0livier: Would the following work?
```return $this->configHelper->getBackendSearchEngine() === 'mysql';```

Comment (damcou) [reply ViewModel/Adminhtml/Common.php]: Yes totally, sorry the function was more complex at the beginning and I forgot to simplfy it /o\ .

**Comment by damcou** [comment]
@vmalyk : Thanks a lot for your suggestions, you're right, we can still improve a lot of things. I add them into our backlog so we can have a look at it ASAP.

File: Helper/ConfigHelper.php [hunk 2]
```diff
@@ -271,6 +272,11 @@ public function useSecureUrlsInFrontend($storeId = null)
         );
     }
 
+    public function getBackendSearchEngine($storeId = null)
+    {
+        return $this->configInterface->getValue(self::BACKEND_SEARCH_ENGINE, ScopeInterface::SCOPE_STORE, $storeId);
+    }
+
```
bsuravech: We can pull this information from \Magento\Search\Model\EngineResolver::getCurrentSearchEngine in Common.php instead of getting the information here.

ViewModel/Adminhtml/Common.php
```diff
@@ -217,6 +229,39 @@ public function isClickAnalyticsTurnedOnInAdmin()
         return $this->configHelper->isClickConversionAnalyticsEnabled();
     }
 
+    public function isMysqlUsed()
+    {
+        return $this->configHelper->getBackendSearchEngine() === 'mysql';
+    }
```
bsuravech: \Magento\Search\Model\EngineResolver::CATALOG_SEARCH_MYSQL_ENGINE const can be used here instead of matching against string.

PR #892, opened 2019-10-17
chore: increases phpstan level to 1

nunomaduro:
This pull request increases the level of PHPStan to: **1**.

This will increase the type safety on the future pull requests to the extension. 💪🏻

Note for @damcou @bsuravech: Some magento core classes use PHP magic `__call` methods - as example, on the code bellow the method `getPageId` don't really exists within the class. This means that from now we need to be explicit about them, adding the `@method int getPageId()` annotation.

```php
/**
 * @method int getPageId()
 */
class LandingPage extends Result {}
```


nunomaduro:
Requires `algolia/magento2-tools:^v0.1.6`.

Helper/Data.php
```diff
@@ -553,10 +553,6 @@ private function getCategoryRecords($storeId, $collection, $potentiallyDeletedCa
                 unset($potentiallyDeletedCategoriesIds[$categoryId]);
             }
 
-            if (isset($categorysToIndex[$categoryId]) || isset($categorysToRemove[$categoryId])) {
```
damcou: I think there was a mistake here and the line should be 
`
if (isset($categoriesToIndex[$categoryId]) || isset($categoriesToRemove[$categoryId])) {`

PR #895, opened 2019-10-18
chores(insights): migration / order conversion tracking 

bsuravech:
**Summary**
Order conversion logic needed to store queryID and other required parameters for conversion event. 

Additionally, I added some missing things for the analytics migration:
 - update configuration defaults for hits classes to reflect ISv3 classes 
 - update normal hits to get the indexName like the infinite hits one. 

**Result**
Added new column **algoliasearch_query_param** to the `quote_item` and `sales_order_item` tables. Used two events:

`catalog_controller_product_init_before` -  to store the request parameters to checkout session on product page load 
`checkout_cart_product_add_after` - to set quote_item after add from data set in checkout session

Created a plugin for \Algolia\AlgoliaSearch\Plugin\QuoteItem to convert the algoliasearch_query_param from quote_item to sales_item.

Outputted the event data a template called by onepage/success to set an object `algoliaOrderConversionJson` that will be iterated by click_conversion_analytics.js.


Setup/UpgradeData.php
```diff
@@ -0,0 +1,68 @@
...
+                    'type' => TABLE::TYPE_TEXT,
+                    'nullable' => true,
+                    'comment' => 'Reference for Algolia analytics order conversion',
+                ]
+            );
+
+            /** @var \Magento\Sales\Setup\SalesSetup $salesSetup */
+            $salesSetup = $this->salesSetupFactory->create(['resourceName' => 'sales_setup', 'setup' => $setup]);
+            $salesSetup->addAttribute(
+                'order_item',
```
damcou: Shouldn't it be sales_order_item ?

bsuravech: \Magento\Sales\Setup\SalesSetup::$_flatEntityTables contains 'order_item' as the array key for the correct table name.

PR #896, opened 2019-10-21
Added alt attribute for img tags

vmalyk:
**Summary**

ALT attribute is missed on Algolia Search page or Category page. 

**Result**

ALT attribute has been added in Hit template and for categories and images in autocomplete menu.


damcou:
Thanks for this @vmalyk  !

PR #899, opened 2019-10-22
Fix product image helper method to return set images 

bsuravech:
Updated the getProductImage method to return thumbnail set if found or fetch appropriate image configured from simple configurable products.

Helspcout Issue #3019721

### DIFF
--- a/Helper/Image.php
+++ b/Helper/Image.php
@@ -70,7 +70,7 @@ class Image extends \Magento\Catalog\Helper\Image
             if ($this->getImageFile()) {
                 $model->setBaseFile($this->getImageFile());
             } else {
-                $model->setBaseFile($this->getProductImage());
+                $model->setBaseFile($this->getProductImage($model));
             }
         }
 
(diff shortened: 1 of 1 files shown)


Helper/Image.php
```diff
@@ -81,14 +81,20 @@ protected function initBaseFile()
...
+     * @return mixed|string|null
      */
-    private function getProductImage()
+    private function getProductImage(\Magento\Catalog\Model\Product\Image $model)
     {
-        $imageUrl = $this->getProduct()->getImage();
-        if (!$this->getImageFile() && $this->getType() !== 'image'
-            && $this->getProduct()->getTypeId() == ProductTypeConfigurable::TYPE_CODE) {
-            $imageUrl = $this->getConfigurableProductImage() ?: $imageUrl;
+        if ($this->getImageFile()) {
```
damcou: I don't really understand why you added this condition here as it's already checked on line70 and it seems to be the only place where the method getProductImage() is called (in the else statement)

bsuravech: I must have been tired. I can cut it down to save some rows :D

jefflenz:
This is still an issue with 2.3.2-p2 Enterprise with latest version

PR #901, opened 2019-10-26
Fix version to 1.12.1 for unsubscribe Mview migration (PR #870)

vmalyk:
**Summary**

Migration isn't applied for installed version 1.12.0 and fix won't be worked after upgrade to 1.12.1 or higher. 

**Result**
Changed version for migration to 1.12.1 ( the same version as set in PR`s milestone field)


vmalyk:
@bsuravech, please review this PR.

Thank you!

Issue #913, opened 2019-11-13
Search result page error.


luan123z:
Hello guys, I'm using Magento 2.3.1 with the latest version of Algolia Search.
I've completely configured it but I cannot search for any products.
When I hit the search button, it directs me to the error page.
"there has been an error processing your request"

Below is my log file, can you please have a look?
[1533785306247.txt](https://github.com/algolia/algoliasearch-magento-2/files/3839213/1533785306247.txt)

bsuravech:
Hi @luan123z,

Thanks for reaching out to us. The error message is related to our Facets configuration which is being thrown because this configuration is empty. We will create a pull request to add a condition to check if this is empty to prevent any further errors in this scenario. 

If you would like to configure the Facets, you can do so by going to **Stores > Configuration > Algolia Search > Instantsearch Results Page > Facets**. 

We'll tag this issue in the following PR for your review when it is ready. 

Thanks!

luan123z:
> Hi @luan123z,
> 
> Thanks for reaching out to us. The error message is related to our Facets configuration which is being thrown because this configuration is empty. We will create a pull request to add a condition to check if this is empty to prevent any further errors in this scenario.
> 
> If you would like to configure the Facets, you can do so by going to **Stores > Configuration > Algolia Search > Instantsearch Results Page > Facets**.
> 
> We'll tag this issue in the following PR for your review when it is ready.
> 
> Thanks!

Hi @bsuravech 

I've configured Facets http://prntscr.com/pwo7pa but the error is still there.

However, there is one strange error while saving the config: http://prntscr.com/pwo7u3

damcou:
Hello there, I'm closing the issue as we didn't hear from you for a while. In case you still experience the issue after the release, feel free to respond and re-open the issue. Thank you!

Issue #910, opened 2019-11-06
Warning: count(): Parameter must be an array ... in Index.php on line 749


mikaelmattsson:
After selecting a filter and then reloading the page, making the url look like this:
`?country_of_manufacture=Italien~Frankrike`

I get an error
```

Exception #0 (Exception): Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/project/vendor/algolia/algoliasearch-client-php/src/AlgoliaSearch/Index.php on line 749
--
  |  
  | Exception #0 (Exception): Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/project/vendor/algolia/algoliasearch-client-php/src/AlgoliaSearch/Index.php on line 749
  | <pre>#1 count() called at [vendor/algolia/algoliasearch-client-php/src/AlgoliaSearch/Index.php:749]
  | #2 AlgoliaSearch\Index->getAlgoliaFiltersArrayWithoutCurrentRefinement() called at [vendor/algolia/algoliasearch-client-php/src/AlgoliaSearch/Index.php:733]
  | #3 AlgoliaSearch\Index->getDisjunctiveQueries() called at [vendor/algolia/algoliasearch-client-php/src/AlgoliaSearch/Index.php:653]
  | #4 AlgoliaSearch\Index->searchWithDisjunctiveFaceting() called at [vendor/algolia/algoliasearch-client-php/src/AlgoliaSearch/Index.php:615]
  | #5 AlgoliaSearch\Index->search() called at [vendor/algolia/algoliasearch-magento-2/Helper/AlgoliaHelper.php:115]
  | #6 Algolia\AlgoliaSearch\Helper\AlgoliaHelper->query() called at [vendor/algolia/algoliasearch-magento-2/Helper/Data.php:133]
  | #7 Algolia\AlgoliaSearch\Helper\Data->getSearchResult() called at [vendor/algolia/algoliasearch-magento-2/Helper/AdapterHelper.php:74]
  | #8 Algolia\AlgoliaSearch\Helper\AdapterHelper->getDocumentsFromAlgolia() called at [vendor/algolia/algoliasearch-magento-2/Adapter/Algolia.php:108]
  | #9 Algolia\AlgoliaSearch\Adapter\Algolia->query() called at [vendor/magento/module-search/Model/SearchEngine.php:42]
  | #10 Magento\Search\Model\SearchEngine->search() called at [vendor/magento/framework/Search/Search.php:85]
[...]

Issue #858, opened 2019-09-20
Accessing a not configured CATEGORY_CUSTOM_RANKING value is problematic


ThomasNegeli:
**Bug: What is the current behavior?**
Cronjob indexer_update_all_views stores the following message in the cron_schedule table 
```Warning: array_filter() expects parameter 1 to be array, boolean given in /var/www/src/releases/72/src/vendor/algolia/algoliasearch-magento-2/Helper/ConfigHelper.php on line 603```

**Bug: What is the expected behavior?**
If no ```CATEGORY_CUSTOM_RANKING```is configured, this should not raise a warning.

**Bug: What is the proposed solution?**
Check values before using them

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**
Magento 2.2.9 Commerce B2B
Algolia 1.11.3

bsuravech:
Hi @ThomasNegeli - Thanks for reporting! We have added it to a pull request to cast the empty value for customRanking configuration to return an empty array to prevent this warning message.

damcou:
Hi there, the fix has been merged, feel free to apply to your Magento instance if needed.
The next version will be released in January 2020.

Thanks for your feedback.
Best.

PR #914, opened 2019-11-13
Minor updates for optimization

bsuravech:
1) adding array_unique to remove duplicate productIds in indexing
2) add condition to check if facet configuration has value (#913)
3) update numericFilters parameter to pass array for count() php 7.2 error (#910)
4) add condition to check if customRankings configuration for product and category has value (#858)


bsuravech:
closed in favor of #932

Issue #919, opened 2019-11-14
Algolia_AlgoliaSearch breaks Magento_GoogleShoppingAds admin page


Pirosauro:
**Do you want to request a *feature* or report a *bug*?**
Bug

**Bug: What is the current behavior?**
Algolia_AlgoliaSearch breaks Magento_GoogleShoppingAds admin page. (https://devdocs.magento.com/extensions/google-shopping-ads/)

**Bug: What is the expected behavior?**
Navigate to Marketing > Advertising Channels > Google Shopping ads. A page should appear, unfortunately AlgoliaSearch includes prototype (legacy_build) which breaks the page load.

**Bug: What is the proposed solution?**
.

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**
Magento 2.2.4 
Algolia_AlgoliaSearch (algolia/algoliasearch-magento-2) 1.12.1 
Magento_GoogleShoppingAds 3.0.1

damcou:
Hi there, 
The fix has been merged, feel free to apply to your Magento instance if needed. 
The next version will be released in January 2020.

Thanks for your feedback.
 Best.

PR #921, opened 2019-11-15
Admin notices refactoring

damcou:
Refactored the "messy" common.phtml template, now :

- common.phml is split now between **common.phtml** (included in every admin page) and **configuration.phtml** (included only in system config page)
- Most of the logic in now handled by **NoticeHelper** and **AssetHelper** rather that the view model.
- Notices are now completely handled inside the **NoticeHelper** and no longer directly in the template.
- common.phtml was removed from default layout cause it broke Google Shopping Page (https://github.com/algolia/algoliasearch-magento-2/issues/919)


view/adminhtml/templates/common.phtml
```diff
@@ -312,15 +37,15 @@ document.addEventListener("DOMContentLoaded", function(event) {
...
-              new Ajax.Request(
-                url,
-                {
-                    parameters: {'store_id': storeId, 'query_text': value, 'query_id': queryId},
-                    method: 'post',
-                    asynchronous: false,
-                    onComplete: function (transport) {
-                    var response = JSON.parse(transport.responseText);
-                    isValid = response.isValid;
+            new Ajax.Request(
```
bsuravech: @damcou Is the reason why we are keeping prototype for this Ajax class? Wouldn't be better to convert this to jquery's Ajax? https://api.jquery.com/jquery.ajax/

view/adminhtml/templates/configuration.phtml
```diff
@@ -0,0 +1,149 @@
...
+            algoliaEventsTracking.track();
+
+		});
+	});
+
+	requirejs([
+		'jquery',
+		'Magento_Ui/js/lib/validation/validator',
+		'mage/translate',
+		'prototype'
```
bsuravech: We can remove prototype from here if we're not using it.

view/adminhtml/templates/configuration.phtml
```diff
@@ -0,0 +1,149 @@
...
+				if(selectValue !== '1') {
+					$input.hide();
+				}
+				else {
+					$input.show();
+				}
+			}
+
+			// EVENT TRACKING
+            algoliaEventsTracking.track();
```
bsuravech: This needs to go to the common file so that we track all events. It might be best to put it with tracking.phtml just after the object if not common.

view/adminhtml/templates/configuration.phtml
damcou: Actually, if you remove it, the field is not disabled (I guess this is a priority issue between prototype and jquery (?) )

PR #935, opened 2019-12-02
Use ::class instead string name in Job classes

vmalyk:
**Summary**
Hi!

Improved small difference in code styling between new classes for Jobs and another models, resource models and collections. In addition, it`s using ::class - it` support of Magento Code style. 

Thnx!


vmalyk:
@damcou , can you look?

Issue #941, opened 2019-12-04
getLastCheck causes segfault on server


peterjaap:
Our backend suddenly gave a 502 after flushing the cache.

After following processes on the server, I found that it segfaulted. I `strace`-d the process and this is what I found;

```
read(12, "", 8192)                      = 0
read(12, "", 8192)                      = 0
close(12)                               = 0
mmap(NULL, 2097152, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f2052c00000
madvise(0x7f2052c00000, 2097152, MADV_HUGEPAGE) = 0
mmap(NULL, 2097152, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f2052a00000
madvise(0x7f2052a00000, 2097152, MADV_HUGEPAGE) = 0
access("/data/web/releases/20191203111529/vendor/magento/theme-adminhtml-backend/Magento_Security/templates/page/activity_link.phtml", F_OK) = -1 ENOENT (No such file or directory)
access("/data/web/releases/20191203111529/vendor/magento/module-security/view/adminhtml/templates/page/activity_link.phtml", F_OK) = 0
lstat("/data/web/releases/20191203111529/vendor/magento/module-security/view/adminhtml/templates/page/activity_link.phtml", {st_mode=S_IFREG|0664, st_size=721, ...}) = 0
lstat("/data/web/releases/20191203111529/vendor/magento/module-security/view/adminhtml/templates/page", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat("/data/web/releases/20191203111529/vendor/magento/module-security/view/adminhtml/templates", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat("/data/web/releases/20191203111529/vendor/magento/module-security/view/adminhtml", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat("/data/web/releases/20191203111529/vendor/magento/module-security/view", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
stat("/data/web/releases/20191203111529/vendor/magento/module-security/view/adminhtml/templates/page/activity_link.phtml", {st_mode=S_IFREG|0664, st_size=721, ...}) = 0
[...]

PR #949, opened 2019-12-09
Use Guzzle to for version checking

damcou:
Two changes in the ExtensionNotification class :
- Use Guzzle to perform the call to Github API
- Log the exception if there's any

### DIFF
--- a/Model/ExtensionNotification.php
+++ b/Model/ExtensionNotification.php
@@ -3,29 +3,25 @@
 namespace Algolia\AlgoliaSearch\Model;
 
 use Algolia\AlgoliaSearch\Helper\ConfigHelper;
+use GuzzleHttp\Client;
 use Magento\Framework\App\CacheInterface;
 
 class ExtensionNotification
 {
     const CHECK_FREQUENCY = 86400; // one day
 
-    const REPOSITORY_URL = 'https://api.github.com/repos/algolia/algoliasearch-magento-2/releases/latest';
+    const REPOSITORY_URI = '/repos/algolia/algoliasearch-magento-2/releases/latest';
 
-    const CURL_OPT = [
-        'http' => [
-            'method' => 'GET',
-            'header' => [
-                'User-Agent: PHP',
-            ],
-            'timeout' => 10,
-        ],
-    ];
+    const APPLICATION_JSON_HEADER = ['Content-Type' => 'application/json'];
 
     /** @var CacheInterface */
     protected $cacheManager;
 
     /** @var ConfigHelper */
     private $configHelper;
+
+    /** @var Client */
+    private $guzzleClient;
 
     private $repoData = null;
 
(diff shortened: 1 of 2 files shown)


peterjaap:
LGM!

vmalyk:
Hi!

I have couple concerns about this PR:
1) By default Guzzle isnt`t present in  Magento Composer requirements, But might be to add in composer.json
2) Using Guzzle for 1 request - it looks overhead for me and might be better way -  use Magento Framework classes for HTTP client factory ([here](https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/HTTP/ClientFactory.phpl) and used [CURL class](https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/HTTP/Client/Curl.php) . In this case you won`t add extra requirement from Point 1.

@damcou , @bsuravech ,
Please review,
Thanks.

peterjaap:
@vmalyk 

Yes it is:

```
composer why guzzlehttp/guzzle
magento/framework                  102.0.3  requires  guzzlehttp/guzzle (^6.3.3)  
magento/product-community-edition  2.3.3    requires  guzzlehttp/guzzle (^6.3.3) 
```

vmalyk:
@peterjaap ,

but it`s not used for magento 2.2.x.

peterjaap:
Yes it's introduced in 2.3.3. So we could just add this as a requirement for this extension for people who run out-dated M2 versions, and then it'll be installed automatically.

peterjaap:
But I'm not set on having Guzzle at all. It's just a nice library to use for this stuff and it's included anyway for the newest M2 version so I'm all in favor of using it.

damcou:
Hello @peterjaap and @vmalyk , 

Thanks again for your help on this topic.
I think that adding guzzle as a composer requirement shouldn't be a problem and I'd stick to this solution if it's confirmed that it solves the https://github.com/algolia/algoliasearch-magento-2/issues/941 issue.

vmalyk:
@damcou , please check latest commit - looks as now we have duplicated "require" section. Can you fix and add guzzle to top section?

Thnx!

damcou:
@vmalyk Thanks again for spotting that :) . It's fixed.

Model/ExtensionNotification.php
```diff
@@ -128,12 +138,19 @@ private function checkExtensionVersion()
...
+                    'headers' => self::APPLICATION_JSON_HEADER,
+                ]
             );
-            $this->repoData = json_decode($json);
+
+            if ($response->getStatusCode() != 200) {
+                throw new \Exception($response->getReasonPhrase());
+            }
+
+            $this->repoData = json_decode($response->getBody()->getContents());
```
Serializator: Magento has its own interface for (de)serialization (`Magento\Framework\Serialize\SerializerInterface`) with a specific implementation for JSON (de)serialization (`Magento\Framework\Serialize\Serializer\Json`)

bsuravech: Love that this suggestion is coming from the Serializator.

damcou: Thanks a lot for your suggestion @Serializator , I made the requested change ;)

PR #974, opened 2020-01-28
fix(ISv4): set hierarhicalUiState to current path

tkrugg:
**Summary**
This change injects the `algoliaConfig.request.path` inside the hierarchicalMenu state to that it shows refined on category pages.

**Result**


view/frontend/web/instantsearch.js
```diff
@@ -99,17 +100,6 @@ requirejs(['algoliaBundle', 'Magento_Catalog/js/price-utils'], function (algolia
 			routing: window.routing,
 		};
 
-		if (algoliaConfig.request.path.length > 0 && window.location.hash.indexOf('categories.level0') === -1) {
```
tkrugg: hey @bsuravech note that in this PR I've removed `areCategoriesInFacets` case because I don't think it's still relevant, but please if you think otherwise let me know and I'll add it back.

PR #977, opened 2020-02-10
Update B2B / Catalog Permissions raw SQL 

bsuravech:
B2B/Catalog permissions update raw SQL to use db methods. While updating, I found some condition errors that didn't filter visibility correctly. This should now index the correct permission value and filtering properly.

### DIFF
--- a/Factory/SharedCatalogFactory.php
+++ b/Factory/SharedCatalogFactory.php
@@ -48,7 +48,7 @@ class SharedCatalogFactory
     {
         return $this->objectManager->create('\Magento\SharedCatalog\Model\ResourceModel\Permission');
     }
-
+    
     public function getSharedCatalogResource()
     {
         return $this->objectManager->create('\Magento\SharedCatalog\Model\ResourceModel\SharedCatalog');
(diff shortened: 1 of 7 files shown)


Factory/SharedCatalogFactory.php
```diff
@@ -65,14 +65,13 @@ public function getSharedCategoryCollection()
...
-                WHERE customer_group_id IN (SELECT customer_group_id FROM shared_catalog) 
-                GROUP BY category_id;
-            ";
-
-            $this->sharedCategoryCollection = $connection->fetchPairs($query);
+            $select = $connection->select()
+                ->from($indexResource->getMainTable(), [])
+                ->columns('category_id')
+                ->columns(['permissions' => new \Zend_Db_Expr("GROUP_CONCAT(CONCAT(customer_group_id, '_', permission) SEPARATOR ',')")])
+                ->group('category_id');
```
damcou: Is it expected that the WHERE condition hasn't been ported ?

bsuravech: Updated!

PR #994, opened 2020-03-17
Personalization Frontend 

bsuravech:
**M2 Personalization**

The Personalisation feature on Magento 2 has changes that will affect how Click and Conversion Analytics works in the extension. The reason is that the insights client funnels all events to both Personalisation and Analytics depending on if they contain specific attributes. 

Personalisation will accept all events while Click and Conversion Analytics will only accept events that contain `queryID`.

This means that the extension has to review how Click and Conversion Analytics is handled alongside Personalisation as all events selected for CC will be funneled into Personalisation. 

As CC and Personalisation are combined, I have made the decision to remove the click_and_conversion_analytics.js for a general insights.js file to handle CC and personalisation events. 

You can read more about the implementation in the internal doc here:

https://docs.google.com/document/d/1wjjQkkHsQDFIh_MwNzGn76SRAQVd4hTPA8alMBpCPHk/edit#

**Case** | **Pass/Fail**
-- |--
CC Analytics Enabled / Disabled |
Personalisation Enabled / Disabled |
**Events** |  
View Product |
Click Product |
Click Product (Perso Only) |
Click Recommended Product |
Click Facet |
Conversion Wishlist |
Conversion Add to Cart |
Conversion Add to Cart (Perso Only) |
Conversion Place Order |
Conversion Place Order (Perso Only) |
**UserToken** |  
Before logging in see if anonymous userToken is being passed into click/converted events |
set UserToken by logging into customer, `aa-search` cookie should be set with new userToken |


Model/Observer/Insights/CheckoutOnepageControllerSuccessAction.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * @param Observer $observer
+     *
+     * @return $this|void
+     */
+    public function execute(Observer $observer)
+    {
+        $this->logger->info('event fired');
```
damcou: Do you want to keep those logs ?

Helper/InsightsHelper.php
```diff
@@ -0,0 +1,167 @@
...
+class InsightsHelper
+{
+    const ALGOLIA_ANON_USER_TOKEN_COOKIE_NAME = '_ALGOLIA';
+
+    const ALGOLIA_CUSTOMER_USER_TOKEN_COOKIE_NAME = 'aa-search';
+
+    /** @var ConfigHelper */
+    private $configHelper;
+
+    /** @var Configuration\PersonalizationHelper */
```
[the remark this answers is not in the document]
bsuravech: I think I missed this one ahah

Model/Observer/Insights/CheckoutOnepageControllerSuccessAction.php
bsuravech: whooops. I forgot to clean this one up.

Model/Observer/Insights/CheckoutCartProductAddBefore.php
```diff
@@ -0,0 +1,49 @@
...
+        /** @var \Magento\Catalog\Model\Product $product */
+        $product = $observer->getEvent()->getProduct();
+        $requestInfo = $observer->getEvent()->getInfo();
+
+        if (!$this->configHelper->isClickConversionAnalyticsEnabled($product->getStoreId())
+            || $this->configHelper->getConversionAnalyticsMode($product->getStoreId()) !== 'place_order') {
+            return;
+        }
+
+        if (isset($requestInfo['queryID'])) {
```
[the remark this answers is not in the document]
bsuravech: I'll add a check for it. Insights does complain if queryID is empty.

Helper/InsightsHelper.php
```diff
@@ -0,0 +1,167 @@
...
+     * @return string|null
+     */
+    private function getUserToken()
+    {
+        $userToken = $this->cookieManager->getCookie(self::ALGOLIA_CUSTOMER_USER_TOKEN_COOKIE_NAME);
+        if (!$userToken) {
+            if ($this->customerSession->isLoggedIn()) {
+                // set logged in user
+                $userToken = $this->setUserToken($this->customerSession->getCustomer());
+            } else {
```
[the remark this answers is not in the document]
bsuravech: For readability, I kept it as if/else :P

Model/Observer/Insights/CheckoutCartProductAddAfter.php
```diff
@@ -0,0 +1,80 @@
...
+     * @param Observer $observer
+     * ['quote_item' => $result, 'product' => $product]
+     */
+    public function execute(Observer $observer)
+    {
+        /** @var \Magento\Quote\Model\Quote\Item $quoteItem */
+        $quoteItem = $observer->getEvent()->getQuoteItem();
+        /** @var \Magento\Catalog\Model\Product $product */
+        $product = $observer->getEvent()->getProduct();
+
```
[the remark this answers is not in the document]
bsuravech: I stored the variable :)

Block/Configuration.php
```diff
@@ -209,6 +211,36 @@ public function getConfiguration()
...
+                        'eventName' => 'Recommended Product Clicked',
+                        'enabled' => $persoHelper->isProductRecommendedTracked(),
+                        'selector' => $persoHelper->getProductRecommendedSelector(),
+                        'method' => 'clickedObjectIDs',
+                    ],
+                ],
+                'filterClicked' => [
+                    'eventName' => 'Filter Clicked',
+                    'enabled' => $persoHelper->isFilterClickedTracked(),
+                    'method' => 'clickedFilters',
```
tkrugg: Are the the only events the user can send ?
For a pure person perspective it's best that the user can freely pick their event names. I understand there are tradeoff magento needs to make to help the setup of person but I'm curious on how easy it can be to add move event names

view/adminhtml/templates/configuration.phtml
```diff
@@ -62,6 +62,11 @@ document.addEventListener("DOMContentLoaded", function(event) {
 					$('#algoliasearch_cc_analytics_cc_analytics_group table').hide();
 				}
 
+				// Hide personalization config if not available
+				if (personalizationStatus < 2) {
```
tkrugg: from a JS good practice perspective, JS developers prefer explicit the meaning of the constant:
```js
var PERSONALIZATION_NON_AVAILABLE_STATUS = 0;
var PERSONALIZATION_AVAILABLE_NOT_ENABLED_STATUS = 1
if (personalizationStatus === PERSONALIZATION_NON_AVAILABLE_STATUS || PERSONALIZATION_AVAILABLE_NOT_ENABLED_STATUS) { 
   ...
 }
```

Even better if we expose the constant from the backend to the frontend in one common place (it's more searchable in the codebase)

view/frontend/templates/instant/facet.phtml
```diff
@@ -1,7 +1,7 @@
 <script type="text/template" id="refinements-lists-item-template">
-    <div class="{{cssClasses.label}} {{#isRefined}}checked{{/isRefined}}">
-        <input class="{{cssClasses.checkbox}}" {{#isRefined}}checked{{/isRefined}} type="checkbox">
+    <label class="{{cssClasses.label}} {{#isRefined}}checked{{/isRefined}}">
+        <input class="{{cssClasses.checkbox}}" {{#isRefined}}checked{{/isRefined}} type="checkbox" value="{{value}}" />
         {{value}}
         <span class="{{cssClasses.count}}">{{count}}</span>
-    </div>
+    </label>
```
tkrugg: Just to make sure: this is potentially a breaking change. Is that alright?

view/frontend/web/insights.js
```diff
@@ -0,0 +1,250 @@
...
+            }
+        },
+
+        initializeAnalytics: function() {
+            algoliaAnalytics.init({
+                appId: this.config.applicationId,
+                apiKey: this.config.apiKey
+            });
+
+            var userAgent = 'insights-js-in-magento (' + this.config.extensionVersion + ')';
```
tkrugg: what's `this.config.extensionVersion` ? is it refering to the version of the `algoliasearch-magento-2` or something else?

view/frontend/web/insights.js
```diff
@@ -0,0 +1,250 @@
...
+            algoliaAnalytics.addAlgoliaAgent(userAgent);
+
+            var userToken = getCookie('aa-search');
+            if (userToken && userToken != '') algoliaAnalytics.setUserToken(userToken);
```
tkrugg: it's best to always use triple equal sign operators: `userToken !== ''` instead of `userToken != ''`. 

loose equality/inequality are often hard to reason on. 

```
false != 0   // -> false 
false !== 0 // -> true
```

view/frontend/web/insights.js
```diff
@@ -0,0 +1,250 @@
...
+                        containers.push('.' + elem.className);
+                    }
+
+                    algolia.registerHook('afterInstantsearchStart', function (search, algoliaBundle) {
+                        var selectors = document.querySelectorAll(containers.join(', '));
+                        selectors.forEach(function (e) {
+                            e.addEventListener('click', function (event) {
+                                var attribute = this.dataset.attr;
+                                var elem = event.target;
+                                if (elem.matches("input[type=checkbox]") && elem.checked) {
```
tkrugg: `matches` will break on IE 9: see https://caniuse.com/#feat=matchesselector
why not use jquery like you did earlier with clickedObjectIDs ? 

jQuery has an API that you can use cross browser: https://api.jquery.com/is/

view/frontend/web/insights/add-to-cart-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+                && algoliaConfig.ccAnalytics.conversionAnalyticsMode == 'place_order'
+            ) {
+                this._setQueryIdToForm(form)
+            }
+
+            this._super(form);
+        },
+
+        _setQueryIdToForm: function(form) {
+            if (form.find('button[type="submit"]').length
```
tkrugg: is it really `find` ?
I would have expected to have hear `form.querySelector()` instead. Although I can't really tell if form is a DOM element. Is it?

view/frontend/web/insights/add-to-cart-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+                && form.find('button[type="submit"]').data('queryid')) {
+                var queryID = form.find('button[type="submit"]').data('queryid');
+            }
+
+            var queryID = queryID || this._getQueryParamFromCurrentUrl('queryID');
+            if (queryID.length == 0) {
+                return;
+            }
+
+            if (form.find('input[name="queryid"]').length == 0) {
```
tkrugg:
```suggestion
            if (form.find('input[name="queryid"]').length === 0) {
```

view/frontend/web/insights/add-to-cart-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+                form.prepend('<input type="hidden" name="queryID" />');
```
tkrugg: you can maybe just directly 

```
form.prepend('<input type="hidden" name="queryID" value="' + queryID +  '" />'); 
```

and skip the next instruction.

view/frontend/web/insights/add-to-cart-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+            }
+
+            form.find('input[name="queryID"]').val(queryID);
+        },
+
+        _getQueryParamFromCurrentUrl: function(queryParamName) {
```
tkrugg: Maybe this function could be called something more generic as it seems to be able to handle more than just eh queryID?
Something like `parseURL(window.location.href)`.

view/frontend/web/internals/search-insights.js
```diff
@@ -1 +1 @@
...
...wedObjectIDs=function(e){if(!e)throw new Error("No params were sent to viewedObjectIDs function, please provide `objectIDs` to be reported");if(!e.objectIDs)throw new Error("required objectIDs parameter was not sent, view event can not be properly sent without");this.sendEvent("view",e)}.bind(this),this.viewedFilters=function(e){if(!e)throw new Error("No params were sent to viewedFilters function, please provide `filters` to be reported");if(!e.filters)throw new Error("required filters parameter was not sent, view event can not be properly sent without");this.sendEvent("view",e)}.bind(this)}({requestFn:function(){if(t())return I;if(r())return m;throw new Error("Could not find a supported HTTP request client in this environment.")}()});return"undefined"!=typeof window&&function(e){var t,r=e.AlgoliaAnalyticsObject;if(r){var n=(t=this,function(e){for(var r=[],n=arguments.length-1;n-- >0;)r[n]=arguments[n+1];e&&s(t[e])&&t[e].apply(t,r)}),o=e[r];o.queue=o.queue||[];var i=o.queue;i.forEach(function(e){var t=[].slice.call(e),r=t[0],o=t.slice(1);n.apply(void 0,[r].concat(o))}),i.push=function(e){var t=[].slice.call(e),r=t[0],o=t.slice(1);n.apply(void 0,[r].concat(o))}}}.call(g,window),g});
```
tkrugg: This is version 1.3.0. The latest version is 1.4.0. If it's not too much trouble it might be worth updating.
The code is here:
```
https://cdn.jsdelivr.net/npm/search-insights
```

Block/Configuration.php
bsuravech: You can configure the events for view and conversion options as well. The ones identified in the config object are for the click events. 

Regarding the ability to choose their event name, I find the ones that we supplied are very generic enough for the events we identified. I think one of the things we can do to improve this is to translate the event name so that they can be translated to their language or label of choice through Magento's translate feature. 

The requirements for this implementation didn't include a custom section to create other events to track other things. Perhaps this something we can revisit in P2.

view/frontend/templates/instant/facet.phtml
bsuravech: This is and will be updated in the docs as an advisory :/

view/frontend/web/insights.js
bsuravech: Yes, this is to add the extension version to the userAgent!

view/frontend/web/insights.js
bsuravech: Aha this is the same best practice for PHP. Will update this!

view/frontend/web/insights.js
bsuravech: Converted it use : `$(elem).is("input[type=checkbox]")`

view/frontend/web/insights/add-to-cart-mixin.js
bsuravech: It's a jQuery element at this point of the mixin for `mage.catalogAddToCart`.

view/frontend/web/insights/add-to-cart-mixin.js
bsuravech: I think I did this to ensure that if there was already an existing queryID input, that the value would be updated. I'm going to keep as is.

view/adminhtml/templates/configuration.phtml
bsuravech: I think we'll keep as is as it's for the backend side. This status check isn't used anywhere else at the moment. Still great insights for us to know about constant usages in JS!

Block/Configuration.php
tkrugg: Yes definitely. I think the event set right now if very complete to benefit from all the features perso opens. Thanks for clarifying!

PR #1012, opened 2020-05-11
Fix error on URL during Pages indexing

palviggi-flagbit:
**Summary**

* Set store scope using correct function to avoid using Default store domain
* Get a clean UrlInterface instance to avoid problem where domain from previous store in the loop is used.


bsuravech:
HI @palviggi-flagbit ! 

Thanks for the pull request! Our CI for quality tools has indicated some code styling errors. As this is a forked PR, would it be possible for you to correct/lint so we can merge your contribution? Thank you!

PR #1015, opened 2020-05-14
make sure original price range is saved for configurables

palviggi-flagbit:
**Summary**
When indexing configurables, the Original / Discount relationship is lost, This saves the original prices to have the discount visible


damcou:
Hello @palviggi-flagbit , thanks for submitting this PR.
Looks like the code you modified/added doesn't pass the quality-tools check.
Could you please update it ?
Here is the tool you can use to check its validity : https://github.com/algolia/algoliasearch-magento-2/blob/develop/.github/CONTRIBUTING.md#coding-style

Thanks again

PR #1021, opened 2020-05-26
getSalesData refactor, prevent ... IN ('0') query

whadziewicz:
**Summary**
On systems with many orders, 100k+, indexing `ordered_qty` with "IN ('0')" can cause heavy load on MySQL engine.

**Result**
After refactor, empty collection skips searching in `sales_order_item` table

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -739,25 +739,29 @@ class Data
             return [];
         }
 
-        $ordersTableName = $this->resource->getTableName('sales_order_item');
+        $salesData = [];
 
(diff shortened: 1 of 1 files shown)


bsuravech:
Hey @whadziewicz ! 

Thanks for the PR! This is a good optimisation and will prevent unnecessary load for sales data. I see that there is a raw SQL that we would like to convert. I will take over this PR and make those changes. Thank you so much for your contribution!

bsuravech:
Update! We'll merge in the PR and reopen a new one for the raw SQL fix. Thanks!

PR #1025, opened 2020-06-01
Removed obsolete trigger for catalog_product_entity_media_gallery

vmalyk:
Hi!

**Summary**

Magento version: 2.2.x-2.3.x  

Sometimes We get  reports from clients about stuck  on re-index products from Magento to Algolia. All indices are "On Schedule" mode. After research We found reason - It's related with trigger to "catalog_product_entity_media_gallery" table and value_id column. If Images are updated often on huge catalog you can get in algolia_products_cl table next case:   
 version_id	entity_id,
14	17500,
15	17500,
16	17500,
17	17500,
18	17500,
19	17500,
20	17500,
21	17500,
22	17500,
*23	1193238,* (it's value_id and it blocks updates)
24	17500, 
25	17500

after checking in Magneto code, I found difference - This trigger was added in Magento 2.1 and It was removed in Magento 2.2.0:
Magento 2.1: https://github.com/magento/magento2/blob/2.1/app/code/Magento/Catalog/etc/mview.xml 
Magent 2.2 (since 2.2.0-RC1.1) : https://github.com/magento/magento2/commit/9fe3b8a0cdff5c9f4677c1890ca2bf52718acae5#diff-b82ecede1475a66d3135943308251166 

**Result**

I've removed obsolete mview subscription to wrong column.

Please review,Thnx

### DIFF
--- a/etc/mview.xml
+++ b/etc/mview.xml
@@ -27,7 +27,6 @@
             <table name="catalog_product_entity_decimal" entity_column="entity_id" />
             <table name="catalog_product_entity_gallery" entity_column="entity_id" />
             <table name="catalog_product_entity_int" entity_column="entity_id" />
-            <table name="catalog_product_entity_media_gallery" entity_column="value_id" />
             <table name="catalog_product_entity_media_gallery_value" entity_column="entity_id" />
             <table name="catalog_product_entity_text" entity_column="entity_id" />
             <table name="catalog_product_entity_tier_price" entity_column="entity_id" />


vmalyk:
@betty, @damcou ,

Looks as develop-1.x branch has outdated circle.yml configuration.
Can you check?

Thnx!

bsuravech:
Hi @vmalyk can you pull down from our develop-1.x branch? I have merged in a backport PR for the CI changes. Thanks!

bsuravech:
Hi @vmalyk ! 

As we are removing a subscription from tables, it would be best to add a script to clean up any old triggers for `catalog_product_entity_media_gallery`. For example:

https://github.com/algolia/algoliasearch-magento-2/blob/develop/Setup/UpgradeSchema.php#L591-L603

vmalyk:
Hi @bsuravech!

Yes, it's good suggestion, but it's actual for Magento 2.2.x only, because since Magento 2.3.0 it has been fixed and Unsubscribe/Subscribe for Mview run on each setup:upgrade:
https://github.com/magento/magento2/blob/2.3/app/code/Magento/Indexer/Setup/Recurring.php#L122-L125
Another thing, if We add that in UpgradeSchema.php - We need to implement 2 different versions - for 1.x and 2.x.  I suggest to create Recurring.php and add both unsubscription (catalog_product_entity_media_gallery and catalog_product_index_price). In this case, it'll be work independed by version and code'll be the same for both versions.

Please review,
Thnx!

Issue #1024, opened 2020-05-28
Exception #0 (Exception): Warning: Division by zero


aversalle:
**Do you want to request a *feature* or report a *bug*?**

report a *bug*

**Bug: What is the current behavior?**

An exception is thrown

**Bug: What is the expected behavior?**

The configuration page should display

**Bug: What is the proposed solution?**

Add a default value other than zero to resolve this issue for new installations with new configuration value yet set

**What is the version of Magento and of Algolia extension you are using? Always use the latest version of the extension one before opening a bug issue.**

Magento ver. 2.3.4-p2
"algolia/algoliasearch-magento-2": "^2.0"

Stack Trace:

1 exception(s):
Exception #0 (Exception): Warning: Division by zero in /Users/aversalle/Sites/nagase-america/vendor/algolia/algoliasearch-magento-2/Helper/Configuration/NoticeHelper.php on line 108

Exception #0 (Exception): Warning: Division by zero in /Users/aversalle/Sites/nagase-america/vendor/algolia/algoliasearch-magento-2/Helper/Configuration/NoticeHelper.php on line 108
<pre>#1 Algolia\AlgoliaSearch\Helper\Configuration\NoticeHelper->getQueueNotice()
#2 call_user_func() called at [vendor/algolia/algoliasearch-magento-2/Helper/Configuration/NoticeHelper.php:86]
#3 Algolia\AlgoliaSearch\Helper\Configuration\NoticeHelper->__construct() called at [vendor/magento/framework/ObjectManager/Factory/AbstractFactory.php:121]
#4 Magento\Framework\ObjectManager\Factory\AbstractFactory->createObject() called at [vendor/magento/framework/ObjectManager/Factory/Dynamic/Developer.php:66]
#5 Magento\Framework\ObjectManager\Factory\Dynamic\Developer->create() called at [vendor/magento/framework/ObjectManager/ObjectManager.php:70]
#6 Magento\Framework\ObjectManager\ObjectManager->get() called at [vendor/magento/framework/ObjectManager/Factory/AbstractFactory.php:167]
#7 Magento\Framework\ObjectManager\Factory\AbstractFactory->resolveArgument() called at [vendor/magento/framework/ObjectManager/Factory/AbstractFactory.php:273]
#8 Magento\Framework\ObjectManager\Factory\AbstractFactory->getResolvedArgument() called at [vendor/magento/framework/ObjectManager/Factory/AbstractFactory.php:236]
[...]

PR #1031, opened 2020-06-05
Prevent division by zero in the Notice Helper

vmalyk:
Backport for 1.x, Issue  #1024


bsuravech:
Thanks for this @vmalyk !

PR #1061, opened 2020-07-10
Restore "search as you type" feature for v2

bsuravech:
In version 2.x, we upgrade instantsearch library from v2 -> v4. With this upgrade, the functionality of the searchBox widget has changed to only accept a container instead of using your own input field has it did before. This change was documented here:

https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/#searchbox

Screenshot 2020-07-10 at 15 40 03

With @eunjae-lee, we were able to restore this dropped feature by creating a customSearchBox.

HS Ticket: #271451

### DIFF
--- a/view/frontend/web/instantsearch.js
+++ b/view/frontend/web/instantsearch.js
@@ -586,6 +586,33 @@ requirejs(['algoliaBundle', 'Magento_Catalog/js/price-utils'], function (algolia
 			};
 		}
 
+		if (!algoliaConfig.autocomplete.enabled) {
+			var customSearchBox = algoliaBundle.instantsearch.connectors.connectSearchBox(function(renderOptions, isFirstRendering) {
+				if (isFirstRendering) {
+					var input = document.querySelector(instant_selector);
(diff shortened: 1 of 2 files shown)


view/frontend/web/instantsearch.js
```diff
@@ -586,6 +586,33 @@ requirejs(['algoliaBundle', 'Magento_Catalog/js/price-utils'], function (algolia
 			};
 		}
 
+		if (!algoliaConfig.autocomplete.enabled) {
+			var customSearchBox = algoliaBundle.instantsearch.connectors.connectSearchBox(function(renderOptions, isFirstRendering) {
+				if (isFirstRendering) {
+					var input = document.querySelector(instant_selector);
+					var clearBtn = input.parentElement.getElementsByClassName('clear-query-autocomplete')[0];
```
eunjae-lee, line 593: Is there a possibility that user could've changed the template around the clear button? (meaning `clearBtn` could be `undefined`?)

bsuravech: Could be possible, yes. I'll add a condition to check around it!

PR #1062, opened 2020-07-13
Reset the index connection on emulate to allow for different credentials per storeview

palviggi-flagbit:
**Summary**

When having multiple stores, with different appId/key per store and using the queue runner, it can happen that one store index ends up in the wrong account, because the index connection is created only once at the start.

This resets the index connection between runs during store emulation, to make sure to use the right credentials.

It might be possible that some other type of operation needs a similar treatment, I'm still checking for that.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -683,6 +683,9 @@ class Data
 
         $this->emulation->startEnvironmentEmulation($storeId, Area::AREA_FRONTEND, true);
         $this->scopeCodeResolver->clean();
+        $this->storeManager->setCurrentStore($storeId);
+        $this->algoliaHelper->resetCredentialsFromConfig();
+
         $this->emulationRuns = true;
 
         $this->logger->stop('START EMULATION');


palviggi-flagbit:
@bsuravech 
I found that this fix is incomplete (at least suggestions and settings can still get end up in the wrong app) but I don't have time assigned to finish it in the near time. 
Should I close the PR or mark it as draft?

damcou:
Hello @palviggi-flagbit ,
The extension isn't compatible with multi accounts configuration.
As you noticed, it would take much more effort to make it fully compatible, for example, the method `resetCredentialsFromConfig()` you use in your PR doesn't take the store id into account.

This improvement has already been identified and added on our roadmap, but we unfortunately don't have any ETA for this, in the meantime, feel free to close this PR.

palviggi-flagbit:
@damcou 
it doesn't take a store because it already reads the current store from the store manager. Since the manager is also used by other parts, it's better to make sure the correct id is set there.

If there's no close ETA for the fix, it would be a good idea to change in the meantime the credentials scope to global instead of per store, since it causes confusion. Also it would be a good idea to make it clear on the documentation.

PR #1066, opened 2020-08-05
[magento2.4-requirements]:

jeroenschipper-guapa:
- Added correct magento framework requirement for magento2.4

**Summary**
This PR adds the possibility to install the module with composer for Magento 2.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
   "version": "2.0.2",
   "require": {
     "php": "~7.0|~7.1|~7.2|~7.3",
-    "magento/framework": "~101.0|~102.0",
+    "magento/framework": "~101.0|~102.0|~103.0",
     "algolia/algoliasearch-client-php": "^2.4",
     "guzzlehttp/guzzle": "^6.3.3",
     "ext-json": "*",


pchoate-blueacorn:
@damcou any idea on the timeline to get this merged in? I have a client with a Magento 2.4 upgrade in progress but we're block by this module not supporting 2.4.

palviggi-flagbit:
is that enough? isn't there any deprecated classes that should be replaced? (especially with the new ES requirement). 
what about php 7.4? we also need to install a new 2.4 instance and would like to not having to update the system by next year but use the latest php

jasuten:
You will get an error when trying to compile the site with the extension, so there are more changes required to get Algolia working with 2.4. e.g
`Class Magento\Framework\Search\Adapter\Mysql\Mapper does not exist`

damcou:
Hello @jeroenschipper-guapa @pchoate-blueacorn @palviggi-flagbit @jasuten 

Just to let you know that the 2.4 compatibility subject has been discussed and will be implemented in our next major release. The ETA for this release is end of September.

Best.

PR #1068, opened 2020-08-25
remove less than IE9 condition from configuration template

bsuravech:
Having this condition here does more harm than good. Customers who have minified and strip HTML comments come across the algoliaBundle missing error often because of this script. 

Since we support IE 11 and higher, we should drop this.


damcou (approved):
👍

VivekShingala:
I am already on version 2.0.2. However, this file is not updated when I run `composer update` or `composer install`. How to get the updated files if I am already on the latest version?

matthewbond:
@VivekShingala This change is on the develop branch at the moment, it has not yet been merged into master: https://github.com/algolia/algoliasearch-magento-2/blob/master/view/frontend/templates/internals/configuration.phtml

PR #1071, opened 2020-08-26
Remove IdentityInterface from Indexing Queue Job model class

bsuravech:
HS Ticket: 283569

Pages are getting cached, but are getting uncached after some minutes, without any change due to the indexing queue cache tags. Identity class needs to be removed from the Indexing Queue Job entity as it is mostly static and not used for customer facing display: 

https://devdocs.magento.com/guides/v2.4/graphql/develop/identity-class.html


damcou (approved):
👍

PR #1072, opened 2020-08-26
Update for multi-condition rules

bsuravech:
JIRA Ticket: MAGE-6 

Convert rules to pass in `conditions` which is an array of `condition` parameters. Backward compatibility for `condition`.


damcou (approved):
👍

PR #1079, opened 2020-09-17
Fix serialization issue with 2.4

damcou:
- Added row ids "on the fly" for config serialized array when row id is missing
- Added some default values in config.xml


bsuravech (approved):
LGTM!

PR #1081, opened 2020-09-18
Update Personalisation Order Conversion 

bsuravech:
HS Ticket: 292821

If `$order` is not set by session, fall back to retrieve order from `order_ids`. 
This supports Magento 2 v2.2.x.


damcou (approved):
LGTM

PR #1082, opened 2020-09-18
Update subproduct adaptive images on color

bsuravech:
JIRA Ticket: MAGE-42

Add image type check for value to skip if placeholder for subproduct images for color.


damcou (approved):
LGTM

PR #1086, opened 2020-09-23
Add Customisation section to README.md 

bsuravech:
Add handy reference of library version for each versions we have of the extension and quick links to the docs to help with customisation.


damcou (approved):
🔥

PR #1093, opened 2020-10-01
Bump to v. 3.0.0

damcou:


bsuravech (changes requested):
Can we update the Customisation JS version table to match develop-2.x branch? Other than that, all looks good!

bsuravech (approved):
LGTM

PR #1096, opened 2020-10-13
Indexing queue grids fixes 

damcou:
- Removed status filtering on queue (MAGE-55)
- Fixed queue log export (MAGE-51)


bsuravech (approved):
👍 🔥 🎉

PR #1101, opened 2020-10-23
Bugfix: Configurable product with broken image when set to "no_selection"

dverkade:
**Summary**

Currently when a config product image is set to "no_selection" it will not look at the images of the child product to determine to images of the product.

**Result**

With the change the behaviour is as expected.


damcou:
@bsuravech Looks good to me, is it ok for you ?

bsuravech (approved):
👍

PR #1103, opened 2020-11-03
Add tags for merchandising queries to use VM if applicable

bsuravech:
[MAGE-63] Add rules tag for visual editor on Query Merchandising and Landing Page Builder features

This will allow the usage of the Visual Editor, if added to customer plan. Adding the visual editor tag will not effect rules for customers who do not have the Visual Editor enabled. 

**NOTE:**
Customers should know that if you edit the rule in the Algolia Dashboard, advanced options other than promote are not supported in the extension. Editing the rule in Magento will override changes in made in Algolia. Adding advance options means that the rule should only be edited in Algolia moving forward. 

[MAGE-63]: https://algolia.atlassian.net/browse/MAGE-63

### DIFF
--- a/Helper/MerchandisingHelper.php
+++ b/Helper/MerchandisingHelper.php
@@ -54,12 +54,16 @@ class MerchandisingHelper
             $condition['pattern'] = $query;
         }
 
-        if (! is_null($banner)) {
+        if (!is_null($banner)) {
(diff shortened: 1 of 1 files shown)


damcou (approved):
LGTM 🚀

PR #1105, opened 2020-11-09
infiniteHits update showMoreLabel to showMoreText

bsuravech:
Community member identified that the showMoreLabel was updated in ISv4 to templates.showMoreText. This has been corrected. 

https://discourse.algolia.com/t/showmorelabel-not-translatable-since-instantsearch-v3-magento-2/11337


damcou (approved):
🚀

PR #1111, opened 2020-11-19
Insights Analytics - Fix Add to Cart Conversion

bsuravech:
- Set the queryId value on product if add to cart form contains a queryID value. 
- Add to cart mixin should always add queryID for both add to cart and place order conversion.


damcou (approved):
LGTM 👍

PR #1112, opened 2020-11-20
Added try/catch on CheckoutSuccess event

damcou:
See MAGE-68


bsuravech (approved):
Looks good to mean. I understand the limit on the productIds and we shouldn't fire more than one event. Happy to approve this as is. Good catch!

PR #875, opened 2019-10-03
Reindex all CMS pages if page is changed

vmalyk:
PR for #874, simple solution

PR #1113, opened 2020-11-24
CMS Page indexing improvement

bsuravech:
Based on PR: #875 

This is to allow for CMS Pages to index sectionally instead indexing atomically.  
 * For indexing queue, merge `rebuildStorePageIndex` jobs if they have `page_ids`
 * Exclude pages in the collection query instead of through loop condition

Might need to clean this up to split jobs for full reindex vs sectional.


damcou (approved):
Works perfectly :)

PR #1114, opened 2020-11-24
Remove ES compat from suggestions

bsuravech:
The ES compatibility module was experimental. Currently is not compatible for ES7 adapter. Will add back when it's updated at some point. For now, we will remove it.


damcou (approved):
👍

PR #1119, opened 2021-01-04
Fetch Algolia additional data only in the extension's sections in the Magento configuration

damcou:


bsuravech (approved):
LGTM! Thanks for the PR!

PR #1120, opened 2021-01-06
Remove image URL manipulation

fredden:
**Summary**

`removeProtocol()` should not be required. This was introduced in 2017 when both HTTP & HTTPS were in general use. Nowadays, only HTTPS is in general use for ecommerce websites. Browsers will typically display passive mixed content such
as images. If websites are having trouble with mixed content this will not be only because of Algolia search.
https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content/How_to_fix_website_with_mixed_content
See also #211

`removeDoubleSlashes()` is introducing a bug: a URL which is a full path only is having an additional slash added to the start, producing a valid-but-wrong URL. For example, '/media//wysiwyg//search.png' becomes '//media/wysiwyg/search.png'. Browsers cope just fine with URLs containing multiple slashes, so no manipulation is required here.
See also #234

**Result**

Better handling of image URLs.

### DIFF
--- a/Helper/Entity/CategoryHelper.php
+++ b/Helper/Entity/CategoryHelper.php
@@ -295,9 +295,6 @@ class CategoryHelper
         ];
 
         if (!empty($imageUrl)) {
-            $imageUrl = $this->imageHelper->removeProtocol($imageUrl);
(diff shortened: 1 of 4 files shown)


damcou (approved):
LGTM 🚀

PR #1121, opened 2021-01-06
Fix to keep the price slider values in the filter when user refresh the page

damcou:


bsuravech (approved):
LGTM! Thanks for the PR!

PR #1123, opened 2021-01-26
Autocomplete add new hook for product source search options

bsuravech:
Add the ability to modify search options for autocomplete product source to allow ease of customisation. 

New event: `beforeAutocompleteProductSourceOptions`


damcou (approved):
LGTM

PR #1124, opened 2021-02-08
Fix Price calculation for products with/without tax and customer prices

damcou:
Reset price calculation, otherwise getFinalPrice, called in $this->getSpecialPrice will always return lowest price from last customer group round

Thanks to the community for the provided fix.


bsuravech (approved):
LGTM!

Johan-We-Provide:
Hi,

I've updated the algoliasearch-magento-2 module to version 3.1.0 and have version 2.3.5 of Magento 2 EE. But in the autocomplete the prices are still without tax while the result page has the prices with tax. Are there some settings where I have to set some preferences? Or is this due to that the result page is the result page of Magento itself?

PR #1132, opened 2021-03-04
Update Module Sequence and indexers

bsuravech:
- clean up indexers to remove `use Magento;`
- update sequence for Magento_Indexer


damcou (approved):
🚀

PR #1133, opened 2021-03-10
Fix insights missing indexName for autocomplete sections

bsuravech:
HS Ticket # 338995

Missing `__indexName` hit param for autocomplete sections: categories, suggestions, and additional sections indexes. This corrects the insights error on those items for click analytics.


damcou (approved):
LGTM

PR #1134, opened 2021-03-10
Refactor popular queries collection and mview class for additional_sections

bsuravech:
- Refactor `getPopularQueries()` by removing the unnecessary for loop and and slice. 
- Correct the mview class for `algolia_additional_sections` indexer
- remove duplicated whitelist connect-src policy


damcou (approved):
LGTM 🚀

PR #1135, opened 2021-03-10
Fix adaptive image issue for updated bundle

bsuravech:
- check to see if color object key is defined to prevent errors


damcou (approved):
🚀

PR #1136, opened 2021-03-12
Add small change on category indexing for Enterprise Edition

damcou:
Prevents a bug related to confusion between row_id and entity_id in case category isn't fetched (can happen if it's a root category) #337360


bsuravech (approved):
Good catch! 🚀

PR #1137, opened 2021-03-15
Refactor loops and fix analytics overview issues

bsuravech:
- refactor loops to use `array_column` for single key fetch
- fix issues with analytics overview and allow access if plan does not return from API


damcou (approved):
🚀

PR #1138, opened 2021-03-16
Fix rangeInput facet to use correct seperatorText and submitText templates

NickdeK:
**Summary**

When using the price range type for a filter that labels would not be translated. 
image

The code has been updated using the InstantSearch docs found here: https://www.algolia.com/doc/api-reference/widgets/range-input/js/#templates

**Result**

image

**Remarks**
There seems to be a currency label in the original code, but I can't find any references on this in the InstantSearch docs. We would love to have currency labels tho!


view/frontend/web/instantsearch.js
```diff
@@ -477,12 +477,10 @@ requirejs(['algoliaBundle', 'Magento_Catalog/js/price-utils'], function (algolia
 				return ['rangeInput', {
 					container: facet.wrapper.appendChild(createISWidgetContainer(facet.attribute)),
 					attribute: facet.attribute,
-					labels: {
-						currency: algoliaConfig.currencySymbol,
-						separator: algoliaConfig.translations.to,
-						button: algoliaConfig.translations.go
-					},
-					templates: templates,
+					templates: Object.assign({
```
bsuravech, line 480: `Object.assign()` is not supported in IE11. As we support Magento version 2.3 which does support IE11, I would recommend that you use the jQuery's `$.extend()` to merge the objects together.

bsuravech (changes requested):
Hi @NickdeK ! Thanks for the pull request! This is a good fix! I have one comment for you. I wish I didn't have to but we have to cover this usecase for the time being! Let me know if you have any questions! Thanks!

bsuravech:
Closing this PR due to inactivity. We've duplicated this ticket with review fixes here: PR #1148 . Thank you!

PR #1141, opened 2021-03-17
Fix max-width breakpoints overlapping min-width breakpoint

NickdeK:
**Summary**

Breakpoints mismatch when using both min-width and max-width on 992px. 
Screenshot 2021-03-17 at 15 53 38

**Result**

Setting the max-width breakpoint to 991px solves this isse.


bsuravech (approved):
Hi @NickdeK ! Thanks for the PR! Just tested and never noticed this issue before! Good catch on this!

PR #1142, opened 2021-03-18
Fix query value in custom instant search box

vmalyk:
Hey! 
**Summary**

Since 2.x releases  If enable Instant search and disable autocomplete - used custom searchBox. But if state has query - instant search has been re-rendered properly, but search box is empty.
 
**Result**

Value is set from searchBox renderOptions and now pre-filled by right value.

Please review,


bsuravech:
Hi @vmalyk ! I hope you're doing well! I just tested the change in our latest dev branch and it seems to be working well. Can you update your branch for us before we merge in? Thanks!

PR #1143, opened 2021-03-29
Removing Proxy features 

damcou:
Removing the following:

- Feature gating
- Config tracking
- Support form


bsuravech (changes requested):
This looks good! There's a few things. You missed some files regarding tracking:

- [x] `view/adminhtml/templates/tracking.phtml` << file can be removed
- [ ] `view/adminhtml/layout/default.xml`<< needs to remove block that references the above file

bsuravech (approved):
Looking good to go! 🚀

PR #1144, opened 2021-03-29
Add additional attributes for customer groups 

bsuravech:
If customer groups are enabled, the generatedAPI key filters need to include the attributes for autocomplete indexes as well. 

We shouldn't have the same generatedAPIKey for all datasets in autocomplete so this is a quickfix and needs to be evaluated later.


damcou (approved):
:rocket:

PR #1145, opened 2021-04-02
Add algolianet.com to csp whitelist

bsuravech:
Noticed a CSP error in the frontend and saw our search was querying from `*.algolianet.com` which is one of our fallbacks:

https://www.algolia.com/doc/guides/scaling/distributed-search-network-dsn/#technical-details


damcou (approved):
Good catch 👍

PR #1149, opened 2021-04-21
Product URL Model: Check request path to empty

vmalyk:
Hi!

In some cases   $product->getRequestPath() returns as NULL and URL Rewrite isn't loaded for product. As result, product gets url with /catalog/product/view/id/*** format. I suggest to use the same expression as in original file and
 check to empty value. 

Please review,


damcou (approved):
Thanks for this @vmalyk , LGTM 👍

PR #1152, opened 2021-05-21
add missing label for queue method rebuildCategoryIndex

tezlopchan:
**Summary**
Label map for queue method `rebuildCategoryIndex` is missing as a result, the queue job list method name is blank.
Screen Shot 2021-05-21 at 12 37 22 PM
 

**Result**
Screen Shot 2021-05-21 at 12 42 15 PM


damcou (approved):
Thanks for spotted this @tezlopchan , LGTM :)

PR #1154, opened 2021-07-19
Unfilter getCoreCategories() from private getCategoryById() 

bsuravech:
HS Ticket: #380520

If customer has disabled filtering Include in Category attribute for categories, the path names for these category types are coming up empty. Need to get categories regardless of this status by getting unfiltered core categories in `getCategoryById()` method.

### DIFF
--- a/Helper/Entity/CategoryHelper.php
+++ b/Helper/Entity/CategoryHelper.php
@@ -470,7 +470,7 @@ class CategoryHelper
 
     private function getCategoryById($categoryId)
     {
-        $categories = $this->getCoreCategories();
+        $categories = $this->getCoreCategories(false);
 
         return isset($categories[$categoryId]) ? $categories[$categoryId] : null;
     }


igorfigueiredogen:
Perfect!
It'll solve my problem.

Steps to reproduce my problem before the update:

Magento EE on Version: 2.3.6-p1 (It must be enterprise edition) 
Algolia in version: 3.1.0 

1) Create a new category (not included_in_menu) and watch it being indexed in algolia. 
2) Create a schedule to update the category created above 
3) After the schedule, create a new category (not included_in_menu and must not be children of the first category we created), and watch it being indexed in algolia. 
- You will see that the path is wrong because it's taking the name of the first category we created 
4) Delete the first category we created and force an update to the second category to be indexed in algolia again. 
- You will see that the path now is blank. 
5) Update the second category by including this in the menu, and watch it being indexed in algolia. 
- Now the path is right, but a should be able to index a category despite the fact that it is included in the menu or not.

igorfigueiredogen (approved):
Perfect!

PR #1155, opened 2021-07-28
Fix max original price for product with children

valeriish:
**Summary**

Fix of issue with displaying of configurable product special price on algolia search results page.
Product original formatted price is wrong.

**Result**


damcou (approved):
Good catch :+1:

PR #1156, opened 2021-08-03
Store scope Category Merchandising

bsuravech:
**Summary**
HS Ticket # 381786

Category Merchandising is not store scoped. Will save all store views with changes regardless of scope. 

**Result**
Event observer `\Algolia\AlgoliaSearch\Model\Observer\Merchandising` will now scope changes to the correct storeview, if selected. If no storeview, it will save to all.


damcou (approved):
LGTM

PR #1168, opened 2021-11-29
Add get/set methods for job instance

JeroenVanLeusden:
**Summary**
Add set/get methods to job class. We need this methods in order to use entity in API and Magento queue.

I've chosen to use `body` instead of `data` since `data` is already use in the global `getData()`/`setData()` methods.

**Result**
Able to use job instance in API and Magento queue.

### DIFF
--- a/Api/Data/JobInterface.php
+++ b/Api/Data/JobInterface.php
@@ -29,5 +29,48 @@ interface JobInterface
     const FIELD_RETRIES = 'retries';
     const FIELD_ERROR_LOG = 'error_log';
     const FIELD_DATA_SIZE = 'data_size';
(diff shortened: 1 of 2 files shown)


damcou (changes requested):
Hello @JeroenVanLeusden ,
This looks fine to me, but apparently some of the changes are not passing the phpcs check, could you make the changes please ? https://github.com/algolia/algoliasearch-magento-2/blob/develop/.github/CONTRIBUTING.md#coding-style

JeroenVanLeusden:
@damcou Fixed!

damcou (approved):
LGTM, thanks for this PR :)

PR #1169, opened 2021-12-07
fix: Update category template in autocomplete to better display categories with img

damcou:


t2hnd (approved):
Looks great, thanks for creating a PR for this!

PR #1170, opened 2022-01-31
Ability to disable Algolia analytics per IP-Address

JeroenVanLeusden:
**Summary**
Ability to exclude certain IP-Addresses from Algolia Analytics. See, https://support.algolia.com/hc/en-us/articles/4406975229969

**Result**
`analytics` parameter is added to search requests. Value is `false` if current IP-Address of customer is in list of added IP-Addresses, otherwise `true`.


JeroenVanLeusden:
Not fully finished, will update when I have the time.

PR #1207, opened 2022-06-24
MAGE-301 Remove mview update via patch

mohitalgolia:
**Summary**

**Result**


ratikant-singh (approved):

rachel-trott (changes requested):
Changes requested

Setup/Data/UpdateMviewPatch.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        return [];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public static function getVersion()
+    {
+        return '1.12.0';
```
cammonro, line 64: This seems problematic as it relates to the `setup_version` of the module at `3.6.1`. That being the case I would not expect this to execute. Is that intentional?

PR #1219, opened 2022-07-25
Updated readme for recommend

mohitalgolia:
**Summary**

**Result**


rachel-trott (changes requested):
- **Recommend:** Algolia Recommend lets you display recommendations such as "Frequently Bought Together" and "Related Products" features on the product detail page.

PR #1250, opened 2022-10-03
Fix saving and loading data by the wrong cache key for the popular queries

abramchukm:
**Summary**
Currently, the implementation of the method `\Algolia\AlgoliaSearch\Helper\Entity\SuggestionHelper::getPopularQueries` is incorrect, because it caches the data by a cache tag `$popularQueriesCacheId` which leads to the wrong popular queries search results on different stores.
**Example**
Let's take a look at the next example. 
You have a store with ID 1 and it has `suits, pants` popular query. Store with ID 2 has `Salon de jardin` popular query. After the cache is cleaned, you visit a store with ID 1, and you have a correct popular query search result (here, the result is cached by tag, specified in the `$popularQueriesCacheId` property). But, for the store with ID 2 it won't be a `Salon de jardin` anymore, but `suits, pants` which is incorrect behavior.
**Result**
A fix adds a `store_id` to the cache tag. This will provide correct results for the requested store and the situation described above will be fixed.

### DIFF
--- a/Helper/Entity/SuggestionHelper.php
+++ b/Helper/Entity/SuggestionHelper.php
@@ -96,7 +96,7 @@ class SuggestionHelper
 
     public function getPopularQueries($storeId)
     {
-        $queries = $this->cache->load($this->popularQueriesCacheId);
+        $queries = $this->cache->load($this->popularQueriesCacheId . '_' . $storeId);
         if ($queries !== false) {
             return $this->serializer->unserialize($queries);
         }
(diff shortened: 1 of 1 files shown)


rachel-trott:
Thank you for @abramchukm We will take a look at this

rachel-trott:
@abramchukm Thank you for the PR. We have brought it in as part of this, https://github.com/algolia/algoliasearch-magento-2/pull/1267

PR #1296, opened 2023-01-11
Bug - When MSI is disabled a full reindex event is created

shaneblandford:
**Summary**

The observer event algoliasearch_reindex_product_on_last_item_purchase_if_msi_disable is fired on each order. If MSI is disabled and the order contains items that are still in stock post order a full reindex event will be created. 

This is due to the array of product ids being empty.

**Result**

I created a simple check to ensure that there were productids in the array.


ratikant-singh:
Hi, @shaneblandford 
Thank you for reporting the issue, we are aware of this issue and we have already fixed it, and it will be available in the next release.

Please find the pull request detail here - https://github.com/algolia/algoliasearch-magento-2/pull/1291

PR #1305, opened 2023-02-02
Replace referenceBlock by referenceContainer following standard in adobe commerce / magento source code

lfluvisotto:
Replace referenceBlock by referenceContainer following standard in adobe commerce / magento source code

**Summary**

According to adobe commerce / magento source code:

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Theme/view/base/page_layout/empty.xml#L13

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Theme/view/base/page_layout/empty.xml#L20

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Theme/view/adminhtml/page_layout/admin-1column.xml#L45

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Theme/view/adminhtml/page_layout/admin-2columns-left.xml#L48

The names "main.content" and "before.body.end" are containers and not blocks so it should use referenceContainer and not referenceBlock.

### DIFF
--- a/view/adminhtml/layout/adminhtml_system_config_edit.xml
+++ b/view/adminhtml/layout/adminhtml_system_config_edit.xml
@@ -4,11 +4,11 @@
     <head>
         <link src="Algolia_AlgoliaSearch::js/config.js"/>
     </head>
-    <referenceBlock name="before.body.end">
+    <referenceContainer name="before.body.end">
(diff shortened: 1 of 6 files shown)


rachel-trott:
@lfluvisotto Thank you for bringing this up. We have added this to our backlog

PR #1307, opened 2023-02-06
Fix for Indexer created for websites excluded in Customer groups - (In Progress)

mohitalgolia:
Fix for Indexer created for websites excluded in Customer groups


cammonro (commented):
@rachel-trott @mohitalgolia Please see my comments with regards to release 3.13.2

PR #1309, opened 2023-02-21
Add primary key to algoliasearch_queue_archive table

maaarghk:
The MySQL documentation can explain to you why it is desirable to enforce primary keys on all tables:

https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_sql_require_primary_key

Observing the code it seems this table is inserted to using "INSERT..SELECT" with columns specified, so ID will populate automatically by auto increment. Haven't tested the alter table to see if it populates, leave that up to you. From memory should be fine

### DIFF
--- a/etc/db_schema.xml
+++ b/etc/db_schema.xml
@@ -34,6 +34,7 @@
         </constraint>
     </table>
     <table name="algoliasearch_queue_archive" resource="default" engine="innodb" comment="Algoliasearch Queue Archive Table">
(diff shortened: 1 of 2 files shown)


rachel-trott:
@maaarghk Thank you for bringing this to our attention. We are working on this along with some other fixes in this PR. https://github.com/algolia/algoliasearch-magento-2/pull/1324. Closing this PR.

PR #1310, opened 2023-02-24
Updated instantsearch code

mohitalgolia:
**Summary**

**Result**

### DIFF
--- a/Block/Configuration.php
+++ b/Block/Configuration.php
@@ -70,6 +70,8 @@ class Configuration extends Algolia implements CollectionDataSourceInterface
         $level = '';
         $categoryId = '';
         $parentCategoryName = '';
+        $currentCategory = '';
+        $childrenCategories = array();
 
         $addToCartParams = $this->getAddToCartParams();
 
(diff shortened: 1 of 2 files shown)


Block/Configuration.php
```diff
@@ -88,9 +90,13 @@ public function getConfiguration()
 
             if ($category && $category->getDisplayMode() !== 'PAGE') {
                 $category->getUrlInstance()->setStore($this->getStoreId());
-
+                $childCategories = $category->getChildrenCategories();
+                foreach ($childCategories as $childCategory){
```
cammonro, line 94: Unfortunately this will not handle nested categories - a recursive approach is needed.

view/frontend/web/instantsearch.js
```diff
@@ -420,48 +420,74 @@ requirejs(['algoliaBundle', 'Magento_Catalog/js/price-utils'], function (algolia
...
-						header: '<div class="name">' + (facet.label ? facet.label : facet.attribute) + '</div>',
-					}
-				};
+                };
+
+                // hierarchicalMenuParams.templates.item = '' +
+                // 	'<a class="{{cssClasses.link}} {{#isRefined}}{{cssClasses.link}}--selected{{/isRefined}}" href="{{url}}">{{label}}' + ' ' +
+                // 	'<span class="{{cssClasses.count}}">{{#helpers.formatNumber}}{{count}}{{/helpers.formatNumber}}</span>' +
+                // 	'</a>';
+                hierarchicalMenuParams.templates.item = '' + '<a class="{{cssClasses.link}} {{#isRefined}}{{cssClasses.link}}--selected{{/isRefined}}"' + 'href="{{categoryUrl}}">{{label}}' + ' ' + '<span class="{{cssClasses.count}}">{{#helpers.formatNumber}}{{count}}{{/helpers.formatNumber}}</span>' + ' ' + '</a>';
```
cammonro, line 485: Setting the URL via template is insufficient. We need to address the event handler in the underlying InstantSearch lib. For this a custom renderer will be needed via a [connector](https://www.algolia.com/doc/api-reference/widgets/hierarchical-menu/js/#connector).

cammonro (commented):
There are some merits to the idea behind this approach (i.e. obtaining the category navigation on the backend and feeding into the `hierarchicalMenu` widget). However ultimately we need to implement a custom renderer to override the click event. 

A stub for the intended functionality has been added to the [pending PR](https://github.com/algolia/algoliasearch-magento-2/pull/1351) for Merch Studio support with a narrowing of available categories via context sensitive anchoring through `transformItems`. At a future time we can enhance the click handler as needed with a new renderer.

PR #1321, opened 2023-03-09
Use getTable from ModuleDataSetupInterface to get table name with prefix

MarcusSchwarz:
This is a follow up to #1313 which handles the same issue for another patch.

**Summary**
If a table prefix is configured, the RecommendConfigPatch will throw an Error because the table 'core_config_data' (without prefix) can't be found.

**Result**
* After the Change, the RecommendConfigPatch can be applied if a table prefix ist configured.
* After the change an integration test sandbox can be set up with a table prefix.

### DIFF
--- a/Setup/Patch/Schema/RecommendConfigPatch.php
+++ b/Setup/Patch/Schema/RecommendConfigPatch.php
@@ -62,7 +62,7 @@ class RecommendConfigPatch implements SchemaPatchInterface
 
         $this->moduleDataSetup->getConnection()->startSetup();
         $connection = $this->moduleDataSetup->getConnection();
-        $table = $connection->getTableName('core_config_data');
+        $table = $this->moduleDataSetup->getTable('core_config_data');
         foreach ($movedConfigDirectives as $from => $to) {
             try {
                 $connection->query('UPDATE ' . $table . ' SET path = "' . $to . '" WHERE path = "' . $from . '"');


rachel-trott:
@MarcusSchwarz This has been released in 3.10.1. Thank you for the contribution

MarcusSchwarz:
Hello @rachel-trott , this is not the patch which got merged into 3.10.1. The code in question is still untouched, see
https://github.com/algolia/algoliasearch-magento-2/blob/b0fc4decce9c4d1c073f66bbac7031caf926299a/Setup/Patch/Schema/RecommendConfigPatch.php#L65

jean-bernard-valentaten:
@rachel-trott The file affected by the PR that has been merged into `3.10.1` (#1313 ) is another than then file affected by this PR here.

#1313: Patches file `Setup/Patch/Schema/AutocompleteConfigPatch.php`
This PR here patches file `Setup/Patch/Schema/RecommendConfigPatch.php`

Please reopen this PR and merge it as the current state prevents the module being installed from scratch when table prefixes are being used.

rachel-trott:
@jean-bernard-valentaten @MarcusSchwarz thank you for pointing this out. I have reopened the PR and we will bring this fix in the next release.

HyflerSD:
Hey @rachel-trott, is there a time set for the next release?

rachel-trott (approved):

PR #1325, opened 2023-03-22
Fix/mage 559

cammonro:
Apply a cap on results retrieved in Autocomplete from query suggestions index to apply the config `algoliasearch_autocomplete/autocomplete/nb_of_queries_suggestions`.

As suggestions are driven from the popularity scores from Magento in `search_query` be sure to run `bin/magento indexer:reindex algolia_suggestions` so that Algolia's index is in sync.


cammonro:
The PR is updated to preclude the need to transform the result set from the suggestions index.

PR #1340, opened 2023-04-11
Special price range for configurable products

bartlubbersen:
Correctly calculate min/max special price range for configurable products with price rules. Currently because of a change made in https://github.com/algolia/algoliasearch-magento-2/pull/1160/files if a configurable product contains a product with a special price this is no longer shown/indexed as a range but just a value. It should correctly show a min/max range of the special price.

This is adjusted in this pull request:

* During min/max calculation use getSpecialPrice function that is used for simple products as well
* Solve issue where configurable price manager override (getRulePrice) also overrides for simple products which caused errors
* Pass along subproducts so getRulePrice doesn't have to retrieve them. Retrieving was missing active/stock filters etc, which resulted in wrong price being displayed.
* Remove previous fix that was always overriding the range for a fixed value for configurable products, also it was causing the earlier calculated special price to be moved to original price

Before fix:
image

Correct after fix:
image


bartlubbersen:
Any update on the review? I personally patched this for my project, but this could cause issues for more people without even knowing it. I could update the MR if it's out of date.

denisveg:
@rachel-trott can you guys review this and fix it?

rachel-trott:
@denisveg @bartlubbersen I have added this to our backlog. Thanks for calling this out. We will review it.

PR #1341, opened 2023-04-17
MAGE-615 (formerly MAGE-612) RequireJS / Autocomplete v1 event model 

cammonro:
This PR aims to achieve two things:

1. Enforce RequireJS for all JavaScript libs to ensure proper dependency loading order
2. Provide a first pass at implementing new front end events for Autocomplete v1

Re: #2, it introduces the following new events:

* `afterAutocompleteSources`
* `afterAutocompletePlugins`
* `afterAutocompleteOptions`
* `afterAutocompleteProductSourceOptions`

and deprecates:

* `beforeAutocompleteOptions`
* `beforeAutocompleteProductSourceOptions`

The following event has been removed completely as any old implementation based on Autocomplete v0 will be incompatible with Autocomplete v1:

* `beforeAutocompleteSources`


view/frontend/web/internals/template/autocomplete/products.js
```diff
@@ -15,7 +15,8 @@ define([], function () {
 
         getItemHtml: function ({item, components, html}) {
             return html`<a class="algoliasearch-autocomplete-hit" href="${item.__autocomplete_queryID != null ? item.urlForInsights : item.url}"
-                data-objectId=${item.objectID} data-index=${item.__autocomplete_indexName} data-queryId=${item.__autocomplete_queryID}>
+                data-objectId=${item.objectID} data-index=${item.__autocomplete_indexName} data-queryId=${item.__autocomplete_queryID}
+                data-position=${item.__autocomplete_id + 1}>
```
cammonro, line 19: This logic seems suspect to me and will not derive a proper position index. I think we should extract that position at the `transformResponse` callback where we can access the full hits array or better yet let's look at using `createAlgoliaInsightsPlugin`.

PR #1370, opened 2023-06-15
Feature/mage 674

cammonro:
Faceting params for PLP were previously harvested via backed on initial IS render. This approach generates the params via frontend upon rerender of the product card.

### DIFF
--- a/view/frontend/web/instantsearch.js
+++ b/view/frontend/web/instantsearch.js
@@ -638,6 +638,18 @@ define(
 
             var isStarted = false;
 
+            // Capture active redirect URL with IS facet params for add to cart from PLP
+            search.on('render', () => {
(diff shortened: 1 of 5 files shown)


rachel-trott (approved):

view/frontend/web/instantsearch.js
```diff
@@ -639,16 +639,18 @@ define(
             var isStarted = false;
 
             // Capture active redirect URL with IS facet params for add to cart from PLP
-            search.on('render', () => {
-               const cartForms = document.querySelectorAll('[data-role="tocart-form"]');
-               cartForms.forEach((form, i) => {
-                   const ts = Date.now();
```
nisfulTM: `ts` constant is not actually being used anywhere

cammonro: Good catch - thanks! We'll be sure to remove this.

PR #1375, opened 2023-07-24
Fixed mb_convert_encoding() function deprecated issue in php8.2

mohitalgolia:
**Summary**

**Result**

### DIFF
--- a/Helper/Entity/PageHelper.php
+++ b/Helper/Entity/PageHelper.php
@@ -196,7 +196,14 @@ class PageHelper
         if ($completeRemoveTags && $completeRemoveTags !== [] && $s) {
             $dom = new \DOMDocument();
             libxml_use_internal_errors(true);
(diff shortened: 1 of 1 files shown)


Helper/Entity/PageHelper.php
```diff
@@ -196,7 +196,14 @@ private function strip($s, $completeRemoveTags = [])
         if ($completeRemoveTags && $completeRemoveTags !== [] && $s) {
             $dom = new \DOMDocument();
             libxml_use_internal_errors(true);
-            $dom->loadHTML(mb_convert_encoding($s, 'HTML-ENTITIES', 'UTF-8'));
+            $encodedStr = mb_encode_numericentity(
```
cammonro, line 199: This seems like it should work just fine - but I am wondering why we are encoding HTML entities and then decoding? 

Could this be more simply written as the following?

```
$encodedStr = mb_encode_numericentity($s, [0x80, 0x10fffff, 0, ~0]);
```

Unless of course a purpose is served by the back and forth conversion that is not obvious here.

PR #1402, opened 2023-08-29
Fix issue with promotion class being added to all products in autocom…

GaryHaddon:
**Summary**

Fix promotion class being added to all products in autocomplete dropdown.

**Result**

Search for a product using search autocomplete. The price for the product item in the auto complete dropdown will have the "promotion" class on it regardless of promotion state.

After fix only products on promotion will have the "promotion" class.


rachel-trott:
@GaryHaddon Thanks for submitting this. We will review it.

PR #1424, opened 2023-10-11
Update Merchandising.php

cweiss10:
update references to support@algolia.com to new ticket link

**Summary**
remove references to support@aloglia.com and update them to new ticket link

We are phasing out the support@algolia.com email and instead using a ticket form.
**Result**


rachel-trott:
Thanks @cweiss10 Tracked by MAGE-771

PR #1425, opened 2023-10-13
Fix/integ 707 improved category anchor

cammonro:
It was observed that manually anchoring the `hiearchicalMenu` widget via `transformItems` fails when the number of categories exceeds the [limit defined](https://www.algolia.com/doc/api-reference/widgets/hierarchical-menu/js/#widget-param-limit). 

The result is an inexplicable category menu that does not appear to match the data on browse (in this case the category menu should not appear at all because "Jackets" is at the deepest level):
image

A simpler approach is available via the core InstantSearch API - which has been tested to work with `algoliaConfig.request.path` for both browse and search operations.


cammonro:
Moving to new ticket under this PR: https://github.com/algolia/algoliasearch-magento-2/pull/1426

PR #1428, opened 2023-10-19
Fix issue with promotion class being added to all products in autocom…

mohitalgolia:
…plete

**Summary**

**Result**


GaryHaddon:
Interesting why this was recreated when it seems to be the same as this?
https://github.com/algolia/algoliasearch-magento-2/pull/1402/commits/037c2af42867f24b42a35aa163d5b4355ef99a97

PR #1432, opened 2023-10-31
Feature/mage 731

cammonro:
Use of the registry is deprecated. This is a drop-in replacement that provides the added benefit of exposing `\Magento\Catalog\Api\Data\CategoryInterface` which allows plugins on the repository interface rather than direct access to underlying `\Magento\Catalog\Model\Category`.

We should do the same for products following the similar pattern on event `catalog_controller_product_init_after`.


cammonro:
Typo in ticket number. Will resubmit.

PR #1491, opened 2024-02-22
[BUG] Fix for syntax error in construct

denisveg:
**Summary**
The comma on the last item in the constructor is not allowed prior to PHP 8.0 and it's causing syntax issues. Check this for more information: https://php.watch/versions/8.0/trailing-comma-parameter-use-list

CleanShot 2024-02-22 at 12 45 54

**Result**
This PR fixes the syntax error in the construct and make it backward-compatible.


denisveg:
cc @rachel-trott

cammonro:
Thank you for this @denisveg ! We do not officially support PHP 7 as [it is EOL](https://www.php.net/supported-versions.php) but this is a reasonable ask. We'll be sure to merge into an upcoming release.

PR #1493, opened 2024-03-05
Fix trailing comma

frqnck:
**Summary**

```
syntax error, unexpected ')', expecting variable (T_VARIABLE)
```


fiko:
it is important for PHP 7 compatibility, please to take as priority one

cammonro:
Thank you @frqnck! I don't like these trailing commas either even if allowed in PHP 8. 
@fiko we will merge this but please keep in mind that Algolia will not continue supporting PHP 7 in future releases as it is EOL. 
image

cammonro:
Moving to release 3.13.3

PR #1495, opened 2024-03-25
fix recommended blocks feature flags

piotrkwiecinski:
**Summary**

Properly handle feature toggle in recommendations. Each value can be 0 or 1 which is not handled by isset.

**Result**

Recommendation blocks are correctly hidden when a flag is set to No.


cammonro:
Thank you @piotrkwiecinski for identifying this issue!

cammonro:
Moving to release 3.13.3.

PR #1499, opened 2024-04-09
Generic CSS Selectors Need to be more specific

PromInc:
**Summary**

CSS selectors in `recommend.css` are too generic and can cause negative effects to other uses of the CSS classes on the page that are not related to Algolia.

**Result**

Based on the DOM structure in `vendor/algolia/algoliasearch-magento-2/view/frontend/web/internals/template/recommend/products.js` this CSS will create the same result for the Algolia implementation but will not affect other uses across a Magento site.

In my case, I was using the MageDellight SubscribeNow Pro module.  When viewing the customers subscriptions in their account this CSS was breaking the layout of the pages.

### DIFF
--- a/view/frontend/web/internals/recommend.css
+++ b/view/frontend/web/internals/recommend.css
@@ -1,4 +1,4 @@
-.product-img {
+.recommend-item .product-img {
     width: 180px;
 }
 .auc-Recommend-list {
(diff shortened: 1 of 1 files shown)


cammonro:
Thank you @PromInc ! Good call out. We'll be sure to include this helpful change in a future release.

PR #1507, opened 2024-05-07
Feature/mage 860

mrahman3177:


etc/csp_whitelist.xml
```diff
@@ -12,7 +12,7 @@
                 <value id="algolia-api" type="host">*.algolia.net</value>
                 <value id="algolia" type="host">*.algolia.com</value>
                 <value id="algolia-fallback" type="host">*.algolianet.com</value>
-                <value id="algolia-insights-io" type="host">*.insights.algolia.io</value>
+                <value id="algolia-insights-io" type="host">insights.algolia.io</value>
```
cammonro, line 15: @mrahman3177 I believe opening connections on the subdomain is still a possible scenario - so, to be safe, I'd prefer it if we account for both possibilities with two rules (since this cannot be captured in one rule):
```suggestion
                <value id="algolia-insights-io" type="host">*.insights.algolia.io</value>
                <value id="algolia-insights-io-bare" type="host">insights.algolia.io</value>
```

cammonro (changes requested):

cammonro (approved):

PR #1513, opened 2024-05-15
MAGE-879: decompoundedAttributes removed from merging settings

mrahman3177:
Based on my limited knowledge, decompoundedAttributes setting is specific to Algolia dashboard. It should not be part of the merged settings. So I added in the remove list.

Now config saves without any error
Screenshot 2024-05-15 at 12 59 41 PM

### DIFF
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -289,7 +289,7 @@ class AlgoliaHelper extends AbstractHelper
         } catch (\Exception $e) {
         }
 
-        $removes = ['slaves', 'replicas'];
+        $removes = ['slaves', 'replicas', 'decompoundedAttributes'];
 
         if (isset($settings['attributesToIndex'])) {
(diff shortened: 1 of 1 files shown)


Helper/AlgoliaHelper.php
```diff
@@ -289,7 +289,7 @@ public function mergeSettings($indexName, $settings, $mergeSettingsFrom = '')
         } catch (\Exception $e) {
         }
 
-        $removes = ['slaves', 'replicas'];
+        $removes = ['slaves', 'replicas', 'decompoundedAttributes'];
```
cammonro, line 292: @mrahman3177 As this issue has been resolved with PHP API client v4, let's apply this fix to v3.13 only. (Do not merge forward to 3.14)

cammonro (approved):
Looks good but as noted - let's only apply to patch release on v`3.13`. v`3.14` will use the new PHP connector which resolves this issue and precludes the need to explicitly filter the `decompoundedAttributes` setting.

PR #1514, opened 2024-05-20
MAGE-2 Class CSS name conflict from other modules fix

sgeleon:
Some name of CSS class have very common name that can conflict with other class from other modules. I think we should be using more specific selector for it.

```
#{some id} .recommend-component {
    margin-bottom: 80px;
}
```


cammonro:
Agree! We will address in an upcoming release.

PR #1515, opened 2024-05-21
Feature/mage 866

mrahman3177:


Helper/AnalyticsHelper.php
```diff
@@ -58,25 +49,17 @@ class AnalyticsHelper
...
     public function __construct(
-        AlgoliaHelper $algoliaHelper,
-        ConfigHelper $configHelper,
-        IndexEntityDataProvider $entityHelper,
-        Logger $logger,
-        string $region = 'us'
+        private ConfigHelper            $configHelper,
+        private IndexEntityDataProvider $entityHelper,
+        private Logger                  $logger,
+        private ResolverInterface       $localeResolver
```
cammonro, line 61: @mrahman3177 Rather than do a straight cherry pick here, can we sift out the property promotion (stick with old style constructor code) to avoid compatibility issues with PHP 7?

ViewModel/Adminhtml/Analytics/Overview.php
```diff
@@ -17,34 +18,21 @@ class Overview implements \Magento\Framework\View\Element\Block\ArgumentInterfac
...
-    ) {
-        $this->backendView = $backendView;
-        $this->analyticsHelper = $analyticsHelper;
-        $this->indexEntityDataProvider = $indexEntityDataProvider;
-    }
+        protected BackendView             $backendView,
+        protected AnalyticsHelper         $analyticsHelper,
+        protected IndexEntityDataProvider $indexEntityDataProvider,
+        protected ResolverInterface       $localeResolver
+    ) { }
```
cammonro, line 35: As above let's keep property promotion in feature release.

composer.json
```diff
@@ -5,6 +5,7 @@
   "license": ["MIT"],
   "version": "3.13.2",
   "require": {
+    "php": "~8.1|~8.2",
```
cammonro, line 8: Even though we don't officially support PHP 7, let's wait till 3.14 before we explicitly call out PHP 8.

cammonro (changes requested):
Let's keep the official move to PHP 8 to upcoming version 3.14.

cammonro (approved):

PR #1524, opened 2024-05-31
BUG: Client side encoding for front end XSS protection on InstantSearch search box

cammonro:
**Summary**

Errant decoding allows execution of arbitrary JS directly in the InstantSearch search box. This PR aims to refactor client side encoding / decoding in the `algolia` object and apply to the query before rendering on FE.

**Result**

Disallowed characters are escaped to HTML entities: 

image


damcou (approved):
👍 Works fine :)

mrahman3177 (approved):
Nice!

PR #1528, opened 2024-06-05
MAGE-870 Integration tests

damcou:
- Removed old classes to test against Magento 2.1 and Magento 2.2 values
- Fixed some paths in `ConfigPatch.php`
- Fixed values to test in the Magento 2.4 class
- Changed the `TestCase.php` file to test only the two supported Magento versions
- Updated the `install-config-mysql.php` file.
- Updated the `QueueTest.php` to make most of the tests work (the other ones are still generating locks, that's why they are commented, still haven't found a good solution for it)

➡️ I think this current state is a good start to reintegrate the tests on the repo once we find how to use new circleci images.


Setup/Patch/Schema/ConfigPatch.php
```diff
@@ -81,7 +81,7 @@ class ConfigPatch implements SchemaPatchInterface
 
         'algoliasearch_synonyms/synonyms_group/enable_synonyms' => '0',
 
-        'algoliasearch_advanced/advanced/number_of_element_by_page' => '300',
+        'algoliasearch_advanced/queue/number_of_element_by_page' => '300',
```
cammonro, line 84: Good catch!

Test/Integration/QueueTest.php
```diff
@@ -794,12 +797,9 @@ public function testHugeJob()
...
             (2, NULL, \'class\', \'rebuildStoreProductIndex\', \'{"store_id":"2","product_ids":["9","22"]}\', 3, 0, \'\', 2);');
 
-        /** @var Queue $queue */
-        $queue = $this->getObjectManager()->create(Queue::class);
-
-        $pid = getmypid();
+//        $pid = getmypid();
         /** @var Job[] $jobs */
-        $jobs = $this->invokeMethod($queue, 'getJobs', ['maxJobs' => 10]);
+        $jobs = $this->invokeMethod($this->queue, 'getJobs', ['maxJobs' => 10, 'doLock' => false]);
```
cammonro, line 802: Good workaround as this feature will likely go away in the future.

cammonro (approved):
@damcou These changes look good to me but you can please also rebase and test against `release/3.14.0-beta`? There you can see I was able to avoid the the transaction issue with with a `@magentoDbIsolation` annotation but that led to other issues. Not sure if that's worth a look?

I'm also OK with these updates going into the patch release since ConfigPatch has old paths that should be corrected.

damcou:
@cammonro : I gave a try to the @magentoDbIsolation annotation and indeed, it seems to solves some issues, so I could remove the commented code and added a few skipped tests as you did on the 3.14 branch. 
I'll make the changes in the 3.14 branch tomorrow so we can close this topic :)

PR #1531, opened 2024-06-12
MAGE-721 (including MAGE-835 MAGE-836 MAGE-840) Refactor for granular virtual replicas

cammonro:
Initial refactor for granular virtual replicas including:

- Introduced new admin groups to InstantSearch for improved UX
- Updated `ConfigHelper` to use new paths
- Added data patch to migrate old configurations
- Bugfix for query rule disable on facets with new admin groupings
- Added new sorting admin option via source model
- Added derived virtual replica enablement to `ConfigHelper` based on `ArraySerialized`
- Intro’d simplified data structures to avoid array diff mismatches
- Intro’d new `ReplicaManager` abstraction to map Magento sorting to Algolia replica configuration
- Removed dependencies in backend models to handle replica config updates in Algolia addressing stale data
- Added `ReplicaState` registry for tracking changes to sorting configuration to minimize number of replica build operations
- Added logic to preserve replicas created outside of Magento such as Merchandising Studio "sorting strategies"
- Introduced PHP 8 constructor property promotion on affected classes
- Added stronger typing to affected classes and methods

> [!IMPORTANT]
> This is a work in progress and completion will depend on related tasks in the epic.


Helper/ConfigHelper.php
```diff
@@ -1024,85 +1124,53 @@ public function getSortingIndices($originalIndexName, $storeId = null, $currentC
...
-                ];
+                $attrs[$key]['ranking'] = $this->getSortAttributingRankingSetting($sortAttribute, $attr['sort']);
+                $attrs[$key] = $this->decorateSortAttribute($attrs[$key]);
             }
         }
         $attrsToReturn = [];
-        if (count($attributesToAdd) > 0) {
-            foreach ($attrs as $key => $attr) {
-                if ($attr['attribute'] == 'price' && isset($attributesToAdd[$attr['sort']])) {
+        if (count($attributesToAdd)) {
```
damcou, line 1159: I gave it a try, and my replicas are not created anymore. I tried to have a look at this method and it seems that `$attrsToReturn` is only filled when `$attributesToAdd` is not empty (which seems to be the case if the customer groups are not enabled, which is the case for me and `getSortingIndices` is returning an empty array no matter the sorting config in the admin) . Maybe I misunderstood something in the process ?

cammonro: Good catch @damcou ! That issue is from legacy code that probably wasn't obvious before because of the multiple ways replicas were getting written back to Algolia so the needed were _evenutally_ created. My goal is to run it through one funnel and I've been mainly testing with customer groups so this is a helpful find.

I pushed a fix for this issue but it doesn't handle all scenarios. There is more work to do with regards to toggling virtual vs standard on an existing index along with customer group pricing enablement. There are separate tickets for these items which I will follow up in a future PR.

cammonro:
As this is a work in progress - rebasing against epic branch to avoid destabilizing replica functionality on current `release/3.14.0-beta`.

PR #1534, opened 2024-06-13
MAGE-901: showing title for empty recommendation issue addressed

mrahman3177:
With this code change, when there is no recommendation, only title and empty placeholder will not show anymore

### DIFF
--- a/view/frontend/web/recommend.js
+++ b/view/frontend/web/recommend.js
@@ -42,9 +42,9 @@ define([
                                 position: index + 1,
                             }));
                         },
-                        headerComponent({html}) {
-                            if (this.props.recommendations.length === 0) {
-                                return html
+                        headerComponent({html, recommendations}) {
+                            if (!recommendations.length) {
+                                return '';
                             }
                             return recommendProductsHtml.getHeaderHtml(
                                 html,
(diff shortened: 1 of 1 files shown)


view/frontend/web/recommend.js
```diff
@@ -43,6 +43,9 @@ define([
                             }));
                         },
                         headerComponent({html}) {
+                            if (this.props.recommendations.length === 0) {
```
cammonro, line 46: I don't think we should be referencing `this` in the callback as it exposes the underlying implementation. Instead we should destructure the passed arguments to the callback and reference it that way.

view/frontend/web/recommend.js
```diff
@@ -43,6 +43,9 @@ define([
                             }));
                         },
                         headerComponent({html}) {
```
cammonro, line 45:
```suggestion
                        headerComponent({html, recommendations}) {
                            if (!recommendations.length) {
                                return '';
                            }
                        
```

cammonro (changes requested):
@mrahman3177 Please see my comments - also I'm wondering if we can update to use the `templates` [API](https://www.algolia.com/doc/api-reference/widgets/trending-items/js/#templates) instead of the [React style functions](https://www.algolia.com/doc/api-reference/widgets/trending-items/react/#widget-param-headercomponent)?

Lastly I see there is `empty()` template support - can that be used here in any way?

cammonro (approved):

PR #1537, opened 2024-06-15
Update ProductHelper.php to prevent false in facets

matthewbond:
We had a report from a customer that their facets show false as a value because there is no check if the product attribute  value is false before passing it on to be potentially added to the attributes of a product. Adding their suggestion which should resolve the issue.

**Summary**

This change modifies the check to make sure that the attribute value is not false before passing it to the $customerData object

**Result**

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -1237,7 +1237,7 @@ class ProductHelper
             $value = $attributeResource->getFrontend()->getValue($product);
         }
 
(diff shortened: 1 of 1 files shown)


matthewbond:
Thanks to https://github.com/markfischmann for suggesting this change!

damcou:
Hi there, we saw that in some cases, users may want to still get a `false` value for this. So this change should stay as a custom change required for an edge case. Closing this PR as we don't want to make this behaviour as the default one.

PR #1539, opened 2024-06-15
The index in Algolia after full reindex has less records than expected

pikulsky:
**Summary**

If a queue has jobs from a full reindex, then some realtime jobs, and then another full reindex jobs, at the end the resulted index has less records.

**Details**

The issue is related to the way how jobs are loaded from the queue in \Algolia\AlgoliaSearch\Model\Queue::getJobs() method.

The jobs loaded from the queue to be processed by QueueRunner indexer consist of three parts:
- "full reindex" jobs - 33% of "max number of jobs per run"
- "realtime" jobs (from 33% up to 100%)
- rest are "full reindex" jobs (up to 100%).

The current logic of loading jobs from the queue is the following:
- step 1. loading 30% of "max number of jobs per run" of "full reindex" jobs (those with "is_full_reindex" = 1)
- step 2. loading "realtime" jobs (those with "is_full_reindex" = 0) - up to "max number of jobs per run"
- step 3. get the max job ID of these jobs above ("max ID of full reindex and realtime jobs", $lastFullReindexJobId in the code)
- step 4. if the number of jobs still less than "max number of jobs per run", then load "full reindex" jobs with IDs AFTER "max ID of full reindex and realtime jobs"

**Steps to reproduce the bug**
1. Full reindex (#<span></span>1) is performed, so the corresponding "full reindex" jobs are added to the queue.
2. Then some entities are updated, so some "realtime" jobs are added to the queue.
3. Another full reindex (#<span></span>2) is performed, so the corresponding "full reindex" jobs are added to the queue.

In this case $lastFullReindexJobId calculated in step 3 will point to "full reindex" jobs from reindex #<span></span>2.
So, jobs added for full reindex #<span></span>2 will be used in reindex #<span></span>1, processed and deleted from the queue.
Full reindex #<span></span>2 will not include these jobs any more.

When full reindex #<span></span>2 finishes, the index in Algolia will miss these records.
With default page number=300 it is missing 300 records for each job added initially for reindex #<span></span>2, but processed and deleted while processing jobs of reindex #<span></span>1.

See the attached screenshot with loaded jobs: "pid" field is filled for loaded jobs:
- from full reindex #<span></span>1, 
- realtime job,  
- jobs from full reindex #<span></span>2.
 

algolia

**Result**

With the fix applied the jobs from full reindex #<span></span>1 will be processed without loading jobs from full reindex #<span></span>2.
The index after full reindex #<span></span>2 will have all records.

### DIFF
--- a/Model/Queue.php
+++ b/Model/Queue.php
@@ -392,8 +392,6 @@ class Queue
             $fullReindexJobs = $this->fetchJobs($fullReindexJobsLimit, true);
             $fullReindexJobsCount = count($fullReindexJobs);
 
-            $lastFullReindexJobId = max($this->getJobsIdsFromMergedJobs($fullReindexJobs));
-
             $realtimeJobsLimit = (int) $maxJobs - $fullReindexJobsCount;
 
             $realtimeJobs = $this->fetchJobs($realtimeJobsLimit, false);
@@ -403,6 +401,7 @@ class Queue
 
             if ($jobsCount > 0 && $jobsCount < $maxJobs) {
                 $restLimit = $maxJobs - $jobsCount;
+                $lastFullReindexJobId = max($this->getJobsIdsFromMergedJobs($fullReindexJobs));
 
                 $restFullReindexJobs = $this->fetchJobs($restLimit, true, $lastFullReindexJobId);
 
--- a/Model/Queue.php
+++ b/Model/Queue.php
@@ -401,7 +401,12 @@ class Queue
 
             if ($jobsCount > 0 && $jobsCount < $maxJobs) {
                 $restLimit = $maxJobs - $jobsCount;
-                $lastFullReindexJobId = max($this->getJobsIdsFromMergedJobs($fullReindexJobs));
+
+                if (count($fullReindexJobs)) {
+                    $lastFullReindexJobId = max($this->getJobsIdsFromMergedJobs($fullReindexJobs));
+                } else {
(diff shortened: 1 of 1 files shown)


cammonro:
@pikulsky This is a fantastic contribution. Thank you for your effort here and for providing supporting details for what looks to be a long standing bug. 💪 

In general if you are seeing lots of full reindexes in the queue it can become pretty inefficient so it may be better to perform those on a schedule outside of the queue BUT regardless this kind of job droppage like you've discovered should not be happening. 

We'll be happy to include your fix in a patch release!

pikulsky:
@cammonro , thank you! And, I totally agree with your note about multiple full re-indexes: actually, it was our mistake to allow it happen without a need.

pikulsky:
@cammonro , there is error in our logs related to the patch (php 8.1):

> ValueError: max(): Argument #1 ($value) must contain at least one element 

so, I will check for the fix and update PR, something like:
```
if (count($fullReindexJobs)) {
    $lastFullReindexJobId = max($this->getJobsIdsFromMergedJobs($fullReindexJobs));
} else {
    $lastFullReindexJobId = max($this->getJobsIdsFromMergedJobs($jobs));
}
```

damcou:
Hi @pikulsky , thanks a lot for this, especially for having perfectly explained how to reproduce the bug. This will be part of the next release ! 💪

PR #1541, opened 2024-06-18
mage-851: Fix inline scripts for 3.13.4

damcou:
Same as https://github.com/algolia/algoliasearch-magento-2/pull/1535 for 3.13.4.

### DIFF
--- a/view/frontend/templates/internals/configuration.phtml
+++ b/view/frontend/templates/internals/configuration.phtml
@@ -1,20 +1,18 @@
 <?php
 
 /** @var \Algolia\AlgoliaSearch\Block\Configuration $block */
+/** @var \Magento\Framework\View\Helper\SecureHtmlRenderer $secureRenderer */
 
 $configuration = $block->getConfiguration();
 
 ?>
-<script>
-    <?php
-    if ($configuration['instant']['enabled'] === true && $configuration['isSearchPage'] === true) :
-        $css = '<style type="text/css">' . $configuration['instant']['selector'] . ' {display:none}</style>';
-    ?>
-	// Hide the instant-search selector ASAP to remove flickering. Will be re-displayed later with JS.
-	document.write('<?php /* @noEscape */ echo $css; ?>');
-    <?php
-    endif;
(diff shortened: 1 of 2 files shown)


hostep:
@damcou, just a small remark, but you guys have this in your `composer.json`:
```
    "magento/framework": "~102.0|~103.0",
```

The SecureHtmlRenderer class was introduced in version 103.0.0, so it's not available in 102.0.x and the code introduced here, will in theory crash with those versions (so all Magento versions lower than 2.4.0).

So you probably want to drop the `~102.0` constraint?

damcou:
Good catch @hostep  :) , 
We will discuss it with the team ( cc @cammonro , @mrahman3177 for visibility)

hostep:
Some additional info in case it helps with decision making, if you want to keep compatibility with 102.0 you can first check if SecureHtmlRenderer class exists or not and if it doesn't exist just output the script as-is.

Also good to know, but the heredoc syntax used in the `view/frontend/templates/layer/view.phtml` file will cause problems on Magento 2.3.x shops when html minification is enabled. Magento fixed that in version 2.4.0 with https://github.com/magento/magento2/commit/41c3f6f4338326ee6efcf04a96555de9a0c3c436, so if you want to keep compatibility with Magento 2.3.x, I would strongly recommend to remove the heredoc syntax.

All these things were discovered in https://github.com/yireo/Yireo_GoogleTagManager2/pull/231 BTW.

cammonro:
@hostep Thank you for calling this out. Very helpful! 

This change was pushed to comply with newer 2.4.7 CSP requirements and to avoid needing to allow `unsafe-inline` for our module to function. But you make some good suggestions (both here and on Yireo) and we will consider them. 

In answer to your question we have no intention of continuing support for Magento 2.3 as we are shifting our future code to PHP 8. We will be sure to remove `~102.0` as a dependency in an upcoming release.

abdulrafaywaqar:
I am installing the latest version in Magento 2.4.2
The compilation stuck on 11%

PR #1542, opened 2024-06-19
Feature/mage 839 Store scoped replica state management

cammonro:
This PR introduces store scoped state management for replicas allowing highly selective changes to be persisted to Algolia minimizing the number of build operations on each store's respective index. 

All replica configuration persistence to Algolia is funneled through the new `ReplicaManager` class while backend models are used to track exactly what was changed for which affected stores via the `ReplicaState` singleton and the `ConfigChecker` utility class. 

Currently the `SaveSettings` observer iterates over all stores by default but replicas will not persist unless a change actually occurs for a given store. Changes to the default scope should not impact overridden websites or stores. Changes to a website should be localized to the associated store indices only (unless explicitly overridden by individual store configurations). A future refactor could make the `SaveSettings` observer more precise but the `ReplicaManager` handles its state management independently regardless.

If the `IndicesConfigurator` is triggered independently of `ReplicaState` management (e.g. via the indexing queue etc.) things will fall back to an "unknown" state which then uses direct comparison against the Algolia configuration to determine whether a change needs to be applied to the index. Hopefully this is flexible enough for a variety of scenarios.

### DIFF
--- a/Model/Product/ReplicaManager.php
+++ b/Model/Product/ReplicaManager.php
@@ -77,13 +77,25 @@ class ReplicaManager implements ReplicaManagerInterface
         }
     }
 
-    protected function getReplicaConfigurationFromAlgolia($primaryIndexName, bool $refreshCache = false)
+    /**
+     * @param $primaryIndexName
+     * @param bool $refreshCache
+     * @return array<string, mixed>
+     * @throws LocalizedException
+     */
+    protected function getReplicaConfigurationFromAlgolia($primaryIndexName, bool $refreshCache = false): array
     {
         if ($refreshCache || !isset($this->_algoliaReplicaConfig[$primaryIndexName])) {
-            $currentSettings = $this->algoliaHelper->getSettings($primaryIndexName);
-            $this->_algoliaReplicaConfig[$primaryIndexName] = array_key_exists('replicas', $currentSettings)
-                ? $currentSettings['replicas']
-                : [];
+            try {
+                $currentSettings = $this->algoliaHelper->getSettings($primaryIndexName);
+                $this->_algoliaReplicaConfig[$primaryIndexName] = array_key_exists('replicas', $currentSettings)
+                    ? $currentSettings['replicas']
+                    : [];
+            } catch (\Exception $e) {
+                $msg = "Unable to retrieve replica settings for $primaryIndexName: " . $e->getMessage();
+                $this->logger->error($msg);
+                throw new LocalizedException(__($msg));
+            }
         }
         return $this->_algoliaReplicaConfig[$primaryIndexName];
     }
@@ -103,6 +115,7 @@ class ReplicaManager implements ReplicaManagerInterface
      *
      * @param string $primaryIndexName
      * @return string[]
+     * @throws LocalizedException
      */
     protected function getMagentoReplicaConfigurationFromAlgolia(string $primaryIndexName): array
     {
@@ -130,6 +143,7 @@ class ReplicaManager implements ReplicaManagerInterface
     /**
      * @param string $primaryIndexName
      * @return array
+     * @throws LocalizedException
      */
     protected function getNonMagentoReplicaConfigurationFromAlgolia(string $primaryIndexName): array
     {
--- a/Model/Product/ReplicaManager.php
+++ b/Model/Product/ReplicaManager.php
@@ -100,13 +100,12 @@ class ReplicaManager implements ReplicaManagerInterface
      * relevant to the Magento integration
      *
      * @param string $primaryIndexName
-     * @param int $storeId
      * @return string[]
      */
     protected function getMagentoReplicaConfigurationFromAlgolia(string $primaryIndexName): array
     {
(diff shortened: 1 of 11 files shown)


Registry/ReplicaState.php
```diff
@@ -4,32 +4,58 @@
...
+                : self::REPLICA_STATE_CHANGED;
+        }
+        return $state;
     }
 
-    public function isStateChanged(): bool
+    public function setChangeState(int $state, int $storeId): void
     {
-        return $this->updatedSortConfiguration !== $this->originalSortConfiguration;
+        if ($state < self::REPLICA_STATE_UNCHANGED || $state > self::REPLICA_STATE_UNKNOWN) {
```
damcou, line 55: Is there a scenario where it's needed ?

Model/Product/ReplicaManager.php
```diff
@@ -169,55 +194,79 @@ protected function getPossibleMagentoReplicaSettings(string $indexName, int $sto
...
     {
         $sortingIndices = $this->configHelper->getSortingIndices($indexName, $storeId);
-        $newMagentoReplicas = $this->transformSortingIndicesToReplicaSetting($sortingIndices);
-        $oldMagentoReplicas = $this->getMagentoReplicaConfigurationFromAlgolia($indexName, $storeId);
-        $nonMagentoReplicas = $this->getNonMagentoReplicaConfigurationFromAlgolia($indexName, $storeId);
-        $oldMagentoReplicaIndices = $this->getBareIndexNamesFromReplicaSetting($oldMagentoReplicas);
-        $newMagentoReplicaIndices = $this->getBareIndexNamesFromReplicaSetting($newMagentoReplicas);
+        $newMagentoReplicasSetting = $this->transformSortingIndicesToReplicaSetting($sortingIndices);
+        $oldMagentoReplicasSetting = $this->getMagentoReplicaConfigurationFromAlgolia($indexName, $storeId);
+        $nonMagentoReplicasSetting = $this->getNonMagentoReplicaConfigurationFromAlgolia($indexName, $storeId);
```
damcou, line 220: It's not clear to me why we need to fetch replicas not belonging to Magento ? Shouldn't we assume that Magento is the only source of truth when it comes to apply settings to its own indices ?

Model/Product/ReplicaManager.php
```diff
@@ -143,21 +169,20 @@ function($sort) use ($mode) {
...
-     * @param string $indexName
+     * @param string $primaryIndexName
      * @param int $storeId
      * @param bool $refreshCache
      * @return array
      * @throws LocalizedException
      * @throws NoSuchEntityException
      */
-    protected function getPossibleMagentoReplicaSettings(string $indexName, int $storeId, bool $refreshCache = false): array
+    protected function getPossibleMagentoReplicaSettingsFromConfig(string $primaryIndexName, int $storeId, bool $refreshCache = false): array
```
damcou, line 179: I'm not a huge fan of the "possible" wording, it brings some uncertainty whereas this should be clear and absolute, if the index name has the right prefix, it belongs to Magento, and we shouldn't care about it if it doesn't. 
(But maybe I don't have all the keys to understand here)

damcou (approved):
Some questions more than change requests, but overall, this works fine ! 🙌

Model/Product/ReplicaManager.php
cammonro: Great question. Unfortunately that is not the case anymore as tools like Merchandising Studio can create sorting strategies independent of Magento that the integration could potentially blow away destroying merchandising work in the process.

Registry/ReplicaState.php
cammonro: Honestly, it should never happen! But "never say never" they say. Probably me just being paranoid in case someone else uses the class and passes an unexpected value. If this was TypeScript I could restrict it with typing. We could possibly try using an enum here. I know some people hate enums but I believe you can use them in PHP 8.1.

Model/Product/ReplicaManager.php
cammonro: Good call out. You know I agree with you.I mean it is conceivable that someone could create a replica directly in Algolia (outside of Magento) that uses the exact same prefixing approach that the integration does (which would fool the Magento extension into thinking it was its own managed replica) but that would be a very silly thing to do. I don't think we need to worry about that so I will rename this.

Model/Product/ReplicaManager.php
damcou: Oh I see, and those MS indices have the exact same prefix I guess ?

cammonro: I'm not sure if it's documented but from my testing I've observed that the MS sorting strategies add their own prefix like `cat_sorting_` + the primary index + attribute name. e.g.

`cat_sorting_magento2_dev_default_products_activity` 

As long as this convention is preserved we can rely on our own prefix match. Some further examples:



The last two indices above are not managed by Magento. The first is an arbitrary replica and the second is created by an MS "sorting strategy". The new approach will preserve both of these.

PR #1545, opened 2024-06-21
Feature/mage 837 - virtual replica validator

cammonro:
This PR aims to do the following:
- Introduce validation across replica operations (including admin and indexing) to ensure that granular sorting configurations do not create an undesirable virtual replica scenario.
- Clearly communicate to the end user what the cause of the replica error was, including calling out scenarios where customer group pricing is preventing the activation of virtual replicas.
- Automatically roll back any configuration changes made in the admin that will result in an invalid configuration and lead to an Algolia / Magento configuration mismatch.
- Link both customer group pricing enablement and sorting in the admin to the same underlying replica state so both can be validated independently.
- Maintain awareness of _applied_ scope in Magento and _affected_ stores in Algolia.
- Communicate which store is responsible for any encountered replica errors in case it must be remedied outside of the current scope. 

Standard replica limit exceeded error:
image

Customer group limit exceeded error:
image

Enablement of customer group pricing error:
image

Indexing error:
image

### DIFF
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -443,6 +443,10 @@ class ConfigHelper
         return [];
     }
 
+    protected function serialize(array $value): string {
+        return $this->serializer->serialize($value);
+    }
+
     /**
      * @param $value
      * @return array|bool|float|int|mixed|string|null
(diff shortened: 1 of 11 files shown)


Helper/ConfigHelper.php
```diff
@@ -233,32 +168,21 @@ class ConfigHelper
...
+        protected DirectoryList                                        $directoryList,
+        protected Magento\Framework\Module\ResourceInterface           $moduleResource,
+        protected Magento\Framework\App\ProductMetadataInterface       $productMetadata,
+        protected Magento\Framework\Event\ManagerInterface             $eventManager,
+        protected SerializerInterface                                  $serializer,
+        protected GroupCollection                                      $groupCollection,
+        protected GroupExcludedWebsiteRepositoryInterface              $groupExcludedWebsiteRepository,
+        protected CookieHelper                                         $cookieHelper
+    )
+    {}
```
damcou, line 185: I'm just wondering, is it a whole new way to create constructors in Magento introduced in the last past 3 years ? It's so cool not to have to do it the "old way" 😃

cammonro: @damcou Agree! [Constructor property promotion](https://php.watch/versions/8.0/constructor-property-promotion) is one of my favorite new features in PHP 8 and cleans up this code a lot.

PR #1553, opened 2024-06-28
MAGE-947: restrict recommend CSS to algolia managed blocks

damcou:
This PR ports the change suggested by @sgeleon here: https://github.com/algolia/algoliasearch-magento-2/pull/1514

- Added a #algoliaRecommend block which acts as a parent for each recommend block
- Applied CSS on the recommend blocks, taking into account the newly created parent block

### DIFF
--- a/view/frontend/web/internals/recommend.css
+++ b/view/frontend/web/internals/recommend.css
@@ -11,7 +11,7 @@
     width: 110px;
     margin: 0 auto;
 }
-.recommend-component {
+#algoliaRecommend .recommend-component {
     margin-bottom: 80px;
 }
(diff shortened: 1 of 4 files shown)


view/frontend/web/internals/recommend.css
```diff
@@ -11,7 +11,7 @@
     width: 110px;
     margin: 0 auto;
 }
-.recommend-component {
+#algoliaRecommend .recommend-component {
```
cammonro, line 14: 👍 

To be absolutely safe, I think we should apply this specificity to _all_ Algolia Recommend related DOM elements.

cammonro (changes requested):

cammonro (approved):

mrahman3177 (approved):
Looks good. Good Job!

PR #1555, opened 2024-07-01
Feature/mage 941 - Refactor / intro service classes

cammonro:
Our "helper" classes have been getting kind of out of control. I use quotes because in many ways they represent services or models rather than straight up utility methods and some have turned into basically "god classes" where division of responsibility is unclear and this makes iterative or simultaneous improvement difficult. 

Completely fixing this is a much larger effort so this refactor seeks to take a step in the direction of adding more composability:

- Separating objects with more distinct roles related to upcoming or evolving features 
- Distinguishing models and helpers from services starting with 4 services in the new `Service` namespace
- Centralizing logic where possible, e.g. using traits or abstract classes where it (hopefully) makes sense to apply DRY principles
- Allow entity helpers to be able to supply information about the entities they manage to the caller
- Deprecating and replacing misleading APIs 
- Adding clarity to naming conventions

More specifically this PR:

- Centralizes how index names are generated via the new service `IndexNameFetcher`
- Introduces a shared interface to the service for all entity helpers via `AbstractEntityHelper`
- Reclassifies new 3.14 revenue analytics feature (formerly the `Events` model ) as the service `EventsProcessorInterface` (since it interacts with multiple models i.e Magento and Algolia)
- For the same reason this also reclassifies the new 3.14 granular virtual replica feature `ReplicaManagerInterface` as a service
- Functions related to transformation of the sorting attributes in Magento has been removed from the bloated `ConfigHelper` class and turned into a new `SortingTransformer` service
- And lastly (which was the genesis of this entire effort) the hard (6 year old) requirement for supplying an index name for generating replicas has been removed to abstract the handling of replicas based on store scope alone. This was done to remove the potentially dangerous implication that temp or arbitrary indices could be used with replica management. (It is not possible to move Algolia indices that have attached replicas so a temp index would never be used in this way.)

Two other things were done along the way:

- Certain constructors were simplified to use PHP 8 constructor property promotion and useless PHPDoc comments removed
- An ongoing effort to introduce stronger typing was also applied as part of this effort while working on affected classes/methods

### DIFF
--- a/Api/Insights/EventProcessorInterface.php
+++ b/Api/Insights/EventProcessorInterface.php
@@ -9,7 +9,7 @@ use Magento\Quote\Model\Quote\Item;
 use Magento\Sales\Model\Order;
 use Magento\Store\Model\StoreManagerInterface;
 
-interface EventsProcessorInterface
+interface EventProcessorInterface
 {
     /** @var string */
     public const EVENT_KEY_SUBTYPE = 'eventSubtype';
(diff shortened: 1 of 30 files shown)


damcou (approved):
Huge work on this ! 💯 Tested locally and it seems fully functional 💪 
I like the new index name management, this is definitely more consistent now, and I think we should use more traits as you did.
One small comment, the PR is pretty massive and it's hard to be fully sure that I didn't miss anything. Maybe next time you should avoid to mix different types of refactoring (for example, by putting the constructor revamping or types specification in another PRs ?) 

Great job ! ❤️

cammonro:
@damcou Fair feedback! Sometimes you just kind of get in the zone you know? Will keep that in mind for future. Thank you for reviewing. 🙏

cammonro:
One more small change - a little OCD but renamed `EventsProcessorInterface` to `EventProcessorInterface` for a more consistent nomenclature (e.g. `ReplicaManager`, `IndexNameFetcher` etc.) and because it may process _one_ (or more) events at a time.

PR #1562, opened 2024-07-09
Feature/mage 848 - Granular virtual replica migration

cammonro:
This PR includes:
- Refactored CLI classes for code reuse
- A new `disable-virtual-replicas` command to facilitate upgrades and allow a way to selectively roll back invalid configurations
- Forcing replica state on replica rebuilds to bypass Algolia API latency
- Bug fixes for the `ConfigChecker` class to handle more nested scoping config scenarios 
- A new data patch to migrate legacy global virtual configuration to granular sorting attribute configuration
- A new data patch to run after to fully rebuild the replica configuration 
- Error handling if replica limit is exceeded with detailed instructions on how to remedy (this should hopefully prevent upgrading with unexpected results to existing replica configurations)

Sample CLI output:
image
image

Sample `setup:upgrade` failure:
image

### DIFF
--- a/Console/Command/ReplicaDisableVirtualCommand.php
+++ b/Console/Command/ReplicaDisableVirtualCommand.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Console\Command;
+
+use Algolia\AlgoliaSearch\Api\Console\ReplicaSyncCommandInterface;
+use Algolia\AlgoliaSearch\Console\Traits\ReplicaSyncCommandTrait;
+use Magento\Framework\Console\Cli;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+class ReplicaDisableVirtualCommand extends AbstractReplicaCommand implements ReplicaSyncCommandInterface
+{
+    use ReplicaSyncCommandTrait;
+
+    protected function getReplicaCommandName(): string
+    {
+        return 'disable-virtual-replicas';
+    }
+
+    protected function getCommandDescription(): string
+    {
+        return 'Disable virtual replicas for all product sorting attributes and revert to standard replicas';
+    }
+
+    protected function getStoreArgumentDescription(): string
+    {
+        return 'ID(s) for store(s) to disable virtual replicas (optional), if not specified disable virtual replicas for all stores';
+    }
+
+    protected function getAdditionalDefinition(): array
+    {
+        return [];
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $this->output = $output;
+        $this->input = $input;
+        $this->setAreaCode();
+
+        $storeIds = $this->getStoreIds($input);
+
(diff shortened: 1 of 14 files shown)


Console/Command/ReplicaDisableVirtualCommand.php
```diff
@@ -0,0 +1,178 @@
...
+            ConfigHelper::SORTING_INDICES,
+            ScopeInterface::SCOPE_STORES,
+            $storeId)
+        ) {
+            $isStoreScoped = true;
+            $this->disableVirtualReplicaSortConfig(ScopeInterface::SCOPE_STORES, $storeId);
+        }
+
+        if (!$isStoreScoped) {
+            $this->output->writeln("<info>Virtual replicas are not configured at the store level for $storeName. You will need to re-run this command for all stores.</info>");
```
damcou, line 133: Don't we want to allow to disable virtual replicas on one particular store even if there's no value set at store level ?

Setup/Patch/Data/MigrateVirtualReplicaConfigPatch.php
```diff
@@ -0,0 +1,162 @@
...
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Exception\NoSuchEntityException;
+use Magento\Framework\Serialize\SerializerInterface;
+use Magento\Framework\Setup\ModuleDataSetupInterface;
+use Magento\Framework\Setup\Patch\DataPatchInterface;
+use Magento\Framework\Setup\Patch\PatchInterface;
+use Magento\Store\Model\ScopeInterface;
+use Magento\Store\Model\StoreManagerInterface;
+
+class MigrateVirtualReplicaConfigPatch implements DataPatchInterface
```
damcou, line 26: Is the order of patch application important somehow between the 2? How is it defined ?

damcou (approved):
Looks good 👋 (two little questions though)

Console/Command/ReplicaDisableVirtualCommand.php
cammonro: This is a good question. In this scenario because the virtual replica is managed at a higher scope there's nothing to update because the store's virtual replica is either managed at the parent website scope or the root default scope.

If we wanted to silently handle this we would need to create an entirely new configuration and save it "on the fly" rather than edit the existing configuration which is how this command is intended to run. Not impossible but might have unintended side effects and go beyond the scope of the intended function. 

I think it's probably too much of an edge case to worry about too much. I see this CLI used in two contexts:

1. Facilitating upgrades from the old configuration and
1. Troubleshooting replica issues when limits are exceeded. 

In the first event, if they have to upgrade and are running into problems with the upgrade then they are best served by running the feature for all stores anyway. 

In the second event, the store level config is a convenience feature but not essential. They can always perform more fine grained configurations within the Magento admin.

Setup/Patch/Data/MigrateVirtualReplicaConfigPatch.php
cammonro: Yes! This patch should run first. Sequence is handled via the dependency chain. The second patch [will not run](https://github.com/algolia/algoliasearch-magento-2/pull/1562/files#diff-26b7610beb757502880d5480b94e43e1f54371981936cc24736c083400f4dab8R29) until this one has.

PR #1506, opened 2024-05-07
FIX BUG: Uncaught Error: Mismatched anonymous define() module fix

sgeleon:
Hi.

Sometimes we will get this error:  "Uncaught Error: Mismatched anonymous define() module fix".

According documentation [Magento 2](https://developer.adobe.com/commerce/php/development/build/component-file-structure/ ), we must collect all js scripts in directory JS.

```
web: Optional directory that contains static files organized into the following subdirectories:

css/source: Contains a theme's less configuration files that invoke mixins for global elements from the [UI library](https://developer.adobe.com/commerce/frontend-core/guide/css/ui-library/), and the theme.less file that overrides the default variables values.
css/source/lib: Contains view files that override the [UI library](https://developer.adobe.com/commerce/frontend-core/guide/css/ui-library/) files stored in lib/web/css/source/lib.
fonts: The folder to place the different fonts for your theme.
images: Static images folder.
js: **The folder for your JavaScript files.**
```

Magento 2 use the pattern 'js' in the path of url for marking it in the requireJS which don't know scructure of Magento 2.

This requiest fix it.

PR #1563, opened 2024-07-09
MAGE-926: JS folders reorg

damcou:
This PR applies the changes suggested here https://github.com/algolia/algoliasearch-magento-2/pull/1506 by @sgeleon (with an extra fix for a logo path) targetting the 3.14.0-beta branch.


view/frontend/layout/algolia_search_handle.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0"?>
 <page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
     <head>
-        <css src="Algolia_AlgoliaSearch::internals/autocomplete.css" />
-        <css src="Algolia_AlgoliaSearch::internals/grid.css" />
-        <css src="Algolia_AlgoliaSearch::internals/algolia-reset.css" />
-        <css src="Algolia_AlgoliaSearch::internals/instantsearch.v3.css" />
-        <css src="Algolia_AlgoliaSearch::internals/recommend.css" />
+        <css src="Algolia_AlgoliaSearch::js/internals/autocomplete.css" />
+        <css src="Algolia_AlgoliaSearch::js/internals/grid.css" />
```
cammonro, line 5: @damcou If we are aiming to follow [Magento's guidelines](https://developer.adobe.com/commerce/php/development/build/component-file-structure/#common-directories-1) it seems to me we should move CSS to a `css` directory. 

This is typically used for housing Less files in the `frontend` but I see straight CSS used in other places like in the `adminhtml` area in `Magento_GroupedProduct` and `Magento_TwoFactorAuth`. If we decide to switch to Less that would also be a more practical placement.

view/frontend/requirejs-config.js
```diff
@@ -2,29 +2,29 @@ var config = {
...
-            'algoliaInstantSearch': 'Algolia_AlgoliaSearch/instantsearch',
-            'algoliaInsights'     : 'Algolia_AlgoliaSearch/insights',
-            'algoliaHooks'        : 'Algolia_AlgoliaSearch/hooks',
+            'algoliaCommon'       : 'Algolia_AlgoliaSearch/js/internals/common',
+            'algoliaAutocomplete' : 'Algolia_AlgoliaSearch/js/autocomplete',
+            'algoliaInstantSearch': 'Algolia_AlgoliaSearch/js/instantsearch',
+            'algoliaInsights'     : 'Algolia_AlgoliaSearch/js/insights',
+            'algoliaHooks'        : 'Algolia_AlgoliaSearch/js/hooks',
 
             // Autocomplete templates
```
cammonro, line 11: @damcou If we have to move `template` anyway (which will disrupt downstream customization of Autocomplete and Recommend) then it makes sense to me to move this outside of `internals` as well. These file will be regularly customized and referenced so it never really made sense to me to include them under `internals`. 

We also already have a `template` directory (currently for an unused backend render implementation I think). Having two `template` directories might be a little confusing for some. What do you think?

cammonro (changes requested):

view/frontend/requirejs-config.js
damcou: This makes total sense your right, adding the change 👌

damcou:
Thanks for the feedbacks @cammonro , you're right, the folder structure looks better this way :)

cammonro (approved):
Awesome! Looks good.

PR #1565, opened 2024-07-10
Feature/mage 878

mrahman3177:
**Summary**
Include Non Visible Product to Index moved to admin configuration. Based on config settings product collection pulls items
Screenshot 2024-07-09 at 8 38 22 PM


Helper/ConfigHelper.php
```diff
@@ -234,6 +235,21 @@ public function getAutocompleteSelector($storeId = null)
...
 
+    /**
+     * Returns config value
+     *
+     * @param $storeId
+     * @return mixed
+     */
+    public function includeNonVisibleProductsInIndex($storeId = null)
+    {
+        return $this->configInterface->getValue(
```
cammonro, line 246: @mrahman3177 I think we should use `\Magento\Framework\App\Config\ScopeConfigInterface::isSetFlag` here instead and declare a `bool` return type not `mixed`. Let's also be sure to get in the habit of using return type declarations (PHP 7) and parameter type declarations (PHP 5).

cammonro (changes requested):

damcou (approved):
Good job 🙌

cammonro (approved):
LGTM!

PR #1569, opened 2024-07-11
MAGE-958: release number increased and change log updated

mrahman3177:
**Summary**
Release Notes:

Updates:
- Polyfill.io removed from CSP whitelist
- Hide Recommend Titles when not in use
- Incorporate community fix for higher specificity CSS selectors on Recommend

Bug Fix:
- Incorporate community fix for missing DOM element selector


damcou (approved):
🙌

PR #1571, opened 2024-07-16
MAGE-6 Fix: recommend.js get null from script when script loading bef…

sgeleon:
Hi.

I have seen your solution for when the configuration doesn't loading.

I think it’s not good because we can get other bugs if we use this module in other modules because we don’t have any guarantee that we will get object not null, for example:

require('Algolia_AlgoliaSearch/js/recommend.js',

```
function(recommend) {

    // recommend -> this is object or null ???

})
```

I think you should add some checker for it.

```

define([
    'jquery',
    'algoliaBundle',
    'recommend',
    'recommendJs',
    'recommendProductsHtml',
    'domReady!'
],function ($, algoliaBundle, recommend, recommendJs, recommendProductsHtml) {
    'use strict';

    return function (config, element) {

        function checkerConfigLoading() {
            if (typeof algoliaConfig === 'undefined')
                setTimeout(checkerConfigLoading, '500')
            else
                exec()
        }

        function exec() {

           ....

      }
});

```

### DIFF
--- a/view/frontend/web/recommend.js
+++ b/view/frontend/web/recommend.js
@@ -8,107 +8,116 @@ define([
 ],function ($, algoliaBundle, recommend, recommendJs, recommendProductsHtml) {
     'use strict';
 
-    if (typeof algoliaConfig === 'undefined') {
-        return;
-    }
-
     return function (config, element) {
-        $(function ($) {
-            this.defaultIndexName = algoliaConfig.indexName + '_products';
-            const appId = algoliaConfig.applicationId;
-            const apiKey = algoliaConfig.apiKey;
-            const recommendClient = recommend(appId, apiKey);
-            const indexName = this.defaultIndexName;
-            if ($('body').hasClass('catalog-product-view') || $('body').hasClass('checkout-cart-index')) {
-                // --- Add the current product objectID here ---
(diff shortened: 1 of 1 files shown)


damcou:
Hello @sgeleon , thanks for bringing this to our attention.
I wanted to review the changes, but the PR targeted the `develop` branch which we don't use anymore, would it be possible for you to provide the same changes targeting the [algolia:release/3.14.0-beta](https://github.com/algolia/algoliasearch-magento-2/tree/release/3.14.0-beta) branch since retargetting this one leads to some conflicts? Thanks in advance.

sgeleon:
> Hello @sgeleon , thanks for bringing this to our attention.
> I wanted to review the changes, but the PR targeted the `develop` branch which we don't use anymore, would it be possible for you to provide the same changes targeting the [algolia:release/3.14.0-beta](https://github.com/algolia/algoliasearch-magento-2/tree/release/3.14.0-beta) branch since retargetting this one leads to some conflicts? Thanks in advance.

Hello @damcou.

I created a new request.

https://github.com/algolia/algoliasearch-magento-2/pull/1624

cammonro:
Thank you for bringing this to our attention. Although the behavior you’ve described is not something we’ve witnessed, I can see how the current implementation could be improved to prevent the chance of it occurring. The polling approach you have suggested is not something we will be incorporating at this time but we will be taking your concerns into consideration in a future redesign of the FE libraries.

PR #1574, opened 2024-08-01
MAGE-891: Handle tokens expiration

damcou:
Ensure authenticated and non-authenticated tokens lifetime:
On page load If the customer is logged in, we renew `_ALGOLIA_MAGENTO_AUTH`

In such case, the `web/cookie/cookie_lifetime` value from admin is applied to follow the `PHPSSESID` cookie lifetime

### DIFF
--- a/Helper/InsightsHelper.php
+++ b/Helper/InsightsHelper.php
@@ -158,6 +158,30 @@ class InsightsHelper
     }
 
     /**
+     * Sets a token for a non-authenticated user
+     *
+     * @return void
+     */
+    public function setNonAuthenticatedUserToken(): void
+    {
+        $metaData = $this->cookieMetadataFactory->createPublicCookieMetadata()
+            ->setDuration($this->configHelper->getCookieLifetime())
+            ->setPath('/')
+            ->setHttpOnly(false)
+            ->setSecure(false);
+
+        try {
+            $this->cookieManager->setPublicCookie(
+                InsightsHelper::ALGOLIA_ANON_USER_TOKEN_COOKIE_NAME,
+                (string) $this->cookieManager->getCookie(InsightsHelper::ALGOLIA_ANON_USER_TOKEN_COOKIE_NAME),
+                $metaData
+            );
+        } catch (LocalizedException $e) {
+            $this->logger->error("Error writing anonymous user cookie: " . $e->getMessage());
+        }
+    }
+
+    /**
      * @param int|null $storeId
      * @return bool
      */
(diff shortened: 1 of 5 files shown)


Helper/InsightsHelper.php
```diff
@@ -157,6 +157,30 @@ public function setAuthenticatedUserToken(Customer $customer): string|null
...
 
+    /**
+     * Sets a token for a non-authenticated user
+     *
+     * @return void
+     */
+    public function setNonAuthenticatedUserToken(): void
+    {
+        $metaData = $this->cookieMetadataFactory->createPublicCookieMetadata()
+            ->setDuration($this->configHelper->getCookieLifetime())
```
cammonro, line 168: I'm not sure we want to refresh this token's expiration at this time. By default it is [six months](https://www.algolia.com/doc/guides/sending-events/concepts/usertoken/#adjust-the-cookie-expiration-time) and having a long lived anonymous cookie has value for tracking perso data for those who take a long time before making a purchase decision.

cammonro (changes requested):
Nice work! But unless we have a valid reason for adjusting the anonymous cookie expiration I think we should only apply the default Magento cookie lifetime to the authenticated cookie `_ALGOLIA_MAGENTO_AUTH`.

Helper/InsightsHelper.php
```diff
@@ -165,7 +168,7 @@ public function setAuthenticatedUserToken(Customer $customer): string|null
     public function setNonAuthenticatedUserToken(): void
     {
         $metaData = $this->cookieMetadataFactory->createPublicCookieMetadata()
-            ->setDuration($this->configHelper->getCookieLifetime())
+            ->setDuration(InsightsHelper::ALGOLIA_USER_TOKEN_DEFAULT_LIFETIME)
```
cammonro, line 171: @damcou Won't this perpetually renew? If so not sure if that is the desired behavior? My feeling is we should just leave the anonymous cookie alone. Perhaps however we should explicitly delete it on logout which will result in the Insights lib generating a fresh `_ALGOLIA` cookie. That probably makes the most sense. WDYT?

cammonro (changes requested):

Helper/InsightsHelper.php
damcou: @cammonro You're right, it's renewed at each page load ... Since it's not clear what should be done with the `_ALGOLIA` token, I'll remove this behaviour and remove it on logout as you suggest ;)

cammonro (approved):

cammonro:
Looks good but let's not merge just yet. Let's plan on releasing with `3.14.1`.

PR #1575, opened 2024-08-01
MAGE-961 Clarify current backend rendering behavior

cammonro:
The overall approach here is slated to be revisited in the future, but in the meantime this PR seeks to eliminate an unsupported feature and clarify the confusing configuration setting for backend render.


damcou (approved):
LGTM 👍

PR #1577, opened 2024-08-02
Feature/mage 962 - RC2

cammonro:
This PR aims to address the "Area code is already set" error encountered during Marketplace technical review (could not reproduce on Adobe Commerce Cloud) and includes some missing updates in the change log.


damcou (approved):
Looks good 🙌

rachel-trott (approved):

PR #1581, opened 2024-08-06
Feature/mage 962 - Integration test fix

cammonro:
This issue was identified by integration testing. 

During `setup:install`, Magento will perform dependency injection for all CLI commands. Unfortunately, if any constructor dependencies ultimately depend on the store manager this will trigger the dependency chain prematurely leading to the following error:

```
Base table or view not found: 1146 Table 'magento_integration_tests.flag' doesn't exist
```

By switching to proxies these dependencies are deferred until needed. 

NOTE: If not able to add to 3.14.0 let's tag for 3.14.1.


damcou (approved):
I didn't know that trick, looks good 🙌

PR #1583, opened 2024-08-08
MAGE-928: Fix generated url for insights

damcou:
The URL formatted for Insights does not handle the case where URL already has parameters (which includes a "?"), this fix adds a "&" prior to the generated string in such case.  (see related ticket in MAGE-928).


cammonro (approved):
Not a fan of the legacy code duplication here but the fix looks good and will be easy to deploy across releases. 

However, I'd like to refactor this down the road where we centralize calls in some kind of a utility library that can be loaded as a RequireJS dependency.

PR #1586, opened 2024-08-14
Quickfix: Remove a warning during the indexing of entities with no `created_at` attribute

damcou:
While performing a category reindexing in the console, I noticed an unwanted warning related to a missing check on the `created_at` attribute.

image

This should fix it.


cammonro (approved):
@damcou This seems like a sensible fix but I'm curious as to why this value would not be in the object to begin with? All products should have `created_at` data or things like "newest first" sorting will not work? Can we investigate this further? cc @mrahman3177

damcou:
> @damcou This seems like a sensible fix but I'm curious as to why this value would not be in the object to begin with? All products should have `created_at` data or things like "newest first" sorting will not work? Can we investigate this further? cc @mrahman3177

Created MAGE-994 for investigate more :)

PR #1587, opened 2024-08-20
Feature/mage 940

mrahman3177:
**Summary**

Validations put in place before enable recommendations.


Model/Observer/RecommendSettings.php
```diff
@@ -0,0 +1,216 @@
...
+            ) {
+                $this->validateFrequentlyBroughtTogether($changedPath);
+            }
+
+            // Validate before enable FBT on cart page
+            if (
+                $changedPath == $this->configHelper::IS_RECOMMEND_FREQUENTLY_BOUGHT_TOGETHER_ENABLED_ON_CART_PAGE
+                && $this->configHelper->isRecommendFrequentlyBroughtTogetherEnabledOnCartPage()
+            ) {
+                $this->validateFrequentlyBroughtTogether($changedPath);
```
cammonro, line 57: Unless I'm missing something I feel like we should be validating only once for each model. e.g. if the FBT model passed for PDP it should pass for cart no?

Model/Observer/RecommendSettings.php
```diff
@@ -0,0 +1,216 @@
...
+        } catch (\Exception $e) {
+            $this->configWriter->save($changedPath, 0);
+            throw new LocalizedException(__($e->getMessage()));
+        }
+    }
+
+    /**
+     * @return string
+     */
+    private function getProductId(): string
```
cammonro, line 202: Just curious - Why are we casting to a string? Is this because the API endpoint expects `objectID` as a string?

Model/Observer/RecommendSettings.php
```diff
@@ -0,0 +1,216 @@
...
+    {
+        if ($this->productId === '') {
+            $searchCriteria = $this->searchCriteriaBuilder
+                ->addFilter('status', 1)
+                ->create();
+            $result = $this->productRepository->getList($searchCriteria);
```
cammonro, line 208: I'm concerned about this approach. This seems like a very non performant query especially for large catalogs. We should be applying `setPageSize()` and `setCurrentPage` to the `SearchCriteria`. We also are not accounting for out of stock or non visible products (and our new index non visible product setting). These could also be filters. I honestly think this could be abstracted as a service like `SampleProductFetcher`.

Model/RecommendManagement.php
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * @param string $productId
+     * @param string $model
+     * @param float|int $threshold
+     * @return array
+     * @throws AlgoliaException
+     * @throws NoSuchEntityException
+     */
+    private function getRecommendations(string $productId, string $model, float|int $threshold = 42): array
```
cammonro, line 117: Why are we setting this default value to `42`? Is this the answer to [Life, the Universe and Everything](https://news.mit.edu/2019/answer-life-universe-and-everything-sum-three-cubes-mathematics-0910#:~:text=The%20number%2042%20is%20especially,know%20the%20answer%20to%2042.)?

cammonro: Just spitballing here but would [0](https://www.algolia.com/doc/api-reference/api-methods/get-frequently-bought-together/#method-param-threshold) perhaps be better? To increase chances of getting something back? Also think we should provide an indicator in the PHPDoc as to what this value actually does.

Model/Observer/RecommendSettings.php
```diff
@@ -0,0 +1,216 @@
...
+            $recommendations = $this->recommendManagement->getTrendingItemsRecommendation();
+            // When no recommendations suggested, most likely trained model is missing
+            if (empty($recommendations['renderingContent'])) {
+                $this->configWriter->save($changedPath, 0);
+                throw new LocalizedException(__(
+                    "It appears that there is no trained model available for Trending Items recommendation for the AppID: %1.",
+                    $this->configHelper->getApplicationID()
+                ));
+            }
+        } catch (\Exception $e) {
```
cammonro, line 171: Lots of redundancy here with the error handling. Seems we could abstract this into a more generic method?

Model/Observer/RecommendSettings.php
```diff
@@ -0,0 +1,216 @@
...
+            if (empty($recommendations['renderingContent'])) {
+                $this->configWriter->save($changedPath, 0);
+                throw new LocalizedException(__(
+                    "It appears that there is no trained model available for Frequently Bought Together recommendation for the AppID: %1.",
+                    $this->configHelper->getApplicationID()
+                ));
+            }
+        } catch (\Exception $e) {
+            $this->configWriter->save($changedPath, 0);
+            throw new LocalizedException(__($e->getMessage()));
```
cammonro, line 128: Seems to me we should be providing a more informative message in the exception. e.g.
```suggestion
            throw new LocalizedException(__("Unable to save your Recommend configuration due to the following error: " . $e→getMessage()));
```
  WDYT?

cammonro (changes requested):
Nice job @mrahman3177 ! 🙌 

Although I do have a lot of feedback. Please take a look.

Let's also be sure to rebase this against `release/3.14.1`.

Model/RecommendManagement.php
mrahman3177: 42 is my favorite number :). Just kidding. I found it in this PHP example here https://www.algolia.com/doc/rest-api/recommend/#tag/recommendations/operation/getRecommendations

Or maybe put 50 in there and not send threshold on method calls

Model/RecommendManagement.php
```diff
@@ -0,0 +1,137 @@
...
+        return $this->client;
+    }
+
+    /**
+     * @param string $name
+     * @return string
+     * @throws AlgoliaException
+     * @throws NoSuchEntityException
+     */
+    private function getFullIndexName(string $name): string
```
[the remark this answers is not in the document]
mrahman3177: Okey great! I'll use the IndexNameFetcher

Model/Observer/RecommendSettings.php
mrahman3177: I agree with out of stock and non visible filter should be added in the SearchCriteria to be in the safe side.  Adding setPageSize is also a good idea. It will put less stress on array_reverse and array_pop. I'll make that change in a few.

Model/Observer/RecommendSettings.php
mrahman3177: Yes, objectID accepts string

Model/Observer/RecommendSettings.php
mrahman3177: Respectfully disagree. Looks similar but not redundant. changedPath is different, different recommendations pulled, throws different error msg, try/catch and empty check is similar. We can try combine them together with switch case or others, but I feel like readability will be impacted.

Api/RecommendManagementInterface.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Api;
```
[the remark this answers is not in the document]
mrahman3177: Thanks for your feedback. I respectfully disagree with this. I think it would be benefited if we had a lots of interfaces under the Api directory. This is the 2nd interface in this directory. To me it make sense to keep it here.

Api/RecommendManagementInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+interface RecommendManagementInterface
```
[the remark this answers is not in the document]
mrahman3177: ManagementInterface is a standard naming convention in Magento's Service Contract design pattern. ManagementInterface is not for writing only. Any method that doesn't belong to RepositoryInterface typically goes to ManagementInterface. We can name whatever we want. I'd vote for ManagementInterface instead of TesterInterface

Model/Observer/RecommendSettings.php
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * @param Observer $observer
+     * @throws LocalizedException
+     */
+    public function execute(Observer $observer)
+    {
+        foreach ($observer->getData('changed_paths') as $changedPath) {
+            // Validate before enable FBT on PDP
+            if (
+                $changedPath == $this->configHelper::IS_RECOMMEND_FREQUENTLY_BOUGHT_TOGETHER_ENABLED
```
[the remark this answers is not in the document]
mrahman3177: Good catch! I was too carried over with "this". Will fix in a few.

Model/Observer/RecommendSettings.php
mrahman3177: I thought about it, however, I feel like that will make spaghetti code. To me this approach gives it better readability.

Model/Observer/RecommendSettings.php
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+    }
+
+    /**
+     * @param string $changedPath
+     * @return void
+     * @throws LocalizedException
+     */
+    protected function validateFrequentlyBroughtTogether(string $changedPath): void
```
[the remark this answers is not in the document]
mrahman3177: I can't imagine any other part of the code will get benefit of this validator. We are just getting recommendations and checking empty result. There is no business logic here. I feel like this validator is very specific to this observer.

Model/Observer/RecommendSettings.php
mrahman3177: That's a good point. I'll add this suggested change in a few.

Api/RecommendManagementInterface.php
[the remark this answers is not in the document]
cammonro: Not sure I understand this objection. We actually have many interfaces under `Api` (more than 2) but they are currently grouped by subdirectory. My suggestion was to try and keep things better organized. We only provide recommendations for products so that fits in my mind. I suppose we could keep it at the root level if you wish. What is the benefit to doing it this way?

Api/RecommendManagementInterface.php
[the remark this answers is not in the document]
cammonro: You make a good point about following Magento's Service Contract patterns. `TesterInterface` was a poor example. The thing is, there are all kinds of interface naming conventions in Magento but I'm not sure this qualifies as a `ManagementInterface` since all it's really doing is reading from an API. If you think it might ever do more than that then leave it. Otherwise, something like `ProviderInterface` might be another.

Model/Observer/RecommendSettings.php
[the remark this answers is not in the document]
cammonro: It's hard to anticipate everything and you shouldn't try so I agree with focusing on the task at hand. At the same time, I like to try and adhere to SRP and keep observers as light as possible as a best practice. This makes the code easier to understand, maintain and test IMO. (A validator is a simple abstraction I think.) We can cross that bridge later I suppose but oftentimes that means never which leads to tech debt... 😅

Model/Observer/RecommendSettings.php
cammonro: Hmmm. Instead we have extra processing... I'm not sure nesting conditionals would make it less readable?

Model/Observer/RecommendSettings.php
cammonro: Cool -  thanks for clarifying. Good use of casting here. 👍

Model/RecommendManagement.php
cammonro: Perhaps whoever wrote that is a Hitchhiker's Guide to the Galaxy fan. 😂 I'm in favor of whatever might make the code cleaner but perhaps the value doesn't matter - as long as the logic works consistently for the use case of verifying the model. Have we tested that having no results on a trained model still validates successfully?

Model/Observer/RecommendSettings.php
cammonro: These things could be parameterized by the generic handler method no?

Api/RecommendManagementInterface.php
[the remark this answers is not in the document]
mrahman3177: Sorry for the confusion, I was talking about interfaces in the root Api directory. There is only one interface in there now. Most ManagementInterfaces are located in the root Api directory. I prefer to follow the pattern. To me that's the default place. We should ask this, what's the benefit of moving it to product directory? :). We can try organizing it but to me it's not unorganized as it is now.

Api/RecommendManagementInterface.php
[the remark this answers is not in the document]
mrahman3177: I'm a big fan of following patterns. `ManagementInterface` sounds more native to me. This is the beginning of `RecommendManagementInterface`. I envision more recommend business logics should reside in here.

Model/Observer/RecommendSettings.php
mrahman3177: Not sure which nesting approach are you referring to. Please provide an example

Model/Observer/RecommendSettings.php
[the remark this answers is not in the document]
mrahman3177: I agree with you that we should keep observer as light as possible. Even though there is no business logic in here, there are too many validations in one observer. I can break it into 4 different observers. What do you think?

Model/Observer/RecommendSettings.php
mrahman3177: Yes we can, then more if conditions to figure out which recommendations we are dealing with, call the right method and throw appropriate errors. I think breaking this into multiple observers will look better.

Model/RecommendManagement.php
mrahman3177: Yes, I tested. I'll add my screenshots in here.

Api/RecommendManagementInterface.php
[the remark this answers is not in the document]
damcou: What about creating an `Algolia\AlgoliaSearch\Api\Recommend` subdirectory ?

Api/RecommendManagementInterface.php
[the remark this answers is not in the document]
cammonro: If you envision more logic here then this justifies the intent of the naming convention. [Naming things is hard!](https://martinfowler.com/bliki/TwoHardThings.html) Thanks for clarifying. 🙂

Model/Observer/RecommendSettings.php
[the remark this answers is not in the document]
cammonro: That's just a different way of splitting things I suppose. There is business logic though in the sense that we are ensuring that the models are valid before we allow them to be enabled. `Observer` doesn't tell me that. But `Validator` does. Let's leave it for now  - don't want to block over this.

Model/Observer/RecommendSettings.php
damcou: Something like : 
```                
if (
    (
        $changedPath == ConfigHelper::IS_RECOMMEND_FREQUENTLY_BOUGHT_TOGETHER_ENABLED 
        && $this->configHelper->isRecommendFrequentlyBroughtTogetherEnabled()
    ) || (
        $changedPath == ConfigHelper::IS_RECOMMEND_FREQUENTLY_BOUGHT_TOGETHER_ENABLED_ON_CART_PAGE
        && $this->configHelper->isRecommendFrequentlyBroughtTogetherEnabledOnCartPage()
    ) 
) {
    $this->validateFrequentlyBroughtTogether($changedPath);
}
```
If I understand the discussion ?

mrahman3177: That works for me.

cammonro: Exactly what I was thinking @damcou Thank you!

Model/RecommendManagement.php
```diff
@@ -0,0 +1,108 @@
...
+use Algolia\AlgoliaSearch\Helper\ConfigHelper;
+use Algolia\AlgoliaSearch\Service\IndexNameFetcher;
+use Magento\Framework\Exception\NoSuchEntityException;
+
+class RecommendManagement implements RecommendManagementInterface
+{
+    /**
+     * @var null|RecommendClient
+     */
+    private ?RecommendClient $client = null;
```
cammonro, line 17: Let's switch to `protected` to allow developers to customize this implementation via preference if they need.

```suggestion
    protected ?RecommendClient $client = null;
```

Model/RecommendManagement.php
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * @param ConfigHelper $configHelper
+     * @param IndexNameFetcher $indexNameFetcher
+     */
+    public function __construct(
+        private readonly ConfigHelper          $configHelper,
```
cammonro, line 24: `protected` here as well.

Model/RecommendManagement.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /**
+     * @param string $productId
+     * @param string $model
+     * @param float|int $threshold
+     * @return array
+     * @throws NoSuchEntityException
+     */
+    private function getRecommendations(string $productId, string $model, float|int $threshold = 50): array
```
cammonro, line 88:
```suggestion
    protected function getRecommendations(string $productId, string $model, float|int $threshold = 50): array
```

and so on...

Observer/RecommendSettings.php
```diff
@@ -0,0 +1,197 @@
...
+     * @return void
+     * @throws LocalizedException
+     */
+    protected function validateFrequentlyBroughtTogether(string $changedPath): void
+    {
+        try {
+            $recommendations = $this->recommendManagement->getBoughtTogetherRecommendation($this->getProductId());
+            if (empty($recommendations['renderingContent'])) {
+                throw new LocalizedException(__(
+                    "It appears that there is no trained model available for the AppID: %1.",
```
cammonro, line 101:
```suggestion
                    "It appears that there is no trained model available for Algolia application ID %1.",
```

Observer/RecommendSettings.php
```diff
@@ -0,0 +1,197 @@
...
+use Magento\Framework\Event\ObserverInterface;
+use Magento\Framework\Event\Observer;
+use Magento\Framework\Exception\LocalizedException;
+
+class RecommendSettings implements ObserverInterface
+{
+    /**
+     * @var string
+     */
+    private $productId = '';
```
cammonro, line 20:
```suggestion
    protected $productId = '';
```
etc.

Observer/RecommendSettings.php
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * @return string
+     */
+    private function getProductId(): string
+    {
+        if ($this->productId === '') {
+            $searchCriteria = $this->searchCriteriaBuilder
+                ->addFilter('status', 1)
+                ->addFilter('quantity_and_stock_status', 1)
+                ->addFilter('visibility', [2, 3, 4], 'in')
```
cammonro, line 184: Can we use constants here?

cammonro (changes requested):
Looking much better! Just a few more changes. We also need to rebase against `release/3.14.1` (or more likely at this point `release/3.14.2`)?

Observer/RecommendSettings.php
```diff
@@ -161,38 +128,66 @@ protected function validateTrendingItems(string $changedPath): void
...
+                ->addFilter(
+                    self::VISIBILITY,
+                    [
+                        Visibility::VISIBILITY_IN_CATALOG,
+                        Visibility::VISIBILITY_IN_SEARCH,
+                        Visibility::VISIBILITY_BOTH
+                    ],
+                    'in')
+                ->setPageSize(1)
+                ->setCurrentPage(1)
```
mrahman3177, line 182: setCurrentPage is default to 1, fyi.

Observer/RecommendSettings.php
```diff
@@ -161,38 +128,66 @@ protected function validateTrendingItems(string $changedPath): void
...
+    }
+
+    /**
+     * @param string $changedPath - config path to be reverted if validation failed
+     * @param string $recommendationMethod - name of method to call to retrieve method from RecommendManagementInterface
+     * @param string $modelName - user friendly name to refer to model in error messaging
+     * @return void
+     * @throws LocalizedException
+     */
+    protected function validateRecommendation(string $changedPath, string $recommendationMethod, string $modelName): void
```
mrahman3177, line 142: Ok, that's what you meant by parameterization. It looks much better for sure. Thanks for spending time and put it together.

Observer/RecommendSettings.php
cammonro: Good point - I'll remove that!

cammonro (approved):

PR #1590, opened 2024-08-22
MAGE-1010 Fix Magento coding standards error with proxy injection

cammonro:
CodeSniffer yields a complaint on our injection method for proxy objects. Using `di.xml` over constructor is the preferred approach here.

Before applying fix:
image
After:
image


mrahman3177 (approved):
Looks good to me!

PR #1591, opened 2024-08-26
MAGE-999: algoliaSearchLib added to requireJs

mrahman3177:
**Summary**

algoliaSearchLib added to requireJs


view/frontend/web/js/lib/algolia-search.min.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Skipped minification because the original files appears to be already minified.
+ * Original file: /npm/algoliasearch@5.1.1/dist/algoliasearch.umd.js
+ *
+ * Do NOT use SRI with dynamically generated files! More information: https://www.jsdelivr.com/using-sri-with-dynamic-files
+ */
+/*! algoliasearch.umd.js | 5.1.1 | © Algolia, inc. | https://github.com/algolia/algoliasearch-client-javascript */
```
cammonro, line 7: Version 5 of `algoliasearch` lib is brand new and though it's hard to tell the exact version of the bundle (because the original repo was never updated properly) we are most definitely on some flavor of [4.x](https://github.com/algolia/algoliasearch-extensions-bundle/blob/ISv4/package.json#L17). 

So my feeling here is that while we should update to v5 of the library as soon as possible, we should _not_ do so until we've successfully migrated off of `algoliaBundle`. I think it would be best at this stage of the journey to only update to the last stable version 4 of `algoliasearch` which appears to be [4.24.0](https://www.jsdelivr.com/package/npm/algoliasearch?version=4.24.0&tab=files). 

The reason for my thinking is that by doing this we minimize the variables of the migration and don't end up chasing our tails on something that could be related to the v5 upgrade and not actually the bundle migration. What do you think?

cammonro (changes requested):
Please see my comments and let me know what you think about staying on v4.x of `algoliasearch`? 

Also can we please rebase the epic branch off of `3.14.1-rc2`?

view/frontend/web/js/lib/algolia-search.min.js
mrahman3177: I totally agree with this. Removing algoliaBundle is a big task and we want this to be smooth. We don't want to introduce additional challenges in this task. One step at a time make sense to me. Will fix it soon.

cammonro (approved):
Right on!

view/frontend/web/js/lib/algolia-search.min.js
martinslaganovskis: Hi, @cammonro, since the [EOL for v4](https://www.algolia.com/doc/libraries/sdk/v1/versions#javascript) is coming, are you planning to update the module to support v5 till August deadline? Or do you suggest we update the libraries ourselves.

cammonro: @martinslaganovskis Thanks for calling this out! 

Yes the plan would be to update to version 5 before that time. If you need it sooner, we've removed the `algoliaBundle` dependency so it's now much easier for end users to do it themselves. 

But we will need to fully test and verify that there are no issues before we ship it on our end. 

cc @damcou @rachel-trott

martinslaganovskis: Hi @cammonro , just wanted to know if there is any progress on the v5 update? We are getting close to the point where we will need proceed with the upgrade on our side so just wanted to see if perhaps its already coming (so we don't do double work). Thank you!

cammonro: Hi @martinslaganovskis We just shipped the 3.19 beta for ingestion but it does not yet include v5 of the JS lib. Updating the JS lib was the plan for 3.19 GA - but I'll need to sync with the team on Monday to confirm and get back to you. Getting all our JS clients on v5 by August is important so I appreciate you flagging this!

cammonro: Hey @martinslaganovskis - just wanted to hit you back and let you know that yes we are planning on shipping v5 of the Algolia JS API client with v3.19 next month. 👍

PR #1593, opened 2024-08-26
MAGE-999: autocomplete umd added to the library and algoliaSearch umd…

mrahman3177:
**Summary**
Autocomplete umd js added to the library and algoliaSearch implemented for the authentication.


view/frontend/web/js/autocomplete.js
```diff
@@ -13,6 +15,8 @@ define([
 ], function (
     $,
     algoliaBundle,
+    algoliaSearch,
```
cammonro, line 18:
```suggestion
    algolisearch,
```

view/frontend/web/js/autocomplete.js
```diff
@@ -13,6 +15,8 @@ define([
...
+    algoliaAutocomplete,
```
cammonro, line 19:
```suggestion
    autocomplete,
```

view/frontend/web/js/autocomplete.js
```diff
@@ -39,7 +43,7 @@ define([
      * Initialise Algolia client
      * Docs: https://www.algolia.com/doc/api-client/getting-started/instantiate-client-index/
      **/
-    const searchClient = algoliaBundle.algoliasearch(
+    const searchClient = algoliaSearch(
```
cammonro, line 46:
```suggestion
    const searchClient = algoliasearch(
```

cammonro (changes requested):
This is looking good but I think we should stick to Algolia lib naming conventions within our code but use prefixing for dereferencing in RequireJs.

view/frontend/web/js/autocomplete.js
mrahman3177: That sounds good to me. A little more native than camel case.

view/frontend/web/js/autocomplete.js
mrahman3177: Make sense. Keeping object name smaller have much benefit. Will fix it in a few.

cammonro (approved):

PR #1596, opened 2024-08-27
Feature/mage 1000 - Mustache alt

cammonro:
Alt Mustache lib. 

Please see MAGE-1000 for details.


cammonro:
Decided to make this a little more pluggable for possible future applications when we start modernizing things a bit more... 

Can call two ways:

```js
const html = await templateEngine.processTemplate(template, data, true);
```

or …

```js
const templateProcessor = await templateEngine.getSelectedEngineAdapter(); 
const html = templateProcessor.process(template, data);
```

or …

```js
const templateProcessor = await templateEngine.getSelectedEngineAdapter(); 
const html = templateProcessor.processAndMeasure(template, data, 'STATS');
```

The last two approaches are for when you cannot use async / await as not all callbacks in InstantSearch accept promises - e.g. the `stats` widget.

We could make this configurable in the future. Meanwhile a user could conceivably select the templating engine they desire via a RequireJS mixin.

Enable with: 

```js
config: {
  mixins: {
      "Algolia_AlgoliaSearch/js/internals/template-engine": {
          "Algolia_CustomAlgolia/js/internals/template-engine-mixin": true,
      }
  },
},

```
A sample template engine mixin might look like this: 
```js
// SAMPLE TEMPLATE ENGINE MIXIN
define(function () {
    "use strict";

    return function (target) {
        const mixin = {
            getSelectedEngineType: function () {
                return target.ENGINE_TYPE_HOGAN;
            },
        };

        return { ...target, ...mixin };
    };
});

```

damcou (approved):
LGTM 🚀

PR #1598, opened 2024-08-29
MAGE-980: Application credentials validation

damcou:
Created an AlgoliaCredentialsManager class and centralize the credentials validation all accross the extension

- ApplicationID + Admin APIKey

```
$this->config->getApplicationID($storeId) && $this->config->getAPIKey($storeId)

// becomes

$this->algoliaCredentialsManager->checkCredentials($storeId)`
```

- ApplicationID + Admin APIKey + SearchOnly ApiKey

```
$this->configHelper->getApplicationID()
  && $this->configHelper->getAPIKey()
  && $this->configHelper->getSearchOnlyAPIKey()

// becomes

$this->algoliaCredentialsManager->checkCredentialsWithSearchOnlyAPIKey($storeId)
```

The algoliaCredentialsManager also handles the display of the error messages (both in console and Magento Admin)

⚠️ : In some places, it's impossible to define which store should be used, so I decided to make the `$storeId` parameter optional (in such case, the check remains unchanged compared to before)

### DIFF
--- a/Helper/AdapterHelper.php
+++ b/Helper/AdapterHelper.php
@@ -4,6 +4,7 @@ namespace Algolia\AlgoliaSearch\Helper;
 
 use Algolia\AlgoliaSearch\Helper\Adapter\FiltersHelper;
 use Algolia\AlgoliaSearch\Helper\Data as AlgoliaDataHelper;
+use Algolia\AlgoliaSearch\Service\AlgoliaCredentialsManager;
 use Magento\CatalogSearch\Helper\Data as CatalogSearchDataHelper;
 
 class AdapterHelper
@@ -11,35 +12,14 @@ class AdapterHelper
     public const INSTANTSEARCH_ORDER_PARAM = 'sortBy';
     public const BACKEND_ORDER_PARAM = 'product_list_order';
 
-    /** @var CatalogSearchDataHelper */
-    private $catalogSearchHelper;
-
-    /** @var AlgoliaDataHelper */
-    private $algoliaHelper;
-
-    /** @var FiltersHelper */
-    private $filtersHelper;
-
-    /** @var ConfigHelper */
-    private $configHelper;
-
-    /**
-     * @param CatalogSearchDataHelper $catalogSearchHelper
-     * @param AlgoliaDataHelper $algoliaHelper
-     * @param FiltersHelper $filtersHelper
-     * @param ConfigHelper $configHelper
-     */
     public function __construct(
-        CatalogSearchDataHelper $catalogSearchHelper,
-        AlgoliaDataHelper $algoliaHelper,
-        FiltersHelper $filtersHelper,
-        ConfigHelper $configHelper
-    ) {
-        $this->catalogSearchHelper = $catalogSearchHelper;
-        $this->algoliaHelper = $algoliaHelper;
-        $this->filtersHelper = $filtersHelper;
-        $this->configHelper = $configHelper;
-    }
+        protected CatalogSearchDataHelper $catalogSearchHelper,
+        protected AlgoliaDataHelper $algoliaHelper,
+        protected FiltersHelper $filtersHelper,
+        protected ConfigHelper $configHelper,
+        protected AlgoliaCredentialsManager $algoliaCredentialsManager
+    )
+    {}
 
     /**
      * Get search result from Algolia
@@ -142,8 +122,7 @@ class AdapterHelper
         $storeId = $this->getStoreId();
 
         return
-            $this->configHelper->getApplicationID($storeId)
-            && $this->configHelper->getAPIKey($storeId)
+            $this->algoliaCredentialsManager->checkCredentials($storeId)
             && $this->configHelper->isEnabledFrontEnd($storeId)
             && $this->configHelper->makeSeoRequest($storeId);
     }
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -78,7 +78,7 @@ class AlgoliaHelper extends AbstractHelper
     /**
      * @return void
      */
-    public function createClient(): void
+    protected function createClient(): void
     {
         $storeId = $this->getStoreId();
         if ($this->algoliaCredentialsManager->checkCredentials($storeId)) {
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -6,6 +6,7 @@ use Algolia\AlgoliaSearch\Api\SearchClient;
 use Algolia\AlgoliaSearch\Configuration\SearchConfig;
 use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
 use Algolia\AlgoliaSearch\Exceptions\ExceededRetriesException;
+use Algolia\AlgoliaSearch\Service\AlgoliaCredentialsManager;
 use Algolia\AlgoliaSearch\Support\AlgoliaAgent;
 use Magento\Framework\App\Helper\AbstractHelper;
 use Magento\Framework\App\Helper\Context;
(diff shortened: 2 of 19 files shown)


Helper/AdapterHelper.php
```diff
@@ -4,42 +4,22 @@
...
-
-    /** @var ConfigHelper */
-    private $configHelper;
-
-    /**
-     * @param CatalogSearchDataHelper $catalogSearchHelper
-     * @param AlgoliaDataHelper $algoliaHelper
-     * @param FiltersHelper $filtersHelper
-     * @param ConfigHelper $configHelper
-     */
```
mrahman3177, line 31: Just curious, why did we remove the doc blocks from constructors?

Helper/ConfigHelper.php
```diff
@@ -1198,17 +1198,6 @@ public function setCustomerGroupsEnabled(bool $val, ?string $scope = null, ?int
...
-     * @param $storeId
-     * @return bool
-     */
-    public function credentialsAreConfigured($storeId = null)
-    {
-        return $this->getApplicationID($storeId) &&
-            $this->getAPIKey($storeId) &&
-            $this->getSearchOnlyAPIKey($storeId);
-    }
-
```
mrahman3177, line 1211: Can we keep this method and mark as deprecated and then remove it after EOL of this release?

Helper/AdapterHelper.php
damcou: It was made to be consistent with the latest added classes which have no blocks, I also like to remove lines as it doesn't really bring any value in this case.

Helper/ConfigHelper.php
damcou: Sure, I'll re-add it :)

Service/AlgoliaCredentialsManager.php
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public function hasMultipleApplicationIDs(): bool
+    {
+        $applications = [];
+
+        foreach ($this->storeManager->getStores() as $store) {
+            $storeId = $store->getId();
+            if ($this->checkCredentials($storeId)
+                && !isset($applications[$this->configHelper->getApplicationID($storeId)])
+            ) {
```
cammonro, line 80: We could short circuit this operation to return `true` as soon as a second App ID is found. Might provide a slight performance boost on a large number of stores.

Setup/Patch/Data/RebuildReplicasPatch.php
```diff
@@ -52,7 +52,7 @@ public function getAliases(): array
      */
     public function apply(): PatchInterface
     {
-        if (!$this->configHelper->credentialsAreConfigured()) {
+        if (!$this->algoliaCredentialsManager->checkCredentialsWithSearchOnlyAPIKey()) {
```
cammonro, line 55: With multi app now supported I'm wondering if we should include this in the store ID loop?

Model/Indexer/AdditionalSection.php
```diff
@@ -41,30 +26,21 @@ public function execute($ids)
...
 
+            if (!$this->algoliaCredentialsManager->checkCredentialsWithSearchOnlyAPIKey($storeId)) {
+                $this->algoliaCredentialsManager->displayErrorMessage(self::class, $storeId);
+
+                return;
+            }
+
             $this->queue->addToQueue(
-                $this->fullAction,
+                Data::class,
```
cammonro, line 43: Good call! I like this consistency.

Helper/ConfigHelper.php
```diff
@@ -1201,6 +1201,7 @@ public function setCustomerGroupsEnabled(bool $val, ?string $scope = null, ?int
     /**
      * @param $storeId
      * @return bool
+     * @deprecated Use Algolia\AlgoliaSearch\Service\AlgoliaCredentialsManager instead
```
cammonro, line 1204: Good call out on a public method.

Helper/AlgoliaHelper.php
```diff
@@ -95,7 +81,7 @@ public function __construct(
     public function createClient(): void
```
cammonro, line 81: Does this need to be public if only called by `AlgoliaHelper`? As it manages internal state I'm not sure it would be a good candidate for a possible future plugin.

Model/Observer.php
```diff
@@ -14,57 +16,39 @@
  */
 class Observer implements ObserverInterface
```
cammonro, line 17: Just a side note - for our next minor release we should move this under the `Observer` namespace and give it a better name.

cammonro (changes requested):
Awesome job. 😎 

A few questions / concerns if you wouldn't mind taking a look. 

Also perhaps we should consider providing an API interface for the new `AlgoliaCredentialsManager` class?

Helper/AlgoliaHelper.php
damcou: good call, updating it

Model/Observer.php
damcou: understood, added this to my todo list ;)

Setup/Patch/Data/RebuildReplicasPatch.php
damcou: We could consider adding the logic a little bit below (line 69), but I'm wondering why there are two separate foreach ? Do you need to finish all the deletes before starting the syncs ?

Service/AlgoliaCredentialsManager.php
damcou: updating

Setup/Patch/Data/RebuildReplicasPatch.php
cammonro: Good question. My reasons here were two fold, 
1. I was trying to avoid coupling the logic in the `ReplicaManager` class to keep things reusable. 
2. I found that it was possible for replica misconfigurations to cross pollinate across store views which would prevent a surgical "one store at a time" approach to effectively rebuild things. Better to blow it all away first. 

As far as the validation - I think just putting it in the first loop would be sufficient?

damcou: I've decided to check the two loops, but display the message only for the first one, if that's ok for you ? https://github.com/algolia/algoliasearch-magento-2/pull/1598/commits/7bef4ccbe69d600bb9352237c021a83bff8510b9

Setup/Patch/Data/RebuildReplicasPatch.php
```diff
@@ -67,10 +62,19 @@ public function apply(): PatchInterface
         $storeIds = array_keys($this->storeManager->getStores());
         // Delete all replicas before resyncing in case of incorrect replica assignments
         foreach ($storeIds as $storeId) {
+            if (!$this->algoliaCredentialsManager->checkCredentialsWithSearchOnlyAPIKey($storeId)) {
+                $this->logger->warning("Algolia credentials are not configured for store $storeId. Aborting replica rebuild patch. If you need to rebuild your replicas run `bin/magento algolia:replicas:rebuild`");
```
cammonro, line 66:
```suggestion
                $this->logger->warning("Algolia credentials are not configured for store $storeId. Skipping auto replica rebuild for this store. If you need to rebuild your replicas run `bin/magento algolia:replicas:rebuild`");
```
Just a little tweak to the warning message language I think is warranted here.

Setup/Patch/Data/RebuildReplicasPatch.php
cammonro: Yes I'm good with that guard clause style and I think the way you've now handled this scenario is a little more graceful and works better for multi app. Just one small tweak to the language in the log message is suggested as we are not aborting the patch but just moving on to the next store.

cammonro (approved):
Good to go! 👍

PR #1603, opened 2024-08-30
Release/3.14.1 - patch release

cammonro:
### Updates
- Token clean up on session expiration

### Bug Fixes
- Query string append issue addressed
- Fixed issue with "base table not found" in integration tests due to new CLI command classes
- Removed a warning during indexing of entities with no `created_at` attribute
- Fixed incorrect Recommend paths resulting from earlier JS asset reorg
- README.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # CHANGE LOG
 
+## 3.14.1
+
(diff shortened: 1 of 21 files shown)


mrahman3177 (approved):
Looks good to me.

PR #1604, opened 2024-09-03
MAGE-1018: release version upgraded to 3.14.2

mrahman3177:
### **Summary**

**Release Notes:**

**Updates:**
 - Trained model check before enabling recommend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGE LOG
 
+## 3.14.2
+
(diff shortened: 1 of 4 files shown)


cammonro (changes requested):
Can we update [this section](https://github.com/algolia/algoliasearch-magento-2/blob/feature/MAGE-1018/README.md?plain=1#L65) too?

cammonro (approved):

PR #1605, opened 2024-09-04
Feature/mage 998 unbundle instantsearch

cammonro:
This PR removes `algoliaBundle` entirely but work remains to be done for making the optional baggage available from the old bundle (scoped in MAGE-2020). I also am leaving the bundle's minified lib and path alias in RequireJS intentionally for any who may be using it in their custom scripts. 

A mock is provided for `algoliaBundle` that is passed to hooks but a follow up will make this more configurable for users who may need the old stuff. (I can't imagine why though since it is a global dependency that can be loaded with `require` etc. but I would like to try and cover our bases here).

I took the liberty of performing some reorg of the namespaces for Autocomplete, Insights and Recommend for consistency sake.

The changes to `paths` and `map` are intentional to prevent old references from breaking but also not allow them to load the dependency twice because we reference it one way in the core module but a customization might reference the lib using the old name.


view/frontend/requirejs-config.js
```diff
@@ -1,5 +1,5 @@
-var config = {
-    map   : {
+const config = {
```
mrahman3177, line 1: @cammonro , I'm sure there is a good reason why you changed it to `const`. But it looks a little odd to me. 
With my limited FE knowledge, I believe `const` variable is used so that it can not be updated. However, all config should be merged together and create a global config. So, `var` instead of `const` make sense to me. 

Also, I looked at Magento vendor library, I couldn't find Magento is shifting towards `const`.

view/frontend/web/js/instantsearch.js
```diff
@@ -1,12 +1,44 @@
...
+     * However if you've used or require any of these additional libs in your customizations,
+     * you can either augment this mock as you need or include the global dependency in your module
+     * and make it available to your hook.
+     * TODO: Mixin and documentation to come on how to do this... 
+     */
+    const mockAlgoliaBundle = {
+        $,
+        algoliasearch,
+        instantsearch
+    };
```
damcou, line 40: Interesting! Looking forward to see it working :)

damcou (approved):
Good job ! I don't see anything blocking here, this JS reorg seems really promising ! 🙌

cammonro:
Thanks guys! Appreciate the feedback. 🙏

PR #1607, opened 2024-09-06
MAGE-1023 Fix landing page typing

cammonro:
## Summary
This fixes a critical bug with landing pages render introduced in `3.14.0` by PHP strong typing on method return values. 

## Proof of fix
Creating a landing page for "blue shorts":
image
Store front render:
image

### DIFF
--- a/Block/Algolia.php
+++ b/Block/Algolia.php
@@ -215,7 +215,7 @@ class Algolia extends Template implements CollectionDataSourceInterface
         return $this->_urlBuilder->getUrl('checkout/cart/add', $routeParams);
     }
 
-    protected function getCurrentLandingPage(): LandingPage|null|false
+    protected function getCurrentLandingPage(): \Algolia\AlgoliaSearch\Model\LandingPage|null|false
     {
         $landingPageId = $this->getRequest()->getParam('landing_page_id');
         if (!$landingPageId) {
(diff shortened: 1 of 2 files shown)


mrahman3177 (approved):
LGTM!

Block/Algolia.php
```diff
@@ -215,7 +215,7 @@ protected function getAddToCartUrl($additional = []): string
         return $this->_urlBuilder->getUrl('checkout/cart/add', $routeParams);
     }
 
-    protected function getCurrentLandingPage(): LandingPage|null|false
+    protected function getCurrentLandingPage(): \Algolia\AlgoliaSearch\Model\LandingPage|null|false
```
mrahman3177, line 218: Hey @cammonro , I know I approved it. I had an assumption that I would like to clear. The assumption I had was, it's required to have the full path here. Did you try importing this object and using the sort form? If it works with short form, I'd suggest to use that for consistency.

cammonro: @mrahman3177 Well the full import is definitely intentional to distinguish `\Algolia\AlgoliaSearch\Model\LandingPage` from the local `\Algolia\AlgoliaSearch\Block\LandingPage`. I could add an import but as it was only one occurrence I did the once off qualifier.

cammonro: I could make this really clear with an alias but let me know what you think.

PR #1609, opened 2024-09-09
MAGE-1021: Additional changes for multi application IDs

damcou:
This PR added some changes requested during last review:
- Added missing @throws tags and exceptions handling
- `addSegments` renamed to `addAlgoliaUserAgent`
- `Model\Observer` moved to `Observer\AddAlgoliaAssetsObserver`

### DIFF
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -159,7 +159,7 @@ class AlgoliaHelper extends AbstractHelper
     }
 
     /**
-     * @return ListIndicesResponse
+     * @return ListIndicesResponse|array<string,mixed>
      * @throws AlgoliaException
      */
     public function listIndexes()
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -6,8 +6,11 @@ use Algolia\AlgoliaSearch\Api\SearchClient;
 use Algolia\AlgoliaSearch\Configuration\SearchConfig;
 use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
 use Algolia\AlgoliaSearch\Exceptions\ExceededRetriesException;
+use Algolia\AlgoliaSearch\Model\Search\ListIndicesResponse;
+use Algolia\AlgoliaSearch\Model\Search\SettingsResponse;
 use Algolia\AlgoliaSearch\Service\AlgoliaCredentialsManager;
 use Algolia\AlgoliaSearch\Support\AlgoliaAgent;
+use Exception;
 use Magento\Framework\App\Helper\AbstractHelper;
 use Magento\Framework\App\Helper\Context;
 use Magento\Framework\Message\ManagerInterface;
@@ -56,6 +59,9 @@ class AlgoliaHelper extends AbstractHelper
 
     protected static ?int $lastTaskId;
 
+    /**
+     * @throws AlgoliaException
+     */
     public function __construct(
         Context $context,
         protected ConfigHelper $config,
@@ -66,7 +72,7 @@ class AlgoliaHelper extends AbstractHelper
         parent::__construct($context);
 
         $this->createClient();
-        $this->addSegments();
+        $this->addAlgoliaUserAgent();
 
         // Merge non castable attributes set in config
         $this->nonCastableAttributes = array_merge(
@@ -77,6 +83,7 @@ class AlgoliaHelper extends AbstractHelper
 
     /**
      * @return void
+     * @throws AlgoliaException
      */
     protected function createClient(): void
     {
@@ -99,7 +106,7 @@ class AlgoliaHelper extends AbstractHelper
      * @return void
      * @throws AlgoliaException
      */
-    protected function addSegments(): void
+    protected function addAlgoliaUserAgent(): void
     {
         $clientName = $this->getClient()->getClientConfig()?->getClientName();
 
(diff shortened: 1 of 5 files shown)


Service/AlgoliaCredentialsManager.php
```diff
@@ -46,22 +46,24 @@ public function checkCredentialsWithSearchOnlyAPIKey(int $storeId = null): bool
...
 
-            return;
-        }
+                return;
+            }
 
-        $this->messageManager->addErrorMessage($errorMessage);
+            $this->messageManager->addErrorMessage($errorMessage);
+        } catch (NoSuchEntityException $exception) {
+        }
```
cammonro, line 66: I don't think we should bury this exception as unlikely as it may be? Maybe something like: 
```
$this->messageManager->addErrorMessage(__("Unable to locate store details: %1", $exception->getMessage()));
```
Or lift the `$errorMessage` scope up and concatenate the errors. (Because it might hide the original issue?)

Service/AlgoliaCredentialsManager.php
```diff
@@ -46,22 +46,24 @@ public function checkCredentialsWithSearchOnlyAPIKey(int $storeId = null): bool
      * @param string $class
      * @param int|null $storeId
      * @return void
-     * @throws NoSuchEntityException
      */
     public function displayErrorMessage(string $class, ?int $storeId = null): void
```
cammonro, line 50: I've observed we tend to call this method immediately after `checkCredentialsWithSearchOnlyAPIKey` in several places. Should we just do that automatically from that method and not have to repeat it everywhere? Or make displaying the message an optional boolean parameter in `checkCredentialsWithSearchOnlyAPIKey`?

etc/frontend/events.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Event/etc/events.xsd">
     <event name="layout_load_before">
-        <observer name="algolia_injection" instance="Algolia\AlgoliaSearch\Model\Observer" />
```
cammonro, line 4: Can we move this to a separate PR on `release/3.15.0-beta`?

Helper/AlgoliaHelper.php
```diff
@@ -152,7 +159,7 @@ public function getIndex(string $name)
     }
 
     /**
-     * @return mixed
+     * @return ListIndicesResponse
```
cammonro, line 162: I like that we're adding more specificity here but I've wondered about the response objects in the PHP client API. 

I see the API call for `\Algolia\AlgoliaSearch\Api\SearchClient::listIndices` can return:
```
\Algolia\AlgoliaSearch\Model\Search\ListIndicesResponse|array<string, mixed>
```

Do we need to provide a union type in our code as well? When would one be returned vs. the other?

I would hate for an unexpected type mismatch to create an issue here.

cammonro (changes requested):
@damcou Thanks for making these improvements! I do have some questions / concerns if you wouldn't mind taking a look.

Helper/AlgoliaHelper.php
damcou: You're right with this, indeed, the php client most probably return an array, changing this right now

Helper/AlgoliaHelper.php
```diff
@@ -757,9 +769,9 @@ public function isValidFloat(string $value) : bool {
 
     /**
      * @param $value
-     * @return float|int
+     * @return mixed
      */
-    protected function castAttribute($value)
+    protected function castAttribute($value): mixed
```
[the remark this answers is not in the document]
damcou: That's what I put originally, but while I tested it, it broke the indexing.
The fact is this current @return tag is wrong, the method return basically anything (I had to put `float|int|string|null|array` to make it work again, so I thought that `mixed` would be better.
We can also not specify any return type for this one if you prefer, WDYT ?

Observer/AddAlgoliaAssetsObserver.php
```diff
@@ -1,11 +1,13 @@
 <?php
 
-namespace Algolia\AlgoliaSearch\Model;
```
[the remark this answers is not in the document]
damcou: Understood, gonna roll this back here and create a separate PR 👍

Service/AlgoliaCredentialsManager.php
damcou: The main advantage of this method is to pass the current class so the user can know where the issue happened. So we would need to pass the information to `checkCredentialsWithSearchOnlyAPIKey` with this change. I think this would unnecessarily complicate the signature of this method with something which has nothing to do with credentials check... 
I also like the fact that the logic within the indexers stays humanly readable ("if my credentials are not speficied, then I display an error message") .

This is called 7 times for now in the codebase, I guess this is something we can deal with, WDTY ?

Service/AlgoliaCredentialsManager.php
damcou: Makes sense, updating it.

Helper/AlgoliaHelper.php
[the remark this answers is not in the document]
cammonro: You know on second look I can see that it might not always return those values. It's an odd bit of code - I say keep it `mixed` then. If anything it calls out a need for a potential refactor.

Service/AlgoliaCredentialsManager.php
cammonro: Yeah it's cool. No need to "DRY" ourselves to death. Thanks for thinking about it anyway!

cammonro (approved):
LGTM!

PR #1610, opened 2024-09-09
MAGE-1023 Use alias to clearly distinguish type

cammonro:
Clarify intended type via import alias.


mrahman3177 (approved):
Looks good. Thanks!

PR #1612, opened 2024-09-11
MAGE-983: Fix facet query rules at store level

damcou:
During the test for the multi-application IDs, I noticed that facet query rules configuration wasn't taken into account at store level. This change should fix it.


mrahman3177 (approved):
LGTM

cammonro (approved):

PR #1617, opened 2024-09-17
Feature/mage 899 mixin compat

cammonro:
This PR aims to:
- Treat Algolia libs that inject Autocomplete and InstantSearch as proper `uiComponents`
- Eliminate dependencies on the global `window` object
- Maintain `window` objects for backward compatibility
- Eliminate monolithic anonymous functions
- Allow the use of RequireJS mixins to override targeted functionality in the Algolia Magento extension vs overriding entire JS files enabling code reuse and preventing implementation drift on future updates (from copy/paste overrides) 
- Ensure that modules return mixable objects and that dependencies are loaded properly through the `define` method signature
- Keep UI lib callbacks as pure as possible (and eliminate pollution with obscure implementation) 
- Reduce scoping of variables across mega methods
- Remove and/or modernize dated / obsolete code

### DIFF
--- a/view/adminhtml/web/css/landing-page.css
+++ b/view/adminhtml/web/css/landing-page.css
@@ -1069,53 +1069,6 @@ a.ais-current-refined-values--link:hover {
     margin-bottom: 20px;
 }
 
-/** Search Box */
-
-#algolia-searchbox {
-    position: relative;
-}
-
-#algolia-searchbox .clear-cross, #algolia_instant_selector .clear-cross {
-    position: absolute;
-    display: none;
-    cursor: pointer;
-    width: 16px;
-    height: 16px;
-}
-
-#algolia-searchbox .clear-query-autocomplete {
-    bottom: 22px;
-    right: 9px;
-}
-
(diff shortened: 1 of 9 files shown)


mrahman3177 (approved):
Looks good to me

damcou (approved):
Huge work @cammonro ! 🙌 
Tested locally and it seems to work like a charm ! 🔥 
I really like the new structure of the autocomplete and instantsearch js file, this is way more readable this way (and probably more maintainable as well !) 

Small question about the CSS removal, should I assume the code related to searchbox was totally useless ?

cammonro:
Thanks for the feedback! 

> Small question about the CSS removal, should I assume the code related to searchbox was totally useless ?

No I wouldn't say totally useless because it was clearly impacting customers who were using those old selectors. However, eliminating old code in general prevents the risk of bleed over into customer implementations. 

That particular bit of code really should have been removed 2 years ago when Autocomplete moved from version 0 → version 1 in v`3.8` but removing code can have unintended consequences as well so it's important to always double check!

PR #1618, opened 2024-09-18
MAGE-1030: Fix client creation

damcou:
This change includes:
- Removal of client creation in the constructor to avoid such issues in the future
- Prevented unhandled crash on category merchandising when no credentials are configured at Default level
- Feedback to the user is now handled by the `AlgoliaCredentialsManager` class

### DIFF
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -85,6 +85,7 @@ class AlgoliaHelper extends AbstractHelper
             if ($storeId !== self::ALGOLIA_DEFAULT_SCOPE) {
                 $this->algoliaCredentialsManager->displayErrorMessage(AlgoliaHelper::class, $storeId);
             }
+            return;
         }
 
         $config = SearchConfig::create(
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -55,6 +55,9 @@ class AlgoliaHelper extends AbstractHelper
     /** @var int  */
     protected int $storeId = self::ALGOLIA_DEFAULT_SCOPE;
 
+    /** @var bool */
+    protected bool $userAgentsAdded = false;
+
     protected static ?string $lastUsedIndexName;
 
     protected static ?int $lastTaskId;
@@ -77,15 +80,13 @@ class AlgoliaHelper extends AbstractHelper
 
     /**
      * @return void
+     * @throws AlgoliaException
      */
     protected function createClient(): void
     {
         $storeId = $this->getStoreId();
         if (!$this->algoliaCredentialsManager->checkCredentials($storeId)) {
-            if ($storeId !== self::ALGOLIA_DEFAULT_SCOPE) {
-                $this->algoliaCredentialsManager->displayErrorMessage(AlgoliaHelper::class, $storeId);
-            }
-            return;
+            throw new AlgoliaException('Client initialization could not be performed because Algolia credentials were not provided.');
         }
 
         $config = SearchConfig::create(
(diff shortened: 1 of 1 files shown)


Helper/AlgoliaHelper.php
```diff
@@ -83,23 +77,25 @@ public function __construct(
...
-            $config = SearchConfig::create(
-                $this->config->getApplicationID($storeId),
-                $this->config->getAPIKey($storeId)
-            );
-            $config->setConnectTimeout($this->config->getConnectionTimeout($storeId));
-            $config->setReadTimeout($this->config->getReadTimeout($storeId));
-            $config->setWriteTimeout($this->config->getWriteTimeout($storeId));
-            $this->clients[$storeId] = SearchClient::createWithConfig($config);
-        } else {
-            throw new AlgoliaException('Client initialization could not be performed because Algolia credentials were not provided.');
```
cammonro, line 101: I agree we shouldn't be throwing an exception in the constructor but is there a reason we are eliminating it altogether?

Helper/AlgoliaHelper.php
```diff
@@ -83,23 +77,25 @@ public function __construct(
...
+        if (!$this->algoliaCredentialsManager->checkCredentials($storeId)) {
+            if ($storeId !== self::ALGOLIA_DEFAULT_SCOPE) {
```
cammonro, line 85: What is the intent behind this logic?

Helper/AlgoliaHelper.php
```diff
@@ -119,16 +115,17 @@ protected function addAlgoliaUserAgent(): void
...
-     * @return SearchClient
+     * @return SearchClient|null
      * @throws AlgoliaException
      */
-    public function getClient(): SearchClient
+    public function getClient(): ?SearchClient
     {
         if (!isset($this->clients[$this->getStoreId()])) {
             $this->createClient();
+            $this->addAlgoliaUserAgent();
```
cammonro, line 125: Won't this get called redundantly on the static class method? Or do we care?

cammonro (changes requested):
It's good we caught this scenario. Definitely don't want exceptions preventing injection of the helper class. However, I am concerned about the approach of not throwing exceptions at all.

Helper/AlgoliaHelper.php
damcou: I thought it was better to reuse the method of the `AlgoliaCredentialsManager` class which handles both CLI and admin messages "gracefully". I don't mind changing it if you prefer.

Helper/AlgoliaHelper.php
damcou: @cammonro This is an interesting question. During my tests, I assumed that the user has to fill the credentials in the Default config but with the multi apps approach, I guess they can "forgot" to do it if they fill the credentials at store level. 
The question is, is it still needed now ? Looks like this is not anymore and - I guess - that's ok to let the credentials empty at default config level WDYT ? (cc @mrahman3177 )

Helper/AlgoliaHelper.php
```diff
@@ -119,16 +115,17 @@ protected function addAlgoliaUserAgent(): void
...
         }
 
-        return $this->clients[$this->getStoreId()];
+        return $this->clients[$this->getStoreId()] ?? null;
```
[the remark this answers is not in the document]
damcou: I think you're right, I'll re-add the exception, let's be cautious here.

Helper/AlgoliaHelper.php
damcou: Thoughts after re-adding the exception, I think we need to require the credentials at default config if we want to make sure this method can't return null 🤔

Helper/AlgoliaHelper.php
damcou: Added a boolean to prevent that

Helper/AlgoliaHelper.php
cammonro: I hear ya - we are still using `\Algolia\AlgoliaSearch\Service\AlgoliaCredentialsManager::checkCredentials` which definitely has value - but this particular call impacts deeper in the code and I think the error message display should happen higher up the stack - for instance when we validate in `\Algolia\AlgoliaSearch\Model\Indexer\Category::execute`.

cammonro (approved):
Looks good!

PR #1619, opened 2024-09-20
Including FPT on indexed product final price!

aissyass:
**Summary**

Tested on Magento **2.4.5-p7**.

The FPT configured on the products are not included in the final product price! If we configure the FPT at the level of:
image

And we create a new attribute of type 'Fixed Product Tax' and assign it to an Attribute Set:
image

Then we configure the product price to 59 MAD.
image

And the attribute we created has a tax of 10 MAD.
image

The product is displayed on the product list and on the product page as:
image
image
 

**Actual Result**
The price indexed on Algolia is 59 MAD instead of the customer purchase price of 69 MAD.

**Expected Result**
The price indexed on Algolia includes the FPTs (which is 69 MAD in this case).

### DIFF
--- a/Helper/Entity/Product/PriceManager/ProductWithChildren.php
+++ b/Helper/Entity/Product/PriceManager/ProductWithChildren.php
@@ -55,8 +55,8 @@ abstract class ProductWithChildren extends ProductWithoutChildren
                 } else {
                     $minPrice = $specialPrice[0];
                 }
-                $price     = $minPrice ?? $this->getTaxPrice($product, $subProduct->getFinalPrice(), $withTax);
-                $basePrice = $this->getTaxPrice($product, $subProduct->getPrice(), $withTax);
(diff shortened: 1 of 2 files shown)


aissyass:
@damcou @cammonro @mrahman3177 
Quick reminder about PR #[PR number]. Thanks for taking a look!"

cammonro:
Hi @aissyass ! 👋 

Sorry for the delay in response. You are right! FPT is not a use case we currently handle but we certainly should! 

We will add this to our backlog for inclusion in a future release. 

Thank you for this contribution. 💪

damcou:
Hi @aissyass , thanks a lot for your contribution ! 
Based on your suggestion, we introduced FPT compatibility which should be shipped in our upcoming v3.17.0 : https://github.com/algolia/algoliasearch-magento-2/pull/1822

Closing this PR.
Thanks again !

PR #1620, opened 2024-09-23
MAGE-1055: Fix current integration tests

damcou:
This PR contains:
- Removal of Magento 2.3 class
- Splitting Magento 2.4 class into Magento 2.4.6 and Magento 2.4.7 classes to test against the two older versions
- Fix these values for Magento 2.4
- Fix some config paths which weren't up-to-date (facets and sortings)
- Adapt some of the test class code to prevent flakiness due to `waitForTask` logic which depends on index name.

On fresh install (with `TESTS_CLEANUP` from phpunit.xml turned to "enabled"):
image

The 3 incomplete tests belong to the `QueueTest` class which will be updated in a next task.


Test/Integration/ConfigTest.php
```diff
@@ -236,12 +230,9 @@ public function testExtraSettings()
 
         foreach ($sections as $section) {
             $this->setConfig('algoliasearch_extra_settings/extra_settings/' . $section . '_extra_settings', '{"exactOnSingleWordQuery":"word"}');
-        }
-
-        $indicesConfigurator->saveConfigurationToAlgolia(1);
-        $this->algoliaHelper->waitLastTask();
+            $indicesConfigurator->saveConfigurationToAlgolia(1);
```
cammonro, line 233: This seems like a heavy operation to call within the section loop?

Test/Integration/ConfigTest.php
```diff
@@ -218,15 +218,9 @@ public function testExtraSettings()
         $sections = ['products', 'categories', 'pages', 'suggestions'];
 
         foreach ($sections as $section) {
+            $this->algoliaHelper->waitLastTask();
```
cammonro, line 221: On the first iteration of this loop it seems to me that we would call two wait operations in [immediate succession](https://github.com/algolia/algoliasearch-magento-2/pull/1620/files#diff-e50009685b9036004740bf0f93d4afd3bff470989f1fd0f93a7886fa0bb429e0R216)? Should this perhaps be located towards the end of the loop?

Test/Integration/TestCase.php
```diff
@@ -97,10 +97,10 @@ private function bootstrap()
             return;
         }
 
-        if (version_compare($this->getMagentoVersion(), '2.4.0', '<')) {
-            $this->assertValues = new Magento23();
+        if (version_compare($this->getMagentoVersion(), '2.4.7', '<')) {
```
cammonro, line 100: Small observation, but I'm wondering if we should be more precise with our version checks? e.g. 
```suggestion
        if (version_compare($this->getMagentoVersion(), '2.4.6', '>=')  && version_compare($this->getMagentoVersion(), '2.4.7', '<')) {
```
And skip the test if not matching? We have a separate ticket for this so we can circle back later but I was interested in your thoughts?

cammonro (commented):
Thanks for making these needed changes! Looking good so far but just had a few questions.

Test/Integration/ConfigTest.php
damcou: I understand this makes you react a little bit.
The way it was implemented before doesn't work consistently, I think this is because how the `waitLastTask()` works (by default the waiting is based on the last indexname and the last task id). 
The fact is, the `saveConfigurationToAlgolia()` method updates all the indices (products, categories, pages, suggestions and additional sections) and as we are looping just after, we are not sure that the first checked section index has been successfully updated since the waitLastTask() was waiting for another index (not sure if I'm clear).

BUT, it seems that we have the same problem with the way I implemented it... I think it just works because we let more time to all indices to be updated because it adds more requests ...

Test/Integration/TestCase.php
damcou: Yes totally ! Good idea :) 
I guess this forces us to update the tests after each Magento minor version release, but it's probably not a bad thing after all.

Test/Integration/ConfigTest.php
damcou: I decided to switch back to the old implem and add waiting for all indices in the saveConfigurationToAlgolia method, and it seems to work now.

This shouldn't change anything in the method except for that, WDYT ?

Test/Integration/ConfigTest.php
damcou: Fixed

Test/Integration/ConfigTest.php
cammonro: Yup! That change makes sense. Thanks for explaining the issue!

Test/Integration/TestCase.php
cammonro: Cool! I agree - every minor release we should be reevaluating our tests vs EOL Magento support. We'll address this on a later ticket. 👍

PR #1621, opened 2024-09-24
Bugfix/mage 1066 null prefix

cammonro:
A missing prefix can throw errors with strong types added to `ConfigHelper`. 

This PR ensures that the right type is returned and provides unit tests to validate expected behavior.

### DIFF
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -1242,7 +1242,7 @@ class ConfigHelper
      */
     public function getIndexPrefix(int $storeId = null): string
     {
-        return $this->configInterface->getValue(self::INDEX_PREFIX, ScopeInterface::SCOPE_STORE, $storeId);
+        return (string) $this->configInterface->getValue(self::INDEX_PREFIX, ScopeInterface::SCOPE_STORE, $storeId);
     }
 
     /**
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -446,7 +446,7 @@ class ConfigHelper
     }
 
     protected function serialize(array $value): string {
-        return $this->serializer->serialize($value);
+        return $this->serializer->serialize($value) ?: '';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


damcou (approved):
LGTM 🙌 

I guess what we could do as a starting point for unit tests later is to clear all configurations and see if it doesn't generate any issues 🤔

damcou:
@cammonro Out of curiosity, why that change related to the Magento serializer method was needed since it uses the classic serialize() method under the hood and it's already supposed to return a string ?

cammonro:
Thanks @damcou for weighing in... 

> @cammonro Out of curiosity, why that change related to the Magento serializer method was needed since it uses the classic serialize() method under the hood and it's already supposed to return a string ?

Well while identifying the cause of the unexpected null type on the prefix it made be realize we should be sure of _all_ return types and while unlikely it _is_ possible that the serializer object could also return `false` if it fails for whatever reason (although it may depend on what is used under the hood - for instance [`json_encode`](https://www.php.net/manual/en/function.json-encode.php))

The unit test is to make sure the result is what is expected in this instance. 

So I was aiming to be proactive there even though not directly related to the other bug.

cammonro:
> > Thanks @damcou for weighing in...
> > > @cammonro Out of curiosity, why that change related to the Magento serializer method was needed since it uses the classic serialize() method under the hood and it's already supposed to return a string ?
> > 
> > 
> > Well while identifying the cause of the unexpected null type on the prefix it made be realize we should be sure of _all_ return types and while unlikely it _is_ possible that the serializer object could also return `false` if it fails for whatever reason (although it may depend on what is used under the hood - for instance [`json_encode`](https://www.php.net/manual/en/function.json-encode.php))
> > The unit test is to make sure the result is what is expected in this instance.
> > So I was aiming to be proactive there even though not directly related to the other bug.
> 
> Ok makes sense, thanks ! I think I was looking at an implementation of the Interface you shared that only returned the PHP serialize() method, sorry for that, you're right, there are a bunch of implementations that can return false as well ! 🙌

All good! It's important to ask questions!

PR #1623, opened 2024-09-27
Mage-1056:  Add MultiStore category tests

damcou:
- Added MultiStores related tests for Category entity
- Refactored the MultiStoreTest.php class
- Reorganized the Integration testing folder structure

### DIFF
--- a/Test/Integration/TestCase.php
+++ b/Test/Integration/TestCase.php
@@ -8,8 +8,8 @@ use Algolia\AlgoliaSearch\Helper\ConfigHelper;
 use Algolia\AlgoliaSearch\Setup\Patch\Schema\ConfigPatch;
 use Algolia\AlgoliaSearch\Test\Integration\AssertValues\Magento246;
 use Algolia\AlgoliaSearch\Test\Integration\AssertValues\Magento247;
-use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\App\ProductMetadataInterface;
+use Magento\Framework\ObjectManagerInterface;
 use Magento\Store\Model\ScopeInterface;
 use Magento\TestFramework\Helper\Bootstrap;
 
@@ -21,6 +21,11 @@ if (class_exists('PHPUnit\Framework\TestCase')) {
 
 abstract class TestCase extends \TC
 {
+    /**
+     * @var ObjectManagerInterface
+     */
+    protected $objectManager;
+
     /** @var bool */
     private $boostrapped = false;
 
@@ -99,6 +104,8 @@ abstract class TestCase extends \TC
         if ($this->boostrapped === true) {
             return;
         }
+
+        $this->objectManager = Bootstrap::getObjectManager();
 
         if (version_compare($this->getMagentoVersion(), '2.4.7', '<')) {
             $this->assertValues = new Magento246();
--- a/Test/Integration/MultiStoreCategoriesTest.php
+++ b/Test/Integration/MultiStoreCategoriesTest.php
@@ -82,18 +82,16 @@ class MultiStoreCategoriesTest extends MultiStoreTestCase
         $this->categoriesIndexer->execute([self::BAGS_CATEGORY_ID]);
         $this->algoliaHelper->waitLastTask();
 
-        $this->assertAlgoliaRecordValue(
+        $this->assertAlgoliaRecordValues(
             $this->indexPrefix . 'default_categories',
-            (string)self::BAGS_CATEGORY_ID,
-            'name',
-            self::BAGS_CATEGORY_NAME
+            (string) self::BAGS_CATEGORY_ID,
+            ['name' => self::BAGS_CATEGORY_NAME]
         );
 
-        $this->assertAlgoliaRecordValue(
+        $this->assertAlgoliaRecordValues(
             $this->indexPrefix . 'fixture_second_store_categories',
-            (string)self::BAGS_CATEGORY_ID,
-            'name',
-            self::BAGS_CATEGORY_NAME_ALT
+            (string) self::BAGS_CATEGORY_ID,
+            ['name' => self::BAGS_CATEGORY_NAME_ALT]
         );
     }
 
--- a/Test/Integration/MultiStoreCategoriesTest.php
+++ b/Test/Integration/MultiStoreCategoriesTest.php
@@ -10,7 +10,6 @@ use Magento\Catalog\Api\Data\CategoryInterface;
 use Magento\Catalog\Model\ResourceModel\Category\CollectionFactory;
 use Magento\Framework\Exception\CouldNotSaveException;
 use Magento\Framework\Exception\NoSuchEntityException;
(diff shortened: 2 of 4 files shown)


Test/Integration/AssertValues/Magento247.php
```diff
@@ -4,7 +4,7 @@
 
 class Magento247
 {
-    public $productsOnStockCount = 182;
+    public $productsOnStockCount = 180;
     public $productsOutOfStockCount = 183;
```
cammonro, line 8: I've observed that this is also the expected value on Magento 2.4.6.

Test/Integration/AssertValues/Magento247.php
```diff
@@ -4,7 +4,7 @@
...
     public $lastJobDataSize = 13;
     public $expectedCategory = 17;
```
cammonro, line 10: I've observed that this is also the expected value on Magento 2.4.6.

Test/Integration/Category/MultiStoreCategoriesTest.php
```diff
@@ -0,0 +1,171 @@
...
+     */
+    public function testMultiStoreCategoryIndices()
+    {
+        // Check that every store has the right number of categories
+        foreach ($this->storeManager->getStores() as $store) {
+            $this->assertNbOfRecordsPerStore($store->getCode(), 'categories', $this->assertValues->expectedCategory);
+        }
+
+        $defaultStore = $this->storeRepository->get('default');
+        $fixtureSecondStore = $this->storeRepository->get('fixture_second_store');
```
cammonro, line 62: Very cool use of fixtures!

Test/Integration/Category/MultiStoreCategoriesTest.php
```diff
@@ -0,0 +1,171 @@
...
+    {
+        $defaultStore = $this->storeRepository->get('default');
+
+        // Restore category name in case DB is not cleaned up
+        $this->updateCategory(
+            self::BAGS_CATEGORY_ID,
+            $defaultStore->getId(),
+            [
+                'name' => self::BAGS_CATEGORY_NAME,
+                'is_active' => 1
```
cammonro, line 165: Very thorough. Nice!

Test/Integration/Config/MultiStoreConfigTest.php
```diff
@@ -0,0 +1,89 @@
...
+        $defaultStore = $this->storeRepository->get('default');
+        $fixtureSecondStore = $this->storeRepository->get('fixture_second_store');
+
+        // Change category configuration at store level (attributes and ranking)
+        $attributesFromConfig = $this->configHelper->getCategoryAdditionalAttributes($defaultStore->getId());
+        $attributesFromConfigAlt = $attributesFromConfig;
+        $attributesFromConfigAlt[] = [
+            "attribute" => self::ADDITIONAL_ATTRIBUTE,
+            "searchable" => "1",
+            "order" => "unordered",
```
cammonro, line 53: Solid variation test.

Test/Integration/Product/ProductsIndexingTest.php
```diff
@@ -1,24 +1,50 @@
...
+        $this->stockRegistry = $this->objectManager->get(StockRegistry::class);
+        $this->indexerRegistry = $this->objectManager->get(IndexerRegistry::class);
+
+        $priceIndexer = $this->indexerRegistry->get('catalog_product_price');
+        $priceIndexer->reindexAll();
+    }
+
     public function testOnlyOnStockProducts()
     {
+        $this->assertEquals(true, true);
```
cammonro, line 41: Is this just debugging code?

Test/Integration/Product/ProductsIndexingTest.php
```diff
@@ -27,17 +53,60 @@ public function testIncludingOutOfStock()
...
+    {
+        $empty = $this->getSerializer()->serialize([]);
+
+        $this->setConfig('algoliasearch_products/products/product_additional_attributes', $empty);
+        $this->setConfig('algoliasearch_instant/instant_facets/facets', $empty);
+        $this->setConfig('algoliasearch_instant/instant_sorts/sorts', $empty);
+        $this->setConfig('algoliasearch_products/products/custom_ranking_product_attributes', $empty);
+
         /** @var Product $indexer */
         $indexer = $this->getObjectManager()->create(Product::class);
```
cammonro, line 69: Did we want to use the new member variable here `$this->productsIndexer` or do we want to create a new instance?

Test/Integration/TestCase.php
```diff
@@ -44,6 +49,8 @@ public function setUp(): void
     public function tearDown(): void
     {
         $this->clearIndices();
+        $this->algoliaHelper->waitLastTask();
+        $this->clearIndices(); // Remaining replicas
```
cammonro, line 53: Could lingering replicas be the reason that some indices were not getting cleaned up earlier?

cammonro (approved):
This looks terrific @damcou . Really nice work! 

I really like the new organization and the use of fixtures. 

Just had a few questions but nothing to block merging. 

I have observed that Magento 2.4.6 seems to require the same assertion values as 2.4.7 but let me know if you are seeing something different.

Test/Integration/AssertValues/Magento247.php
damcou: Yes, this is due to the fact that Giftcards are not indexed when they are out of stock. I'm pretty sure that they were at some point, I guess this will require some investigation in MAGE-1076 once we have time.

Test/Integration/Product/ProductsIndexingTest.php
damcou: Ah yes, forgot to remove that indeed, I will do it in the next PR ;)

Test/Integration/Product/ProductsIndexingTest.php
damcou: Yes totally, I'm replacing all the instances, this one will be part of it... in the next PR as well :)

Test/Integration/TestCase.php
damcou: That's what I got in my testing app, with the second clearIndices(), this seems to solve the issue yes (at least when the testing process comes to an end).
I'll try to keep an eye on this.

Test/Integration/AssertValues/Magento247.php
cammonro: Hmm - yeah that's a weird one. Still when I ran the same tests against 2.4.6, I was getting failures on these assertion values. I would expect consistent outcomes on the tests since we set up the test case.

mrahman3177: Just an idea, how about creating a Magento24 class and then both Magento246 and Magento247 extending from Magento24? In that way, we can keep common stuffs on super class.

damcou: Makes sense ! TBH I'm not even sure that there are siginificant changes between minor release, I didn't test the values on 2.4.6 but they must be the same.

PR #1625, opened 2024-10-07
MAGE-1081 Pull fresh settings data if Algolia state changes over life…

cammonro:
While performing integration testing I discovered a bug in the way the `ReplicaManager` evaluates change state as a long-lived singleton.

If the state in Algolia changes over the life of the Magento singleton then its cached replica setting needs to be refreshed as well. 

This settings data is used to determine whether or not custom ranking logic should be applied. (If nothing appears to have changed in Algolia then it does not attempt to update the replica index with a new ranking.)

What is considered a change is based on whether:

- A new replica was added
- A replica was switched from standard to virtual
- A replica was switched from virtual to standard

Retrieving a fresh value from the settings API endpoint to make this comparison should resolve this issue.

### DIFF
--- a/Service/Product/ReplicaManager.php
+++ b/Service/Product/ReplicaManager.php
@@ -133,12 +133,14 @@ class ReplicaManager implements ReplicaManagerInterface
      * relevant to the Magento integration
      *
      * @param string $primaryIndexName
+     * @param bool $refreshCache
      * @return string[] Array of replica index names
      * @throws LocalizedException
+     * @throws NoSuchEntityException
      */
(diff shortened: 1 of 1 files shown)


cammonro:
> LGTM, I guess it doesn't hurt to refresh the data more often in any case, I don't think the process could "spam" the application to get this information, or maybe there's a use-case that I don't think of ?

Thanks for taking a look! The underlying function is memoized because it's called more than once - so it's a small optimization - but caching can always have unintended side effects. 

Adding our integration tests increases the confidence level in the approach here and will enable us to tweak this safely as needed.

cammonro:
Test result without patch:
image

Test result with patch:
image

PR #1627, opened 2024-10-08
Feature/mage 1044 replica testing

cammonro:
This is a work in progress as I have many more use cases I'm working on but this provides a foundation for further replica testing. 

It is used to validate the bug fix needed on https://github.com/algolia/algoliasearch-magento-2/pull/1625/

I also introduced some reflection utilities for mocking and testing protected methods along with config assertions for working with `@magentoDbIsolation disabled`.

image

### DIFF
--- a/Console/Command/ReplicaSyncCommand.php
+++ b/Console/Command/ReplicaSyncCommand.php
@@ -11,7 +11,7 @@ use Algolia\AlgoliaSearch\Exceptions\BadRequestException;
 use Algolia\AlgoliaSearch\Exceptions\ExceededRetriesException;
 use Algolia\AlgoliaSearch\Helper\Entity\ProductHelper;
 use Algolia\AlgoliaSearch\Service\StoreNameFetcher;
-use Magento\Framework\App\State;
+use Magento\Framework\App\State as AppState;
 use Magento\Framework\Console\Cli;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Exception\NoSuchEntityException;
@@ -27,12 +27,12 @@ class ReplicaSyncCommand extends AbstractReplicaCommand implements ReplicaSyncCo
         protected ReplicaManagerInterface $replicaManager,
         protected ProductHelper           $productHelper,
         protected StoreManagerInterface   $storeManager,
-        State                             $state,
+        AppState                          $appState,
         StoreNameFetcher                  $storeNameFetcher,
         ?string                           $name = null
     )
     {
-        parent::__construct($state, $storeNameFetcher, $name);
+        parent::__construct($appState, $storeNameFetcher, $name);
     }
 
     protected function getReplicaCommandName(): string
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -1157,7 +1157,7 @@ class ConfigHelper
      * @param int|null $scopeId
      * @return void
      */
-    public function setSorting(array $sorting, ?string $scope = null, ?int $scopeId = null): void
+    public function setSorting(array $sorting, string $scope = Magento\Framework\App\Config\ScopeConfigInterface::SCOPE_TYPE_DEFAULT, ?int $scopeId = null): void
     {
         $this->configWriter->save(
             self::SORTING_INDICES,
--- a/Test/Integration/Product/ReplicaIndexingTest.php
+++ b/Test/Integration/Product/ReplicaIndexingTest.php
@@ -17,8 +17,6 @@ class ReplicaIndexingTest extends IndexingTestCase
 
     protected ?string $indexSuffix = null;
 
-    protected ?array $ogSortingState = null;
-
     public function setUp(): void
     {
         parent::setUp();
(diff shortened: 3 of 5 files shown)


Test/Integration/Product/ReplicaIndexingTest.php
```diff
@@ -0,0 +1,333 @@
...
+            'attribute'      => $sortAttr,
+            'sort'           => $sortDir,
+            'sortLabel'      => $sortAttr,
+            'virtualReplica' => 1
+        ];
+        $this->configHelper->setSorting($sorting);
+
+        $this->assertConfigInDb(ConfigHelper::SORTING_INDICES, json_encode($sorting));
+
+        $this->refreshConfigFromDb();
```
damcou, line 103: It's not clear to me why this needs to be run here, curious about it

Test/Integration/Product/ReplicaIndexingTest.php
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    /**
+     * @depends testReplicaSync
+     * @magentoConfigFixture current_store algoliasearch_instant/instant/is_instant_enabled 1
+     * @throws AlgoliaException
+     * @throws ExceededRetriesException
+     * @throws \ReflectionException
+     */
+    public function testReplicaRebuild(): void
```
damcou, line 167: Curious again, why is this mock and reflection implementation needed here ? Is it only for protected methods ?

Test/Integration/Product/ReplicaIndexingTest.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        $this->assertTrue($this->hasSortingAttribute($sortAttr, $sortDir));
+    }
+
+    protected function assertNoSortingAttribute($sortAttr, $sortDir): void
+    {
+        $this->assertFalse($this->hasSortingAttribute($sortAttr, $sortDir));
+    }
+
+    public function tearDown(): void
```
damcou, line 328: Can be removed ?

damcou (approved):
Tested on local env, got the green light :) 

I have small additional non-blocking questions related to the implementation though. 
I see in the PR that you have other test cases to add, maybe we could keep it for later ? This seems already pretty thorough no ? 

Note to ourselves : make sure to activate the virtual replicas in the testing applications once we re-add the tests in the CI.

Console/Command/ReplicaSyncCommand.php
```diff
@@ -27,12 +27,12 @@ public function __construct(
         protected ReplicaManagerInterface $replicaManager,
         protected ProductHelper           $productHelper,
         protected StoreManagerInterface   $storeManager,
-        State                             $state,
+        AppState                          $appState,
```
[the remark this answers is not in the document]
cammonro: This was largely a cosmetic change - but while experimenting with ReplicaState in the class I realized that it could get confused with AppState as they are two different things. I left the alias in case I decided to use both states in the future. If you think it's best to leave as it was originally I'll remove the change.

Helper/ConfigHelper.php
```diff
@@ -1157,7 +1157,7 @@ public function getRawSortingValue(?int $storeId = null): string
      * @param int|null $scopeId
      * @return void
      */
-    public function setSorting(array $sorting, ?string $scope = null, ?int $scopeId = null): void
+    public function setSorting(array $sorting, string $scope = Magento\Framework\App\Config\ScopeConfigInterface::SCOPE_TYPE_DEFAULT, ?int $scopeId = null): void
```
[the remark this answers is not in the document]
cammonro: You're right to question this. It was failing my tests when I failed to provided the argument. All other invocations in the current code based currently call the method with an argument so it seemed safe to modify.

Test/Integration/Product/ReplicaIndexingTest.php
cammonro: Yup just for protected methods - I wanted to test the command class in its entirety but `execute` is protected. I also needed to mock the Input/OutputInterface.

Test/Integration/Product/ReplicaIndexingTest.php
cammonro: This was a placeholder for possible future refactoring / use cases but I still wanted to clean up replicas. Yes otherwise it can be removed!

Test/Integration/Product/ReplicaIndexingTest.php
cammonro: This was needed because while testing the writes to the database I discovered that the in-memory config cache for duration of the PHPUnit test function is not updated so calls to ScopeConfigInterface were providing stale data. I can refresh it with `ReinitableConfigInterface` but then it wipes out the earlier bootstrapping which configures via environment variables and also initializes the index name - a reference to which must be preserved before the test can finish.

cammonro:
> Tested on local env, got the green light :)

Green light = good feels! ☺️ 
 
> I have small additional non-blocking questions related to the implementation though. I see in the PR that you have other test cases to add, maybe we could keep it for later ? This seems already pretty thorough no ?
> 
> Note to ourselves : make sure to activate the virtual replicas in the testing applications once we re-add the tests in the CI.

You raise a good question. I was debating this also. I'd like to move on to other tests but I still want to verify against multi store and customer group fixtures. But perhaps those should be new tickets as that increases scope ... let's discuss later today. Thank you for raising this!

PR #1629, opened 2024-10-10
MAGE-1058: Add Multistores Pages tests

damcou:
This PR contains:
- Added new tests related to CMS Pages (multi-stores scenarios)
- Fixed a bug where excluded pages set at store level in the configuration weren't taken into account.

### DIFF
--- a/Test/Integration/Page/MultiStorePagesTest.php
+++ b/Test/Integration/Page/MultiStorePagesTest.php
@@ -1,0 +1,148 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Test\Integration\Page;
+
+use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
+use Algolia\AlgoliaSearch\Exceptions\ExceededRetriesException;
+use Algolia\AlgoliaSearch\Helper\ConfigHelper;
+use Algolia\AlgoliaSearch\Test\Integration\MultiStoreTestCase;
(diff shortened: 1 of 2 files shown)


Test/Integration/Page/MultiStorePagesTest.php
```diff
@@ -0,0 +1,148 @@
...
+     * @throws NoSuchEntityException
+     */
+    protected function setupStore(StoreInterface $store, bool $enableInstantSearch = false): void
+    {
+        // Exclude 2 pages on second store
+        $excludedPages = $store->getCode() === 'fixture_second_store' ?
+            [['attribute' => 'no-route'], ['attribute' => 'home']]:
+            [];
+
+        $this->setConfig(
```
cammonro, line 131: Do we need to tear this down? Should we have a `tearDownStore` method I wonder...? 🤔

cammonro (approved):
Looking good to me. Just had a question about clean up with @magentoDbIsolation disabled.

Test/Integration/Page/MultiStorePagesTest.php
mrahman3177: tearDown() triggers. At the end of the test all indices destroys

damcou: No need to tear it down here I think since the 2 fixture stores are removed by the rollback script of the Magento fixtures.

But having  tearDownStore method on the parent class of the multi stores tests could make sense indeed :)

PR #1633, opened 2024-10-24
MAGE-1093: Fix NeuralSearch compatibility

damcou:
This PR contains:
- Removed `mode` during `setSettings` if index has NeuralSearch enabled to prevent crash during full reindexing with temporary index creation.


cammonro (approved):
👍 This is a solid workaround - let's examine using scopes on the move operation in a future task.

PR #1634, opened 2024-10-29
MAGE-1032: Make Merchandising tools compatible with Multi-Application IDs

damcou:
This PR contains:
- Query Rules can now be created on multiple application using the Store View dropdown
- Category Merchandiser made compatible
- Query Merchandiser made compatible
- Landing Page Builder made compatible

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -927,11 +927,15 @@ class Data
     {
         $indexNames = [];
         $indexNames[0] = [
+            'appId' => $this->configHelper->getApplicationID(),
+            'apiKey' => $this->configHelper->getAPIKey(),
             'indexName' => $this->getBaseIndexName(),
             'priceKey'  => '.' . $this->configHelper->getCurrencyCode() . '.default',
         ];
         foreach ($this->storeManager->getStores() as $store) {
             $indexNames[$store->getId()] = [
+                'appId' => $this->configHelper->getApplicationID($store->getId()),
+                'apiKey' => $this->configHelper->getAPIKey($store->getId()),
                 'indexName' => $this->getBaseIndexName($store->getId()),
                 'priceKey' => '.' . $store->getCurrentCurrencyCode($store->getId()) . '.default',
             ];
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -925,19 +925,29 @@ class Data
      */
     public function getIndexDataByStoreIds(): array
     {
+
         $indexNames = [];
-        $indexNames[0] = [
+        $indexNames[AlgoliaHelper::ALGOLIA_DEFAULT_SCOPE] = [
             'appId' => $this->configHelper->getApplicationID(),
             'apiKey' => $this->configHelper->getAPIKey(),
             'indexName' => $this->getBaseIndexName(),
             'priceKey'  => '.' . $this->configHelper->getCurrencyCode() . '.default',
+            'facets' => $this->configHelper->getFacets(),
+            'currencyCode' => $this->configHelper->getCurrencyCode(),
+            'maxValuesPerFacet' => (int) $this->configHelper->getMaxValuesPerFacet(),
+            'categorySeparator' => $this->configHelper->getCategorySeparator(),
         ];
         foreach ($this->storeManager->getStores() as $store) {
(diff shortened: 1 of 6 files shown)


Helper/Data.php
```diff
@@ -925,15 +925,29 @@ public function getBaseIndexName(int $storeId = null): string
...
+            'facets' => $this->configHelper->getFacets(),
+            'currencyCode' => $this->configHelper->getCurrencyCode(),
+            'maxValuesPerFacet' => (int) $this->configHelper->getMaxValuesPerFacet(),
+            'categorySeparator' => $this->configHelper->getCategorySeparator(),
         ];
         foreach ($this->storeManager->getStores() as $store) {
-            $indexNames[$store->getId()] = [
-                'indexName' => $this->getBaseIndexName($store->getId()),
-                'priceKey' => '.' . $store->getCurrentCurrencyCode($store->getId()) . '.default',
+            $storeId = $store->getId();
```
cammonro, line 941: Can we reduce this redundancy with a shared builder function?

view/adminhtml/templates/landingpage/search-configuration.phtml
```diff
@@ -301,490 +297,501 @@ $isConfig = [
...
-            limit: config.maxValuesPerFacet,
-            operator: 'and',
-            templates: templates,
-            sortBy: ['count:desc', 'name:asc'],
-            cssClasses: {
-              root: 'facet conjunctive'
-            }
-          };
+        $('select[name="store_id"]').on('change', function(e) {
+            initInstantSearch();
```
cammonro, line 697: Does `initAutocomplete` need to be called as well? How do we ensure that it is pulling from the correct application?

view/adminhtml/web/js/components/query-merchandising.js
```diff
@@ -26,12 +25,12 @@ define(
 					$( document ).ready(function() {
 					    self.initSubscribers();
 					});
-					
+
 				},
 
 				initObservable: function () {
 					this._super().observe(
-						'queryText storeId'
```
cammonro, line 34: I am curious why we sometimes handle events with Knockout observables and  sometimes with jQuery?

cammonro (commented):
Nice work here! Had some questions but looking good to me so far.

view/adminhtml/templates/landingpage/search-configuration.phtml
damcou: `initAutocomplete` is defined and called in `initInstantSearch` :
```
L624-629: 
search.on('render', function() {
      initAutocomplete();
      initSortableTable();
});

search.start();
```
This is how it was designed originally (calling `initAutocomplete` in the `search.on()` method), I'm not sure if it would be considered as best practice on a frontend development point of view though but ... it works ;) .

view/adminhtml/web/js/components/query-merchandising.js
damcou: From what I remember, Query Merchandiser is different from the LPB because there are no collapsible sections in it so managing the events was possible from knockout observables and not for the LPB (because the merchandising section is not deployed by default). 

I had to remove it in this PR because the knockout observable automatically triggered the InstantSearch refresh **before** it was re-initialized with the right application so there was a weird behaviour where 2 calls were made (one with the "old" application (knockout observable) and one from the "new" one (jQuery)), which led to a JS error because the index wasn't found on the "old" one. 

This change makes sure that the InstantSearch refresh is performed by jQuery on the right application and doesn't generate any error.

view/adminhtml/templates/landingpage/search-configuration.phtml
cammonro: Ah... I missed that call there. Works for me. Thanks for explaining!

view/adminhtml/web/js/components/query-merchandising.js
cammonro: Ok that makes sense. Thanks for the added context. Seems like we should just stick with one consistent approach for "grokability" and since jQuery seems to cover both use cases, this looks like a move in the right direction. 🤘

cammonro (approved):
👍 LGTM!

PR #1648, opened 2024-11-26
Mage-1095: Data Helper refactoring

damcou:
This PR includes:
- Full refactoring of `Algolia\AlgoliaSearch\Helper\Data`:
  - All the logic related to index building has been moved to newly created classes called "Index Builders"
  - Kept the methods and mark them as "deprecated", they could theoretically be removed but we can let them until a next minor release.
  - Moved the `getSearchResults()` method to a dedicated `BackendSearch` Service
- Created one `IndexBuilder` file per entity in the `Service` directory
- Improved the Indexing Queue grid 
  - Methods have now the same names, so the "class" column has been added to get a better view of the entities.
  - Better description what the job does (Full indexing or specific IDs indexing - see screenshot)
- Removed all occurrences of the deprecated methods within the codebase.
- Updated QueueTest file.

image


Model/Indexer/AdditionalSection.php
cammonro, line 1: This is legacy but I'm wondering if we should take the opportunity to rename this class? `AdditionalSection` doesn't really tell me a whole lot about what this thing is unless I study the implementation and I'd like to make the code easier to understand for other devs. 

Since it is used to create dynamic indices for product attributes for the Autocomplete UI maybe we should reference it as `AutocompleteSection` or something. I thought about `ProductAttribute` but the builder has some logic related to filtering the regular entity indices which doesn't apply ... unless we made that a facade. What do you think?

Model/Indexer/Category.php
```diff
@@ -79,10 +81,10 @@ private function rebuildAffectedProducts($storeId)
         if ($affectedProductsCount > 0 && $this->configHelper->indexProductOnCategoryProductsUpdate($storeId)) {
             $productsPerPage = $this->configHelper->getNumberOfElementByPage();
             foreach (array_chunk($affectedProducts, $productsPerPage) as $chunk) {
-                /** @uses Data::rebuildStoreProductIndex() */
+                /** @uses ProductIndexBuilder::rebuildIndexIds() */
                 $this->queue->addToQueue(
-                    Data::class,
-                    'rebuildStoreProductIndex',
+                    ProductIndexBuilder::class,
```
cammonro, line 86: So much clearer! 👏

Model/Indexer/Category.php
```diff
@@ -101,10 +103,10 @@ private function rebuildAffectedProducts($storeId)
     private function processSpecificCategories($categoryIds, $categoriesPerPage, $storeId)
     {
         foreach (array_chunk($categoryIds, $categoriesPerPage) as $chunk) {
-            /** @uses Data::rebuildStoreCategoryIndex */
+            /** @uses CategoryIndexBuilder::rebuildIndexIds */
             $this->queue->addToQueue(
-                Data::class,
-                'rebuildStoreCategoryIndex',
+                CategoryIndexBuilder::class,
```
cammonro, line 108: This more clearly calls out the two different indexing ops here in this class. 👍

Model/Source/JobMethods.php
```diff
@@ -7,14 +7,18 @@ class JobMethods implements \Magento\Framework\Data\OptionSourceInterface
...
-        'rebuildProductIndex' => 'Product Reindex',
-        'rebuildStoreAdditionalSectionsIndex' => 'Additional Section Reindex',
-        'rebuildStoreSuggestionIndex' => 'Suggestion Reindex',
-        'rebuildStorePageIndex' => 'Page Reindex',
+        'rebuildIndex' => 'Rebuild Full Index',
+        'rebuildIndexIds' => 'Rebuild Specific IDs',
+        'deleteInactiveProducts' => 'Delete Inactive Products',
+        // @deprecated
+        'deleteObjects' => 'Object deletion (deprecated)',
+        'rebuildStoreCategoryIndex' => 'Category Reindex (deprecated)',
```
cammonro, line 15: I like this approach. This looks like a graceful way to handle this migration to the new methods. Have we tested mixed queues post migration from a previous version (< `3.15`)? Any potential issues you can think of with that?

Model/Source/JobMethods.php
```diff
@@ -7,14 +7,18 @@ class JobMethods implements \Magento\Framework\Data\OptionSourceInterface
...
         'moveIndexWithSetSettings' => 'Move Index',
-        'deleteObjects' => 'Object deletion',
-        'rebuildStoreCategoryIndex' => 'Category Reindex',
-        'rebuildCategoryIndex' => 'Category Reindex',
-        'rebuildStoreProductIndex' => 'Product Reindex',
-        'rebuildProductIndex' => 'Product Reindex',
-        'rebuildStoreAdditionalSectionsIndex' => 'Additional Section Reindex',
-        'rebuildStoreSuggestionIndex' => 'Suggestion Reindex',
-        'rebuildStorePageIndex' => 'Page Reindex',
+        'rebuildIndex' => 'Rebuild Full Index',
```
cammonro, line 10: As these jobs could be a first time build - I'm wondering if we steer away from the "Rebuild" language. Perhaps just "Build full index"?

Model/Source/JobMethods.php
```diff
@@ -7,14 +7,18 @@ class JobMethods implements \Magento\Framework\Data\OptionSourceInterface
...
+        'rebuildIndexIds' => 'Rebuild Specific IDs',
```
cammonro, line 11: Instead of "Rebuild Specific IDs" (too technical sounding to me) what if we go with "Update index"?

Service/AbstractIndexBuilder.php
```diff
@@ -0,0 +1,105 @@
...
+use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
+use Algolia\AlgoliaSearch\Helper\AlgoliaHelper;
+use Algolia\AlgoliaSearch\Helper\ConfigHelper;
+use Algolia\AlgoliaSearch\Logger\DiagnosticsLogger;
+use Magento\Framework\App\Area;
+use Magento\Framework\App\Config\ScopeCodeResolver;
+use Magento\Framework\Exception\NoSuchEntityException;
+use Magento\Store\Model\App\Emulation;
+
+abstract class AbstractIndexBuilder
```
cammonro, line 14: I like this introduction of an abstract class. I also think we should have it implement an `IndexBuilder` interface under the `Api` namespace that contains common build methods. The methods would also be abstract in this class.

Service/Category/IndexBuilder.php
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * @param int $storeId
+     * @param int $page
+     * @param int $pageSize
+     * @return void
+     * @throws LocalizedException
+     * @throws NoSuchEntityException
+     * @throws \Exception
+     */
+    public function rebuildIndex(int $storeId, int $page, int $pageSize): void
```
cammonro, line 40: If we veer away from `rebuild` we could simplify this to `buildIndex` or just `build` (and include it in `Api\IndexBuilder`) ?

Service/Category/IndexBuilder.php
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * @param $storeId
+     * @param $categoryIds
+     * @return void
+     * @throws LocalizedException
+     * @throws NoSuchEntityException
+     * @throws \Exception
+     */
+    public function rebuildIndexIds($storeId, $categoryIds = null): void
```
cammonro, line 60: This method name seems a little confusing to me as we are building entities not indices. Some other ideas `buildEntityIds` or just `buildEntities`? 

To really abstract this we could follow the pattern used by Magento indexers and go with:
- `build`
- `buildFull` (calls `build` with empty ids)
- `buildList` (calls `build` with list of entity ids)

These methods could be part of `Api\IndexBuilder`.

Service/Product/BackendSearch.php
cammonro, line 1: Nice job applying SRP

view/adminhtml/ui_component/algolia_algoliasearch_indexing_queue_listing.xml
```diff
@@ -106,6 +106,17 @@
                 </item>
             </argument>
         </column>
+        <column name="class">
+            <argument name="data" xsi:type="array">
+                <item name="options" xsi:type="object">Algolia\AlgoliaSearch\Model\Source\JobClasses</item>
+                <item name="config" xsi:type="array">
+                    <item name="component" xsi:type="string">Magento_Ui/js/grid/columns/select</item>
```
cammonro, line 113: I think adding this class field as a filter in the listing component could be helpful here.

cammonro (changes requested):
Really GREAT work! 🙌

I added some suggestions. Please let me know what you think. Happy to chat further.

Model/Source/JobMethods.php
damcou: Yes ! I filled the queue with a full reindexing in the queue **before** and **after** applying those changes and the queue is processed as it should be (obviously the new jobs can't be merged with the old one because the process checks for the class and the method name to merge the jobs, but I think this is not a big deal)

Service/AbstractIndexBuilder.php
damcou: I thought about this but as we want to keep full compatibility with the old methods (for now) and those methods all have a different signature, I don't know how to regroup them under the same interface (but maybe there's a trick that I'm not aware of), I saw that people sometimes use an `$args` array for that but we come back to the fact that it's not compatible with the old methods)

Service/Category/IndexBuilder.php
damcou: I prefer `buildIndex` since it's more descriptive

Service/Category/IndexBuilder.php
damcou: Since this method is called for updates, I guess we should keep the `rebuild` naming no ? So I'd go for `rebuildEntityIds` if you prefer.

Model/Indexer/AdditionalSection.php
damcou: Answer in JIRA

Service/AbstractIndexBuilder.php
damcou: More on JIRA to elaborate with this whole abstraction thing (coupled with the other comments)

view/adminhtml/ui_component/algolia_algoliasearch_indexing_queue_listing.xml
damcou: Good catch, I forgot it indeed!

Service/AbstractIndexBuilder.php
cammonro: Got it! Follow up thoughts in Jira as well.

Model/Indexer/AdditionalSection.php
cammonro: Slated for a future task - resolved!

Service/AbstractIndexBuilder.php
cammonro: Slated for follow up

cammonro (approved):
Let 'er rip!

PR #1650, opened 2024-11-27
MAGE-1105: Profiling fixes and output formatting

cammonro:
I found a flaw in the recently introduced profiling feature if the calling code fails to properly start or stop a timer so I added some checks here to flag this. I also needed to simplify the way the data exports into CSV (way too verbose) so methods / labels can be shortened but still unique and informative. 

- Added new `DiagnosticsException` to call out improper profiler invocations
- Added internal tracker to determine profiling depth
- Simplified method references while aggregating under the Algolia execution group

Before:
image

After:
image


Setup/.DS_Store
damcou, line 1: 👀

Logger/DiagnosticsLogger.php
```diff
@@ -83,13 +89,26 @@ public function startProfiling(string $timerName): void
...
     }
 
+    /**
+     * @throws DiagnosticsException
+     */
     public function stopProfiling(string $timerName): void
     {
         if (!$this->isProfilerEnabled) return;
 
+        $lastTimerName = array_pop($this->timerStack); //$this->timerStack[count($this->timerStack) - 1];
```
damcou, line 105: 👀

damcou (commented):
Good job ! Seems way clearer this way 😄 

Just need a little bit of cleanup and you can 🚀

PR #1651, opened 2024-11-28
MAGE-1151: Index Builders improvements

damcou:
This PR contains:
- Added `IndexBuilderInterface` for Pages, Suggestions and Additional Sections
- Added `UpdatableIndexBuilderInterface` for Products and Categories (which have update methods)
- Added standardized approach for all Index Builders with common methods (`buildIndex`, `buildIndexFull`, `buildIndexList` defined by the interfaces)
- Updated Queue merging to be compatible with the new Services
- Improved display of the data column in the Indexing Queue in the admin
- Updated Queue Tests


Api/IndexBuilder/IndexBuilderInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Api\IndexBuilder;
+
+interface IndexBuilderInterface
+{
+    public function buildIndex(int $storeId, ?array $ids, ?array $options): void;
```
cammonro, line 7: @damcou Now that I'm looking at this method signature again I'm feeling like we should clarify the second argument to distinguish it from `$storeId` - maybe something like `$entityIds`? WDYT

Api/IndexBuilder/UpdatableIndexBuilderInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Api\IndexBuilder;
+
+interface UpdatableIndexBuilderInterface extends IndexBuilderInterface
+{
+    public function buildIndexList(int $storeId, ?array $ids, ?array $options): void;
```
cammonro, line 7: Same thought on clarifying `$ids` to `$entityIds` in the method sig.

Ui/Component/Listing/Column/Data.php
```diff
@@ -27,9 +27,18 @@ public function prepareDataSource(array $dataSource)
             if (is_array($data)) {
                 foreach ($data as $key => $value) {
                     if (is_array($value)) {
-                        $value = implode(', ', $value);
+                        $stringValue = '';
+                        foreach ($value as $k => $v) {
```
cammonro, line 31: Great idea! I'm thinking we should split this formatter out into a separate function - and call it recursively for nested values (up to a certain depth limit).

cammonro (changes requested):
This is superb! Love it! 😁 

Just a couple of small suggestions.

cammonro (approved):
Awesome! 🙌

PR #1652, opened 2024-11-29
MAGE-1101: Change `getCookie` method 

damcou:
This PR contains:
- Updated `getCookie` method so it takes into account multiple cookie values (i.e. when Magento add a leading dot to the cookie's domain)

### DIFF
--- a/view/frontend/web/js/internals/common.js
+++ b/view/frontend/web/js/internals/common.js
@@ -98,15 +98,26 @@ define(['jquery', 'algoliaBundle'], function ($, algoliaBundle) {
         return check;
     };
 
-    window.getCookie = function (name) {
-        var value = "; " + document.cookie;
-        var parts = value.split("; " + name + "=");
-        if (parts.length == 2) {
-            return parts.pop().split(";").shift();
+    window.getCookie = function(name) {
+        var cookie,
+            i,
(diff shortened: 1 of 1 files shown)


view/frontend/web/js/internals/common.js
```diff
@@ -98,15 +98,26 @@ define(['jquery', 'algoliaBundle'], function ($, algoliaBundle) {
         return check;
     };
 
-    window.getCookie = function (name) {
-        var value = "; " + document.cookie;
-        var parts = value.split("; " + name + "=");
-        if (parts.length == 2) {
-            return parts.pop().split(";").shift();
+    window.getCookie = function(name) {
+        var cookie,
```
cammonro, line 102: While this works (and mirrors the `Magento_PageCache/js/form-key-provider` implementation) for new code introductions it has been my preference that we start adhering to modern JS including the `let`/`const` convention instead.

cammonro (changes requested):
Added a couple of observations. (Additional thoughts in Jira) 

I also think we need to add a RequireJS dependency to ensure that `Magento_PageCache/js/form-key-provider` is properly loaded beforehand.

cammonro (approved):
LGTM!

PR #1656, opened 2024-12-04
MAGE-1153: fix synonyms duplication after PHP v4 upgrade

damcou:
This PR contains:
- A fix where synonyms were systematically re-added after an index `setSettings()` call. Before v`3.14.x` this was previously handled with the old PHP client v3 with a `getVersion` query parameter set to `2` . We now unset the synonyms attribute since it has to be handled in the Algolia dashboard anyway.

### DIFF
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -241,6 +241,11 @@ class AlgoliaHelper extends AbstractHelper
             $settings = $this->mergeSettings($indexName, $settings, $mergeSettingsFrom);
         }
 
+        // Removing synonyms from the settings to avoid duplication (synonyms must be managed within the algolia dashboard)
+        if (isset($settings['synonyms'])) {
+            unset($settings['synonyms']);
+        }
+
         $res = $this->getClient()->setSettings($indexName, $settings, $forwardToReplicas);
 
         self::setLastOperationInfo($indexName, $res);
--- a/Helper/AlgoliaHelper.php
+++ b/Helper/AlgoliaHelper.php
@@ -241,11 +241,6 @@ class AlgoliaHelper extends AbstractHelper
             $settings = $this->mergeSettings($indexName, $settings, $mergeSettingsFrom);
         }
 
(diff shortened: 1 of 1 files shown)


MendyAWS:
I tried applying this fix to version 3.14.3, but it didn't work for me.

Helper/AlgoliaHelper.php
```diff
@@ -241,6 +241,11 @@ public function setSettings(
             $settings = $this->mergeSettings($indexName, $settings, $mergeSettingsFrom);
         }
 
+        // Removing synonyms from the settings to avoid duplication (synonyms must be managed within the algolia dashboard)
+        if (isset($settings['synonyms'])) {
+            unset($settings['synonyms']);
```
cammonro, line 246: We already have an operation that removes problematic attributes from settings at `\Algolia\AlgoliaSearch\Helper\AlgoliaHelper::mergeSettings`. I'm thinking we should centralize the logic there and perhaps make it a discrete function. WDYT?

cammonro (changes requested):

cammonro:
> I tried applying this fix to version 3.14.3, but it didn't work for me.

@MendyAWS thanks for weighing in. We want to make sure this fix solves all occurrences of the issue. Can you please share details on how you are able to reproduce the problem?

Helper/AlgoliaHelper.php
damcou: Makes sense ! Updating it 👍

cammonro (approved):

cammonro:
@MendyAWS  Very helpful thank you! From your data I can see that additional configurations are impacted here. 

I've spun up an additional PR (https://github.com/algolia/algoliasearch-magento-2/pull/1658) with a proposed change. Please let us know if that fixes your use case.

cammonro:
For 3.14, PR https://github.com/algolia/algoliasearch-magento-2/pull/1659

MendyAWS:
@cammonro Thanks! Confirmed working.

PR #1660, opened 2024-12-06
MAGE-1153 Synonym deduplicator

cammonro:
This is an experimental synonym deduplicator that handles 3 of the 4 possible synonym types using the current PHP API client v4. Enhancement of the API will be necessary to support a more comprehensive solution.

### DIFF
--- a/Console/Command/AbstractStoreCommand.php
+++ b/Console/Command/AbstractStoreCommand.php
@@ -72,10 +72,27 @@ abstract class AbstractStoreCommand extends Command
     /**
      * @param InputInterface $input
      * @return int[]
+     * @throws LocalizedException
      */
     protected function getStoreIds(InputInterface $input): array
     {
-        return (array) $input->getArgument(self::STORE_ARGUMENT);
+        return $this->validateStoreIds((array) $input->getArgument(self::STORE_ARGUMENT));
+    }
+
+    /**
+     * @param array $storeIds
+     * @return int[]
+     * @throws LocalizedException
+     */
+    protected function validateStoreIds(array $storeIds): array
+    {
+        foreach ($storeIds as $storeId) {
+            if (!ctype_digit($storeId) || (int) $storeId < 1) {
+                throw new LocalizedException(__("Store ID argument must be an integer"));
+            }
+        }
+
+        return array_map('intval', $storeIds);
     }
 
     protected function getStoreArgumentDefinition(): InputArgument {
--- a/Console/Command/AbstractStoreCommand.php
+++ b/Console/Command/AbstractStoreCommand.php
@@ -15,7 +15,7 @@ use Symfony\Component\Console\Question\ConfirmationQuestion;
 
 abstract class AbstractStoreCommand extends Command
 {
-    protected const STORE_ARGUMENT = 'store';
+    protected const STORE_ARGUMENT = 'store_id';
 
     protected ?OutputInterface $output = null;
     protected ?InputInterface $input = null;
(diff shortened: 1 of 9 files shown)


cammonro:
Added some robustness here to the utility even though it is currently limited in not being able to work with one way synonyms. Meanwhile, the refactoring work on `AbstractStoreCommand` can be leveraged for future store scoped operations via CLI.

Console/Command/SynonymDeduplicateCommand.php
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    protected function getCommandName(): string
+    {
+        return 'deduplicate';
+    }
+
+    protected function getCommandDescription(): string
+    {
+        return "Identify and remove duplicate synonyms in Algolia";
```
damcou, line 40: Maybe add a list of the 3 types that will be deduplicated (synonyms, placeholders and alternative corrections) and add an early warning that one-way synonyms will be deleted even before the confirmation prompt ?

Console/Command/AbstractStoreCommand.php
```diff
@@ -0,0 +1,128 @@
...
+use Magento\Framework\Exception\NoSuchEntityException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+abstract class AbstractStoreCommand extends Command
+{
+    protected const STORE_ARGUMENT = 'store';
```
damcou, line 18: Having `store` as parameter can be confusing since we are expecting a store id. WDYT about renaming it to `store_id` to be clearer with the expected type ?

Console/Command/AbstractStoreCommand.php
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    /**
+     * @param InputInterface $input
+     * @return int[]
+     */
+    protected function getStoreIds(InputInterface $input): array
+    {
+        return (array) $input->getArgument(self::STORE_ARGUMENT);
```
damcou, line 77: Is it possible to check the type of the input from the command ? When I try to run `bin/magento algolia:synonyms:deduplicate foo` , I get an error.

Console/Command/SynonymDeduplicateCommand.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $this->input = $input;
+        $this->output = $output;
+        $this->setAreaCode();
+
+        if (!$this->confirmDedupeOperation()) {
+            return Cli::RETURN_SUCCESS;
+        }
+
+        $storeIds = $this->getStoreIds($input);
```
damcou, line 66: Is `bin/magento algolia:synonyms:deduplicate 1 2` the standard way to expect an array of integers ? Maybe it would be useful to add an example in the `getStoreArgumentDescription` method ?

damcou (commented):
God job on this useful command ! :) .
Added some thoughts, especially regarding the store id(s) parameter (not sure if there are alternatives here)

Console/Command/SynonymDeduplicateCommand.php
cammonro: Bonne idée!

Console/Command/AbstractStoreCommand.php
cammonro: J'aime ça!

Console/Command/AbstractStoreCommand.php
cammonro: Bien sûr ... bonne prise!

I'm falling back on Symfony errors here otherwise the calling class needs to catch and do something with it:
image

Console/Command/SynonymDeduplicateCommand.php
cammonro: C'est correct et je suis d'accord...
image

cammonro:
@damcou Good feedback - merci beaucoup! 

I've made the requested changes and revamped the help a bit based on your feedback:
image

damcou (approved):
This is perfect 🚀 
PS: You're so french ! 😄

PR #1661, opened 2024-12-10
MAGE-1134: AlgoliaHelper refactoring

damcou:
This PR contains:
- Updated `AlgoliaHelper` by removing most of its logic. The main purpose of this class is now to handle the store context while communicating with the Algolia applications now that Multi-Application IDs has been introduced.
- Added `AlgoliaConnector` service which handles the logic previously handled by `AlgoliaHelper`
- Removed all the occurrences of `AlgoliaHelper::setStoreId() `all across the codebase (which was the main purpose of this refactoring)
- Updated all `AlgoliaHelper` methods which now include an optional `$storeId` parameter which ensure that the correct Algolia application is reached
- Fixed Page Tests for Magento 2.4.6

### DIFF
--- a/Service/AlgoliaConnector.php
+++ b/Service/AlgoliaConnector.php
@@ -17,6 +17,26 @@ use Symfony\Component\Console\Output\ConsoleOutput;
 
 class AlgoliaConnector
 {
+    /**
+     * @var string Case-sensitive object ID key
+     */
+    public const ALGOLIA_API_OBJECT_ID = 'objectID';
+
+    /**
+     * @var string
+     */
+    public const ALGOLIA_API_INDEX_NAME = 'indexName';
+
+    /**
+     * @var string
+     */
+    public const ALGOLIA_API_TASK_ID = 'taskID';
+
+    /**
+     * @var int
+     */
+    public const ALGOLIA_DEFAULT_SCOPE = 0;
+
     /** @var int This value should be configured based on system/full_page_cache/ttl
      *           (which is by default 86400) and/or the configuration block TTL
      */
@@ -33,15 +53,12 @@ class AlgoliaConnector
     /** @var string[] */
     protected array $nonCastableAttributes = ['sku', 'name', 'description', 'query'];
 
-    /** @var int  */
-    protected int $storeId = AlgoliaHelper::ALGOLIA_DEFAULT_SCOPE;
-
     /** @var bool */
     protected bool $userAgentsAdded = false;
 
     protected static ?string $lastUsedIndexName;
 
-    protected static ?int $lastTaskId;
+    protected static ?string $lastTaskId;
 
     public function __construct(
         protected ConfigHelper $config,
@@ -57,12 +74,12 @@ class AlgoliaConnector
     }
 
     /**
+     * @param int $storeId
      * @return void
      * @throws AlgoliaException
      */
-    protected function createClient(): void
+    protected function createClient(int $storeId = self::ALGOLIA_DEFAULT_SCOPE): void
     {
-        $storeId = $this->getStoreId();
         if (!$this->algoliaCredentialsManager->checkCredentials($storeId)) {
             throw new AlgoliaException('Client initialization could not be performed because Algolia credentials were not provided.');
         }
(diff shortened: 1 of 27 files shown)


Service/AlgoliaConnector.php
```diff
@@ -0,0 +1,984 @@
...
+
+    /**
+     * Save objects to index (upserts records)
+     * @param string $indexName
+     * @param array $objects
+     * @param bool $isPartialUpdate
+     * @return void
+     * @throws Exception
+     */
+    public function saveObjects(string $indexName, array $objects, bool $isPartialUpdate = false): void
```
cammonro, line 423: What do you think about this idea?... if we were to make this truly stateless and accept `$storeId` for these functions then we don't need `$indexName`. That can be obtained from something like the `IndexNameFetcher`. 

(Of course if we are working with temp indices then that needs to be extrapolated differently. For this we could pass a `$isTempIndex` boolean or even more flexibly an array of options or even some kind of `IndexOptions` object.)

The reason I care about this is because there is the potential for conflict between `$storeId` and `$indexName` and you could conceivably pass an index name that doesn't belong to the store ID. I know this is a misuse of the API but I like the idea of making this more foolproof. It was also the reason I took index names out of the `ReplicaManager` implementation.

This is also a bit more ambitious so if we can agree on the most practical approach maybe we break that up as a separate task.

cammonro (changes requested):
This is a fantastic improvement! So much cleaner and I love the idea of a discrete `AlgoliaConnector` service class. ❤️  By decorating the old implementation we've removed the burden on the caller for maintaining state so definitely a net positive.

I propose we get a little bolder and use this opportunity to make the API even better rather than be tied down to the old approach especially since we have the benefit of the abstraction from AlgoliaHelper to bridge the API while we transition. 

At the very least I think should mean making the new Connector class stateless with regards to how it handles the `$storeId`. Even with the decorators there is a lot of redudant code that I think can still be simplified. 

But awesome work really! 👏

Service/AlgoliaConnector.php
damcou: See comment on JIRA :)

Test/Integration/Page/MultiStorePagesTest.php
```diff
@@ -80,7 +80,9 @@ public function testMultiStorePageIndices()
         $this->pageRepository->save($homePage);
 
         $this->pagesIndexer->execute([self::HOME_PAGE_ID]);
-        $this->algoliaHelper->waitLastTask();
+
+        $this->algoliaHelper->waitLastTask($defaultStore->getId());
```
cammonro, line 84: We use a lot of empty `waitLastTask()` calls. Should we start scoping those?

Service/AlgoliaConnector.php
```diff
@@ -453,12 +455,20 @@ function ($object) use ($action) {
...
-    protected static function setLastOperationInfo(string $indexName, array $response): void
+    protected static function setLastOperationInfo(string $indexName, array $response, ?int $storeId = null): void
     {
         self::$lastUsedIndexName = $indexName;
-        self::$lastTaskId= $response[self::ALGOLIA_API_TASK_ID] ?? null;
+        self::$lastTaskId = $response[self::ALGOLIA_API_TASK_ID] ?? null;
+
+        if (!is_null($storeId)) {
+            self::$lastTaskInfoByStore[$storeId] = [
+                'indexName' => $indexName,
```
cammonro, line 468: Just a note for future. I understand why this is necessary for this situation but I think this is an example where using the `IndexNameFetcher` might have value to help enforce SRP.

Test/Integration/Page/MultiStorePagesTest.php
damcou: I forgot to scope some of them in the multi-stores tests, so I'll do it, thanks !
I think it's also better to keep them unscoped in the other tests to make sure the original implementation still works for "simple application website".

Service/AlgoliaConnector.php
damcou: Understood, I guess this will be part of the work to do in MAGE-1170 :)

cammonro (approved):

PR #1662, opened 2024-12-11
MAGE-1122 intentional price indexing

cammonro:
This will need rebasing against `release/3.14.4-dev` (to be discussed) but this PR provides a feature flagged auto price indexer to address missing pricing record dependencies at time of index.

Introduces injectable service `MissingPriceIndexHandler`.

Tested successfully against 3 major scenarios: 

1. Valid product record missing from price index BUT price index is not marked as invalid by Magento (product should NOT be removed from Algolia search index)
2. Price index is invalid (Price records should be reverified prior to search indexing)
3. Invalid product record  (out of stock etc) missing from price index (product SHOULD be removed from Algolia search index)

As price indexing is entity ID driven / store agnostic and indexing process is called for each store view, a caching associative array is used to track already reindexed records for a single process to avoid redundant indexing.

### DIFF
--- a/Service/Product/MissingPriceIndexHandler.php
+++ b/Service/Product/MissingPriceIndexHandler.php
@@ -2,12 +2,19 @@
 
 namespace Algolia\AlgoliaSearch\Service\Product;
 
+use Magento\Framework\DB\Select;
+use Magento\Catalog\Model\ResourceModel\Product\Collection as ProductCollection;
 use Magento\Catalog\Model\ResourceModel\Product\CollectionFactory;
 use Magento\Framework\Indexer\IndexerRegistry;
 use Magento\Framework\Indexer\IndexerInterface;
+use Zend_Db_Select;
 
 class MissingPriceIndexHandler
 {
+    public const PRICE_INDEX_TABLE = 'catalog_product_index_price';
+    public const PRICE_INDEX_TABLE_ALIAS = 'price_index';
+    public const MAIN_TABLE_ALIAS = 'e';
+
     protected IndexerInterface $indexer;
     public function __construct(
         protected CollectionFactory $productCollectionFactory,
@@ -18,12 +25,12 @@ class MissingPriceIndexHandler
     }
 
     /**
-     * @param array $productIds
-     * @return int[] Array of product IDs that were reindexed by this repair operation
+     * @param string[]|ProductCollection $products
+     * @return string[] Array of product IDs that were reindexed by this repair operation
      */
-    public function refreshPriceIndex(array $productIds): array
+    public function refreshPriceIndex(array|ProductCollection $products): array
     {
-        $reindexIds = $this->getProductIdsToReindex($productIds);
+        $reindexIds = $this->getProductIdsToReindex($products);
         if (empty($reindexIds)) {
             return [];
         }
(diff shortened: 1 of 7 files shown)


Service/Product/MissingPriceIndexHandler.php
```diff
@@ -0,0 +1,193 @@
...
+
+class MissingPriceIndexHandler
+{
+    public const PRICE_INDEX_TABLE = 'catalog_product_index_price';
+    public const PRICE_INDEX_TABLE_ALIAS = 'price_index';
+    public const MAIN_TABLE_ALIAS = 'e';
+
+    protected array $_indexedProducts = [];
+
+    protected IndexerInterface $indexer;
```
damcou, line 23: This interface looks deprecated but I know we already use it within the codebase, should we start using indexers implementing `ActionInterface` instead ?

Service/Product/MissingPriceIndexHandler.php
```diff
@@ -0,0 +1,193 @@
...
+        if (empty($reindexIds)) {
+            return [];
+        }
+
+        $this->logger->log(__("Pricing records missing or invalid for %1 product(s)", count($reindexIds)));
+        $this->logger->log(__("Reindexing product ID(s): %1", implode(', ', $reindexIds)));
+
+        $this->indexer->reindexList($reindexIds);
+
+        return $reindexIds;
```
damcou, line 51: Is it planned to use those ids at some point ? Doesn't seem to be the case in the `DataHelper`.

Service/Product/MissingPriceIndexHandler.php
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * Analyzes a product collection and determines which (if any) records should have their prices reindexed
+     * @param string[]|ProductCollection $products - either an explicit list of product ids or a product collection
```
damcou, line 56: Is there already a case where an array of product ids is passed instead of a ProductCollection ?

damcou (commented):
Good job ! 👏 
I'm just a little bit confused why most of the logic of `MissingPriceIndexHandler` is about recreating a `ProductCollection` whereas the only place where it's called is in `DataHelper` where a collection is already passed. What is the case where an array of ids is passed ?

Service/Product/MissingPriceIndexHandler.php
cammonro: Possibly - I thought it might be useful for future reporting either via logs or returning to a UI where we report missing product prices - perhaps the manual SKU reindexer. Return values don't always need to be used but they can provide utility later.

Service/Product/MissingPriceIndexHandler.php
cammonro: I wanted to keep it flexible. I had an earlier iteration where I was passing IDs but all jobs seem to converge at a single point namely `\Algolia\AlgoliaSearch\Helper\Data::rebuildStoreProductIndexPage`. There are problems with the implementation of that function and we may want to revisit. Also the extra collection operations (which are necessary to keep this code minimally invasive on the regular operation of the query) increase processing time marginally so their may be opportunity for future optimizations where we don't need to do this.

cammonro:
> Good job ! 👏 I'm just a little bit confused why most of the logic of `MissingPriceIndexHandler` is about recreating a `ProductCollection` whereas the only place where it's called is in `DataHelper` where a collection is already passed. What is the case where an array of ids is passed ?

Thank you for the feedback. This is the raw form of the implementation so there may be more tweaks needed for future iterations. I still need to backport to 3.13 and retool for 3.15 which features profiling where I can gauge potential impact. 3.15 also has refactored classes so I need to make sure there are no issues integrating.

PR #1663, opened 2024-12-19
MAGE-1084: Multistore Replicas tests

damcou:
This PR contains:
- Added tests for Replicas with multi stores/applications context
- Created a `ReplicaAssertionsTrait` class which contains all the methods to be used by the `ReplicaIndexingTest` class AND the newly created `MultiStoreReplicaTest` class.
- Updated some methods of the `ReplicaAssertionsTrait` class to handle store switching
- Added a test for customer groups replicas

### DIFF
--- a/Test/Integration/Product/MultiStoreReplicaTest.php
+++ b/Test/Integration/Product/MultiStoreReplicaTest.php
@@ -61,8 +61,6 @@ class MultiStoreReplicaTest extends MultiStoreTestCase
         // Check replica config for color asc
         $this->checkReplicaIsStandard($defaultStore, self::COLOR_ATTR, self::ASC_DIR);
         $this->checkReplicaIsVirtual($fixtureSecondStore, self::COLOR_ATTR, self::ASC_DIR);
-
-        $this->resetAllSortings();
     }
 
     public function testCustomerGroupsConfig()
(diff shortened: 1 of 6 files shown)


cammonro (commented):
Looking good! But I suspect something may be off with setUp/tearDown? 🤔 

Getting different results depending on execution sequence..

.

Test/Integration/Product/MultiStoreReplicaTest.php
```diff
@@ -63,6 +65,35 @@ public function testMultiStoreReplicaConfig()
...
+        $this->assertEquals(7, $this->countStoreIndices($fixtureThirdStore));
+
+        // 13 indices for second store:
+        // - 1 for categories
+        // - 1 for pages
+        // - 1 for suggestions
+        // - 10 for products (1 main and 9 replicas (2 prices sortings * 4 customer groups + 1 other sorting (created_at))
+        $this->assertEquals(13, $this->countStoreIndices($fixtureSecondStore));
+
+        $this->resetAllSortings();
```
cammonro, line 94: Is this still needed if we are also invoking from `tearDown()`?

cammonro (approved):
Looks good to me! Just a question on the `tearDown()` operation.

Test/Integration/Product/MultiStoreReplicaTest.php
damcou: good catch, this is not needed indeed

PR #1665, opened 2024-12-20
MAGE-1122 Refactor intentional price indexing for 3.15

cammonro:
Refactor of https://github.com/algolia/algoliasearch-magento-2/pull/1662 for 3.15 (with profiling)

Some results:

On an extra large fixture (1M products) processing a delta of a 100 products took an additional 19 ms with virtually no RAM consumption:
image

Across a delta of 1000 products over 5 stores the process took 491 ms:
image

On the same size catalog, where 2000 price index records were found missing across 5 stores (but the price index still appears to be valid), the extra reindexing repair operations (on the fly) took approximately 592 ms and ~2.1 MB of RAM. 

image

Performance impact appears to be super linear but sub quadratic - approx. O(n^1.41) observed - and the upper boundary can be contained by the number of records processed per job. 

Full details in Jira.

### DIFF
--- a/Controller/Adminhtml/Reindex/Save.php
+++ b/Controller/Adminhtml/Reindex/Save.php
@@ -13,6 +13,7 @@ use Algolia\AlgoliaSearch\Service\Product\IndexBuilder as ProductIndexBuilder;
 use Magento\Backend\App\Action\Context;
 use Magento\Catalog\Api\ProductRepositoryInterface;
 use Magento\Framework\Controller\ResultFactory;
+use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Exception\NoSuchEntityException;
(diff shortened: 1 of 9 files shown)


Controller/Adminhtml/Reindex/Save.php
```diff
@@ -199,7 +201,7 @@ protected function checkAndReindex($product, $stores)
             $productIds = [$product->getId()];
             $productIds = array_merge($productIds, $this->productHelper->getParentProductIds($productIds));
 
-            $this->productIndexBuilder->rebuildEntityIds($storeId, $productIds);
```
damcou, line 202: Oopsy, good catch ! Sorry for that !

damcou (approved):
LGTM ! 🚀

PR #1667, opened 2025-01-07
MAGE-1097: Handle add to cart redirect for Insights

damcou:
This PR contains a fix where Insights information wasn't kept on a product page after the user tried to add it to cart on a page powered by instantsearch. 
If this product is for example a configurable one, they are redirected to the PDP to select the options (which is the expected Magento behaviour).

### DIFF
--- a/Plugin/AddToCartRedirectForInsights.php
+++ b/Plugin/AddToCartRedirectForInsights.php
@@ -2,6 +2,7 @@
 
 namespace Algolia\AlgoliaSearch\Plugin;
 
+use Algolia\AlgoliaSearch\Helper\ConfigHelper;
 use Magento\Catalog\Api\ProductRepositoryInterface;
 use Magento\Catalog\Model\Product;
 use Magento\CatalogInventory\Api\StockRegistryInterface;
@@ -27,6 +28,7 @@ class AddToCartRedirectForInsights
         protected Session $checkoutSession,
         protected StockRegistryInterface $stockRegistry,
         protected ManagerInterface $eventManager,
+        protected ConfigHelper $configHelper,
     ) {}
 
     /**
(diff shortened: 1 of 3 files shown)


Plugin/AddToCartRedirectForInsights.php
```diff
@@ -0,0 +1,201 @@
...
+        protected ManagerInterface $eventManager,
+        protected ConfigHelper $configHelper,
+    ) {}
+
+    /**
+     * @param Cart $cartModel
+     * @param int|Product $productInfo
+     * @param array|int|DataObject|null $requestInfo
+     *
+     * @return void
```
cammonro, line 39: Ideally for "before" plugins [we should return `null`](https://developer.adobe.com/commerce/php/development/components/plugins/#before-methods) if we are not modifying the arguments. 

[Since `null` is returned anyway](https://www.php.net/manual/en/functions.returning-values.php) when return is omitted this code should work without issue. However, for best practice we should properly docblock this:

```suggestion
     * @return null
```

Plugin/AddToCartRedirectForInsights.php
```diff
@@ -0,0 +1,201 @@
...
+            try {
+                $result = $product->getTypeInstance()->prepareForCartAdvanced($request, $product);
+            } catch (LocalizedException $e) {
+                $this->checkoutSession->setUseNotice(false);
+                $result = $e->getMessage();
+            }
+
+            // if the result is a string, this mean that the product can't be added to the cart
+            // see Magento\Quote\Model\Quote::addProduct()
+            // Here we need to add the insights information to the redirect
```
cammonro, line 76: These comments really help explain the intent here. It might be beneficial to split this out into more discrete methods with clear names as a "self documenting" approach but otherwise this looks good to me. 👍

cammonro (approved):
Awesome job. Added some minor observations.

PR #1668, opened 2025-01-10
MAGE-1170: Handle Index options with `IndexNameFetcher`

damcou:
This PR contains: 
- Added an `IndexOptions` model implementing `IndexOptionsInterface` which would be the component of every index operation coming to `AlgoliaConnector` in the future
- Updated the deprecated `AlgoliaHelper` class to convert all calls to the new format including `IndexOptions` model
- Updated the `AlgoliaConnector` class so it fetches all index names according to the `InderNameFetcher` service
- Removed all the static methods in `AlgoliaConnector` to be able to use the IndexNameFetcher everywhere, notably in the `setLastOperationInfo` methid

### DIFF
--- a/Api/Data/IndexOptionsInterface.php
+++ b/Api/Data/IndexOptionsInterface.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Api\Data;
+
+interface IndexOptionsInterface
+{
+    const STORE_ID = 'store_id';
+
+    const INDEX_SUFFIX = 'index_suffix';
+
+    const IS_TMP = 'is_tmp';
+
+    const ENFORCED_INDEX_NAME = 'enforced_index_name';
+
+    /**
+     * Get field: store_id
+     *
+     * @return int|null
+     */
+    public function getStoreId(): ?int;
+
+    /**
+     * Get field: index_suffix
+     *
+     * @return string|null
+     */
+    public function getIndexSuffix(): ?string;
+
+    /**
+     * Get field: is_tmp
+     *
+     * @return bool
+     */
+    public function isTmp(): bool;
+
+    /**
+     * Get field: enforced_index_name
+     *
+     * @return string|null
+     */
+    public function getEnforcedIndexName(): ?string;
+}
--- a/Api/Data/IndexOptionsInterface.php
+++ b/Api/Data/IndexOptionsInterface.php
@@ -31,7 +31,7 @@ interface IndexOptionsInterface
      *
      * @return bool
      */
-    public function isTmp(): bool;
+    public function isTemporaryIndex(): bool;
 
     /**
      * Get field: enforced_index_name
--- a/Api/Data/IndexOptionsInterface.php
+++ b/Api/Data/IndexOptionsInterface.php
@@ -10,7 +10,7 @@ interface IndexOptionsInterface
 
     const IS_TMP = 'is_tmp';
 
-    const ENFORCED_INDEX_NAME = 'enforced_index_name';
+    const INDEX_NAME = 'index_name';
 
     /**
      * Get field: store_id
@@ -34,9 +34,17 @@ interface IndexOptionsInterface
     public function isTemporaryIndex(): bool;
 
     /**
-     * Get field: enforced_index_name
+     * Get field: index_name
      *
      * @return string|null
      */
-    public function getEnforcedIndexName(): ?string;
+    public function getIndexName(): ?string;
+
+    /**
+     * Set field: index_name
+     *
+     * @param string $indexName
+     * @return void
+     */
+    public function setIndexName(string $indexName): void;
 }
--- a/Model/IndexOptions.php
+++ b/Model/IndexOptions.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Model;
+
+use Algolia\AlgoliaSearch\Api\Data\IndexOptionsInterface;
+use Magento\Framework\DataObject;
+
+class IndexOptions extends DataObject implements IndexOptionsInterface
+{
+    public function getStoreId(): ?int
+    {
+        return $this->hasData(IndexOptionsInterface::STORE_ID) ?
+            (int) $this->getData(IndexOptionsInterface::STORE_ID) :
+            null;
+    }
+
+    public function getIndexSuffix(): ?string
+    {
+        return $this->hasData(IndexOptionsInterface::INDEX_SUFFIX) ?
+            (string) $this->getData(IndexOptionsInterface::INDEX_SUFFIX) :
+            null;
+    }
+
+    public function isTmp(): bool
+    {
+        return $this->hasData(IndexOptionsInterface::IS_TMP) ?
+            (string) $this->getData(IndexOptionsInterface::IS_TMP) :
+            false;
+    }
+
+    public function getEnforcedIndexName(): ?string
+    {
+        return $this->hasData(IndexOptionsInterface::ENFORCED_INDEX_NAME) ?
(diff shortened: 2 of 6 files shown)


Service/AlgoliaConnector.php
```diff
@@ -473,122 +500,139 @@ protected static function setLastOperationInfo(string $indexName, array $respons
...
-     * @param string $toIndexName
-     * @param int|null $storeId
+     * @param IndexOptionsInterface $fromIndexOptions
+     * @param IndexOptionsInterface $toIndexOptions
      * @return void
      * @throws AlgoliaException
+     * @throws NoSuchEntityException
      */
-    public function copyQueryRules(string $fromIndexName, string $toIndexName, ?int $storeId = null): void
+    public function copyQueryRules(IndexOptionsInterface $fromIndexOptions, IndexOptionsInterface $toIndexOptions): void
```
cammonro, line 592: I'm wondering if we should call out via a doc block that copy operations cannot be performed _across_ two different applications?

Service/AlgoliaConnector.php
```diff
@@ -287,26 +304,29 @@ function ($id) {
...
-     * @param string $toIndexName
-     * @param int|null $storeId
+     * @param IndexOptions $fromIndexOptions
+     * @param IndexOptions $toIndexOptions
      * @return void
      * @throws AlgoliaException
+     * @throws NoSuchEntityException
      */
-    public function moveIndex(string $fromIndexName, string $toIndexName, ?int $storeId = null): void
+    public function moveIndex(IndexOptions $fromIndexOptions, IndexOptions $toIndexOptions): void
```
cammonro, line 317: Same as for copy operations I'm thinking we should indicate that these cannot be performed across application IDs.

Service/AlgoliaConnector.php
```diff
@@ -1010,4 +1054,29 @@ protected function getAlgoliaFiltersArrayWithoutCurrentRefinement($filters, $nee
...
+
+    /**
+     * Determines the index name giving it suffix, store id and if it's temporary or not
+     * If an enforced index name has been set, it's automatically returned
+     *
+     * @param IndexOptionsInterface $indexOptions
+     * @return string|null
+     * @throws NoSuchEntityException
+     */
+    protected function getIndexName(IndexOptionsInterface $indexOptions): ?string
```
cammonro, line 1066: Would it make sense to relocate this `getIndexName` logic into the `IndexOptions` object itself?

Service/AlgoliaConnector.php
```diff
@@ -1010,4 +1054,29 @@ protected function getAlgoliaFiltersArrayWithoutCurrentRefinement($filters, $nee
...
+    {
+        if (!is_null($indexOptions->getEnforcedIndexName())) {
+            return $indexOptions->getEnforcedIndexName();
+        }
+
+        if (is_null($indexOptions->getIndexSuffix())) {
+            return null;
```
cammonro, line 1073: Could this be an exception condition?

cammonro (changes requested):
Really like where you going with this. The `IndexOptions` is a very useful abstraction. 

Added some questions and comments. Please let me know what you think.

Api/Data/IndexOptionsInterface.php
```diff
@@ -0,0 +1,42 @@
...
+     * @return bool
+     */
+    public function isTmp(): bool;
+
+    /**
+     * Get field: enforced_index_name
+     *
+     * @return string|null
+     */
+    public function getEnforcedIndexName(): ?string;
```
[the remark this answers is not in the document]
damcou: This can be useful in the integration tests for some assertions (I had the case when I was working on it), so I think this could be a good idea to keep it even after this version.

Api/Data/IndexOptionsInterface.php
```diff
@@ -0,0 +1,42 @@
...
+}
```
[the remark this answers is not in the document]
damcou: Until now I only had to instantiate `IndexOptions` with an array of data (like in `DataObject`) so I never felt the need for setters, I think we'll see on 3.16 if it would change.

Service/AlgoliaConnector.php
damcou: That's what I was thinking at the beginning, but this would require to inject the `IndexNameFetcher` into `IndexOptions` and I would like to keep its instantiation as easy as it is right now (only with one array parameter for data) . Do you see a way to perform such thing ?

Model/IndexOptions.php
```diff
@@ -0,0 +1,37 @@
...
+        return $this->hasData(IndexOptionsInterface::INDEX_SUFFIX) ?
+            (string) $this->getData(IndexOptionsInterface::INDEX_SUFFIX) :
+            null;
+    }
+
+    public function isTmp(): bool
+    {
+        return $this->hasData(IndexOptionsInterface::IS_TMP) ?
+            (string) $this->getData(IndexOptionsInterface::IS_TMP) :
+            false;
```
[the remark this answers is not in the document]
damcou: Good catch! PHPStorm suggested 

`return $this->hasData(IndexOptionsInterface::IS_TMP) && $this->getData(IndexOptionsInterface::IS_TMP);`

as well, I like this one if you're ok with it (no need to cast here)

Service/AlgoliaConnector.php
cammonro: Ah yes this is a good thing to think about... 🤔 

You know... Originally I was thinking we could make the `IndexOptions` a "smart object" (kind of like the "Active Record" pattern) but maybe that makes this class a bit too heavy (and I believe does kind of violate SRP). 

Instead, another idea would be to keep `IndexOptions` as a lightweight DTO (data transfer object) and use the "Builder" pattern. i.e create an `IndexOptionsBuilder` that generates the options to be passed to the `AlgoliaConnector`. 

Essentially that would mean factoring out both `AlgoliaHelper::convertToIndexOptions` and `AlgoliaConnector::getIndexName` into the new `IndexOptionsBuilder` which honors SRP I think. 

The Builder (which can be a singleton and can inject the `IndexNameFetcher` via DI) has one job and one job only - create a valid `IndexOptions` object. Now instead of having `IndexOptions::getEnforcedIndexName` it's just `IndexOptions::getIndexName` because the Builder will have already figured out what the name is. Any methods that use `IndexOptions` can just call `$indexOptions->getIndexName()` and they're good to go. And now responsibilities are properly separated. 

I think I like this approach a bit better than the "Active Record" pattern. What do you think?

Model/IndexOptions.php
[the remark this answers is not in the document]
cammonro: Yeah that works!

Service/AlgoliaConnector.php
damcou: I think this is a great idea ! 

I committed an `IndexOptionsBuilder` class, following your description of it as best as I could.
It has two separate methods:
- `convertToIndexOptions` which replaces the previous `AlgoliaHelper::convertToIndexOptions` to ensure to be able to set directly an index name if needed (for the `AlgoliaHelper` class, or for testing purposes)
- `createIndexOptions` which will be the preferred way to create `IndexOptions` objects in the future (starting 3.16.0) which takes index suffix as an argument and computes the index name with the `IndexNameFetcher`

At the beginning I wanted to use only one method but this would make a lot a calls a bit weird with a lot of `null` params since everything would be optional.

As a result, `IndexOptionsInterface::ENFORCED_INDEX_NAME` with its related getter is not useful anymore and we can use directly `IndexOptionsInterface::INDEX_NAME` as you suggested.

Let me know what you think about it.

cammonro (changes requested):
I think we're almost there! I added some suggestions to further centralize responsibility and (hopefully) keep things cleaner. But this is already much improved.

cammonro (approved):
Génial!

PR #1678, opened 2025-01-15
MAGE-1004 Rebased epic for algoliaBundle removal

cammonro:
Rebased for merge to `-dev` branch in preparation for `3.15.0-beta.2`.

Confirmed to work with standard bundling but investigation continues for advanced bundling options. Full details in Jira ticket [MAGE-1004](https://algolia.atlassian.net/browse/MAGE-1004).

[MAGE-1004]: https://algolia.atlassian.net/browse/MAGE-1004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1682, opened 2025-01-16
MAGE-1004 Babel transpiler fix

cammonro:
This PR contains a small but necessary change from https://github.com/algolia/algoliasearch-magento-2/pull/1678 and is to be used in tandem with the Babel compilation steps from the companion PR outlined in Jira.

### DIFF
--- a/view/frontend/web/js/insights.js
+++ b/view/frontend/web/js/insights.js
@@ -6,7 +6,7 @@ define([
 ], function ($, algoliaAnalyticsWrapper, algoliaCommon) {
     const USE_GLOBALS = true;
 
-    algoliaAnalytics = algoliaAnalyticsWrapper.default;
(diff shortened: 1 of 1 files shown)


damcou (dismissed):
So if I understood correctly, the only difference with https://github.com/algolia/algoliasearch-magento-2/pull/1678 is the last commit about the Babel compatibility. Now that the 3.15.0-dev branch has been updated by the other PR, is it okay to merge this one which also contains all the previous commits ? Or maybe we should create a new PR with only the last commit ?

In any case, I tested the bundling and it seems to work as expected :)

cammonro:
Thank you @damcou - yup it's just one simple commit to address a transpiling issue. I thought Github would rebase post merge on the previous PR but it didn't. However, I was able to "trick" Github into rebasing by temporarily retargeting the PR.

PR #1684, opened 2025-01-17
MAGE-1191 PHP 7 compatibility on last supported minor release

cammonro:
Even though we do not officially support PHP 7 any longer (PHP 7.4 has not be supported since Magento 2.4.3 and support for that version [ended November 2022](https://experienceleague.adobe.com/en/docs/commerce-operations/release/versions#243)), version `3.13` of the Algolia extension has been compatible with PHP 7.4 to date. The recent release of `3.13.7` introduces newer PHP 8 syntax that could have an unintended impact to legacy installs. 

This PR reverts those constructs to preserve PHP 7 compatibility.


damcou (approved):
🚀

PR #1688, opened 2025-01-24
MAGE-1180: fix configurable ranges with customer group prices

damcou:
This PR contains:
- Fixed a bug where configurable product price ranges where not applied as expected when customer group prices are defined (see more explanation on JIRA)


Helper/Entity/Product/PriceManager/ProductWithChildren.php
```diff
@@ -158,28 +158,50 @@ protected function setFinalGroupPrices($field, $currencyCode, $min, $max, $dashe
...
                     $tierPrice = $this->getTierPrice($subProduct, $currencyCode, $withTax);
                     $price     = $this->getTaxPrice($product, $subProduct->getPriceModel()->getFinalPrice(1, $subProduct), $withTax);
+
                     if (!empty($tierPrice[$groupId]) && $specialPrice[$groupId] > $tierPrice[$groupId]){
                         $minPrice = $tierPrice[$groupId];
                     }
-                    $groupPriceList[$groupId]['min'] = min($minPrice, $price);
-                    $groupPriceList[$groupId]['max'] = max($max, $price);
+
+                    if (isset($tierPrice[$groupId]) && $tierPrice[$groupId] !== false) {
```
cammonro, line 179: Does this conflict with the logic that immediately precedes it? Perhaps the conditionals could be merged?

Helper/Entity/Product/PriceManager/ProductWithChildren.php
```diff
@@ -158,28 +158,50 @@ protected function setFinalGroupPrices($field, $currencyCode, $min, $max, $dashe
     {
         if (count($subproducts) > 0) {
             $groupPriceList = [];
+            $subProductsMin = -1;
```
cammonro, line 161: This seems a bit unclear. Can we maybe use named constants here like `PRICE_NOT_SET`? Is there a reason that both variables are initialized differently?

cammonro (commented):
Nice detective work tracking this down. This code here kind of feels like a minefield and will need to be refactored but I'm OK with this short term approach. A couple of minor improvements might help for maintainability to ease refactoring later.

Helper/Entity/Product/PriceManager/ProductWithChildren.php
damcou: TBH I think this is not even needed, I guess it was one of the many tests I made and I forgot to remove it, sorry for that.

Helper/Entity/Product/PriceManager/ProductWithChildren.php
```diff
@@ -158,28 +158,50 @@ protected function setFinalGroupPrices($field, $currencyCode, $min, $max, $dashe
...
+            $subProductsMax = 0;
             /** @var Group $group */
             foreach ($this->groups as $group) {
                 $groupId = (int) $group->getData('customer_group_id');
                 $minPrice = $min;
+
                 foreach ($subproducts as $subProduct) {
```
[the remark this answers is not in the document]
damcou: Makes sense, I split `setFinalGroupPrices` method a little bit so it's easier to read.

view/frontend/web/.DS_Store
cammonro, line 1: [Can we prune this file?](https://preview.redd.it/ds-store-v0-9q0jell7eqfe1.png?auto=webp&s=6d2f325cd90621ee72e1cce0f4ae80516da83777) 😜

cammonro (approved):
Thanks for tidying this up some more. Every little bit helps!

Just one small gotcha slipped by... after that feel free to merge. 🙂

PR #1689, opened 2025-01-27
MAGE-1061 Adding support for additional IS mixins

cammonro:
While [supplying updates](https://github.com/algolia/algoliasearch-custom-algolia-magento-2/pull/15) to our `CustomAlgolia` module to demonstrate the ability to use mixins with the updated `3.15.0` libraries I decide to refactor a few more bits of functionality on InstantSearch to expose setup functions without getting too ambitious. 

There's more work that needs to be done here as the legacy code makes heavy use of closures in what is presently contained in a pretty monolithic `buildInstantSearch` method. I realize we expose `allWidgetConfiguration` for modification through our hooks, but I feel it would also be very convenient to be able to granularly override things through a simple mixin. Future updates here can expose more internals in these JavaScript files to customization without overrides.

### DIFF
--- a/view/frontend/web/js/instantsearch.js
+++ b/view/frontend/web/js/instantsearch.js
@@ -24,178 +24,38 @@ define([
             this.buildInstantSearch();
         },
 
+        isStarted: false,
+
+        /**
+         * Initialize search results using Algolia's InstantSearch.js library v4
+         * Docs: https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/
+         */
         async buildInstantSearch() {
             const templateProcessor = await templateEngine.getSelectedEngineAdapter();
+
             const mockAlgoliaBundle = this.mockAlgoliaBundle();
 
-            /** We have nothing to do here if instantsearch is not enabled **/
-            if (
-                typeof algoliaConfig === 'undefined' ||
-                !algoliaConfig.instant.enabled ||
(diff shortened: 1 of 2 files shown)


view/frontend/web/js/instantsearch.js
```diff
@@ -755,33 +599,224 @@ define([
...
-    
-                search = algoliaCommon.triggerHooks(
+
+            this.startInstantSearch(search, mockAlgoliaBundle);
+
+            this.addMobileRefinementsToggle();
+        },
+
+        /**
+         * @deprecated - these hooks will be removed in a future version
```
damcou, line 609: Should it happen in 3.16 ?

damcou (approved):
Tested locally and it seems to work :) 

I like the fact that we start to split `buildInstantSearch` a little bit . I guess the next step would be to lighten the `allWidgetConfiguration` with separate methods as well, this would be even more readable. WDYT ?

view/frontend/web/js/instantsearch.js
cammonro: Yeah probably - they've been there forever it seems...

cammonro:
> Tested locally and it seems to work :)
> 
> I like the fact that we start to split `buildInstantSearch` a little bit . I guess the next step would be to lighten the `allWidgetConfiguration` with separate methods as well, this would be even more readable. WDYT ?

Yes 100%!

PR #1695, opened 2025-02-04
MAGE-1184: Entity Helpers refactoring

damcou:
This PR contains:

- Updated `AbstractEntityHelper` with a new `getIndexSettings` method to implement
- Introduced `RecordBuilderInterface`
- Introduced a `RecordBuilder` per Entity which is now responsible for formatting any record sent to Algolia
- Updated each Entity Helpers which are now freed from such responsibility
- As a consequence, `getObject` methods from Entity Helpers are now deprecated and will be removed in version 3.16.0 in favour of `RecordBuilderInterface::buildRecord` which is much more self-explanatory

### DIFF
--- a/Api/RecordBuilder/RecordBuilderInterface.php
+++ b/Api/RecordBuilder/RecordBuilderInterface.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Api\RecordBuilder;
+
+use Magento\Framework\DataObject;
+
+interface RecordBuilderInterface
+{
+    public function buildRecord(DataObject $entity): array;
+}
--- a/Helper/Entity/CategoryHelper.php
+++ b/Helper/Entity/CategoryHelper.php
@@ -5,56 +5,45 @@ namespace Algolia\AlgoliaSearch\Helper\Entity;
 use Algolia\AlgoliaSearch\Exception\CategoryEmptyException;
 use Algolia\AlgoliaSearch\Exception\CategoryNotActiveException;
 use Algolia\AlgoliaSearch\Helper\ConfigHelper;
-use Algolia\AlgoliaSearch\Helper\Image;
-use Algolia\AlgoliaSearch\Service\AlgoliaConnector;
+use Algolia\AlgoliaSearch\Service\Category\RecordBuilder as CategoryRecordBuilder;
 use Algolia\AlgoliaSearch\Service\IndexNameFetcher;
 use Magento\Catalog\Model\Category;
 use Magento\Catalog\Model\Category as MagentoCategory;
 use Magento\Catalog\Model\CategoryRepository;
-use Magento\Catalog\Model\ResourceModel\Category as CategoryResource;
 use Magento\Catalog\Model\ResourceModel\Category\Collection as CategoryCollection;
 use Magento\Catalog\Model\ResourceModel\Category\CollectionFactory as CategoryCollectionFactory;
 use Magento\Eav\Model\Config;
-use Magento\Framework\App\ResourceConnection;
 use Magento\Framework\DataObject;
 use Magento\Framework\Event\ManagerInterface;
-use Magento\Framework\Module\Manager;
-use Magento\Store\Model\Store;
+use Magento\Framework\Exception\LocalizedException;
 use Magento\Store\Model\StoreManagerInterface;
 
 class CategoryHelper extends AbstractEntityHelper
 {
     use EntityHelperTrait;
     public const INDEX_NAME_SUFFIX = '_categories';
-    protected $isCategoryVisibleInMenuCache;
-    protected $coreCategories;
-    protected $idColumn;
     protected $categoryAttributes;
     protected $rootCategoryId = -1;
-    protected $categoryNames;
 
     public function __construct(
         protected ManagerInterface          $eventManager,
         protected StoreManagerInterface     $storeManager,
-        protected ResourceConnection        $resourceConnection,
         protected Config                    $eavConfig,
         protected ConfigHelper              $configHelper,
         protected CategoryCollectionFactory $categoryCollectionFactory,
-        protected Image                     $imageHelper,
-        protected CategoryResource          $categoryResource,
         protected CategoryRepository        $categoryRepository,
-        protected Manager                   $moduleManager,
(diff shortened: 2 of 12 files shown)


Api/RecordBuilder/RecordBuilderInterface.php
cammonro, line 1: Not sure what other things would go under `RecordBuilder`? I'm wondering if it would make more sense to have a more generic `Builder` namespace and then we put all the builders under it? Maybe you had something else in mind...

Service/AdditionalSection/RecordBuilder.php
```diff
@@ -0,0 +1,43 @@
...
+     * Builds a Section record
+     *
+     * @param DataObject $entity
+     * @return array
+     */
+    public function buildRecord(DataObject $entity): array
+    {
+        $record = [
+            AlgoliaConnector::ALGOLIA_API_OBJECT_ID => $entity->getData('value'),
+            'value'                                 => $entity->getData('value'),
```
cammonro, line 26: Is this necessary?

Helper/Entity/ProductHelper.php
```diff
@@ -1379,9 +687,7 @@ public function canProductBeReindexed($product, $storeId, $isChildProduct = fals
      */
     public function productIsInStock($product, $storeId): bool
```
cammonro, line 688: Do we want to deprecate this method as well?

Service/Product/RecordBuilder.php
```diff
@@ -0,0 +1,817 @@
...
+        protected ManagerInterface       $eventManager,
+        protected DiagnosticsLogger      $logger,
+        protected Visibility             $visibility,
+        protected StoreManagerInterface  $storeManager,
+        protected ConfigHelper           $configHelper,
+        protected CategoryHelper         $categoryHelper,
+        protected AlgoliaHelper          $algoliaHelper,
+        protected ImageHelper            $imageHelper,
+        protected StockRegistryInterface $stockRegistry,
+        protected PriceManager           $priceManager,
```
cammonro, line 55: I notice you that like using the trailing commas that are now allowed in PHP 8 - is that intentional? For consistency we should agree on best approach and maybe start putting adhering to a coding style guide ... 🤔

cammonro (commented):
@damcou Great work!

All in all this is much better code organization and I would like for us to include it in 3.15. I also like that you cleaned up ambiguity in the types, removed really old references and have provided deprecation notices before we just start yanking out public methods  - which I think makes for a saner upgrade path. The really old unused stuff I'm fine with taking out as that is just clutter. 

Had a few observations / questions if you wouldn't mind taking a look.

Helper/Entity/CategoryHelper.php
```diff
@@ -90,10 +80,7 @@ public function getIndexSettings($storeId)
 
         // Additional index settings from event observer
         $transport = new DataObject($indexSettings);
-        $this->eventManager->dispatch('algolia_index_settings_prepare', [ // Only for backward compatibility
-                'store_id'       => $storeId,
-                'index_settings' => $transport,
-            ]);
+        /** Removed legacy algolia_index_settings_prepare event on 3.15.0 */
```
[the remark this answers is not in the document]
damcou: Yes ! I'm pretty sure we could consider reworking our entire set of events to make them more consistent... but ... baby steps :)

Service/AdditionalSection/RecordBuilder.php
damcou: I guess it could be not necessary (can't remember how it would behave without setting an objectID), I just ported the behaviour from `AdditionalSectionHelper`. Just wanted not to break anything :D

Helper/Entity/ProductHelper.php
damcou: I was about to do it then I remembered that this method has its plugin in the inventory module: https://github.com/algolia/algoliasearch-inventory-magento-2/blob/master/Plugin/Helper/ProductHelperPlugin.php#L50

From what I remember, this module was introduced because we wanted to keep compatibility with Magento 2.3 back in the days, I'm wondering if we could not just add the MSI handling directly in the codebase now ?

Service/Product/RecordBuilder.php
damcou: Not really intentional no, do you prefer that we don't use trailing commas at all ?

Api/RecordBuilder/RecordBuilderInterface.php
damcou: I just wanted to put the `RecordBuilder` directory at the same level as the `IndexBuilder` one. Maybe at some point we would need an interface which extends the RecordBuilderInterface one as in the IndexBuilder directory. 
I can make the `Builder` namespace change in other PR if you'd like but it's gonna change a lot of files, I don't want to overload this PR which is already pretty massive.

cammonro: My thinking here is that once we add an API definition we should not change it unlike our implementation classes which we can progressively refactor. So we should be really sure of our choices on this or strive to be generic where it makes more sense. Since we are introducing both of these as part of 3.15 (and we are still in beta) it makes sense to me to make this API introduction atomic (i.e. get it the way we want before release - if we don't change it now we never change it after). It doesn't have to block the particular PR - just a thought.

Helper/Entity/ProductHelper.php
cammonro: Yes that would be my preference as MSI is pretty much standard as of 2.4. Why make customers install a different module? Maybe we deprecate but don't specifically mention 3.16 in case we don't merge MSI handling until 3.17 for instance. At any rate, this is good validation for us to continue to use this deprecation approach going forward - because if we are using plugins on this stuff then surely other people might as well.

Service/Product/RecordBuilder.php
cammonro: I'm fine either way now that we're no longer supporting PHP 7. I was just thinking maybe we should defined a style guide for stuff. Let's chat about it later. Not a blocker at all! 🙂

Api/RecordBuilder/RecordBuilderInterface.php
damcou: I can create a ticket to tackle this in the next sprint, I'm ok to centralize everything in the same directory.

PR #1699, opened 2025-02-25
Reset customer group collection after filtering

kamilszewczyk:
Without clearing customer group collection, any loops on existing collection will not be filtered

**Summary**

Add field filtering without clearing the collection (or creating new instance of it) will not actually do anything. Excluding websites should speed up the process of indexing by not requiring retrieval of grouped prices in websites that are excluded.
Without this modification this code doesn't do anything.

**Result**

With Customer Group pricing option enabled, any runs of `foreach ($this->groups as $group)` for example in vendor/algolia/algoliasearch-magento-2/Helper/Entity/Product/PriceManager/ProductWithChildren.php::setFinalGroupPrices should only loop through customer groups that are NOT excluded in the current website scope.

Without the change:
image

After the change:
image

### DIFF
--- a/Helper/Entity/Product/PriceManager/ProductWithoutChildren.php
+++ b/Helper/Entity/Product/PriceManager/ProductWithoutChildren.php
@@ -135,6 +135,7 @@ abstract class ProductWithoutChildren
             }
             if(count($excludedGroups) > 0) {
                 $this->groups->addFieldToFilter('main_table.customer_group_id', ["nin" => $excludedGroups]);
+                $this->groups->clear()
             }
         }
         // price/price_with_tax => true/false


cammonro:
@kamilszewczyk  Thank you for identifying this bug! This is a sensible fix and we'll be sure to include it in a future patch release.

damcou:
Thanks a lot for your contribution @kamilszewczyk ! 🙌 
Your fix has been approved and added to our upcoming versions [3.14.5](https://github.com/algolia/algoliasearch-magento-2/pull/1705) and [3.15.0](https://github.com/algolia/algoliasearch-magento-2/pull/1704) ! 
Closing this ticket, thanks again !

PR #1700, opened 2025-02-26
MAGE-1113 Refactor CLI for 3.15

cammonro:
**Summary**

Brings refactoring work on console commands into 3.15+ from https://github.com/algolia/algoliasearch-magento-2/pull/1660.

**Result**

image


damcou (commented):
Thanks for having introduced the `EntityIndexOptionsBuilder`, I think this makes sense. I don't have any additional feedback other than the ones in https://github.com/algolia/algoliasearch-magento-2/pull/1660

damcou (approved):
🚀

PR #1701, opened 2025-02-27
MAGE-1113: Add manual full reindexing + indexing manager page

damcou:
WIP ...

### DIFF
--- a/Model/Indexer/Category.php
+++ b/Model/Indexer/Category.php
@@ -28,6 +28,10 @@ class Category implements \Magento\Framework\Indexer\ActionInterface, \Magento\F
 
     public function execute($categoryIds)
     {
+        if (!$this->configHelper->isCategoriesIndexerEnabled()) {
+            return;
+        }
+
         $storeIds = array_keys($this->storeManager->getStores());
 
         foreach ($storeIds as $storeId) {
--- a/Model/Indexer/Page.php
+++ b/Model/Indexer/Page.php
@@ -30,6 +30,10 @@ class Page implements \Magento\Framework\Indexer\ActionInterface, \Magento\Frame
      */
     public function execute($ids)
     {
+        if (!$this->configHelper->isPagesIndexerEnabled()) {
+            return;
+        }
+
         $storeIds = $this->pageHelper->getStores();
 
         foreach ($storeIds as $storeId) {
--- a/Model/Indexer/Product.php
+++ b/Model/Indexer/Product.php
@@ -35,6 +35,10 @@ class Product implements \Magento\Framework\Indexer\ActionInterface, \Magento\Fr
      */
     public function execute($productIds)
     {
+        if (!$this->configHelper->isProductsIndexerEnabled()) {
+            return;
+        }
+
         $storeIds = array_keys($this->storeManager->getStores());
         $areParentsLoaded = false;
 
--- a/etc/adminhtml/system.xml
+++ b/etc/adminhtml/system.xml
@@ -984,6 +984,69 @@
                 </field>
             </group>
         </section>
+        <section id="algoliasearch_indexing_manager" translate="label" type="text" sortOrder="66" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Indexing Manager</label>
+            <tab>algolia</tab>
+            <resource>Algolia_AlgoliaSearch::algolia_algoliasearch</resource>
+            <group id="full_indexing" translate="label" type="text" sortOrder="10" showInDefault="1" showInWebsite="0" showInStore="0">
+                <label>Full indexing via Magento indexers</label>
+                <comment>Since version 3.16.0, the following Magento indexers are now disabled by default. We highly suggest to keep this configuration as it is and run the newly created CLI commands instead.</comment>
+                <field id="products" translate="label comment" type="select" sortOrder="10" showInDefault="1" showInWebsite="0" showInStore="0">
+                    <label>Enable Products Indexer (algolia_products)</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <comment>
+                        <![CDATA[
+                            We highly recommend to run <code>bin/magento algolia:reindex:products</code> instead
+                        ]]>
+                    </comment>
+                </field>
+                <field id="categories" translate="label comment" type="select" sortOrder="20" showInDefault="1" showInWebsite="0" showInStore="0">
+                    <label>Enable Categories Indexer (algolia_categories)</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <comment>
+                        <![CDATA[
+                            We highly recommend to run <code>bin/magento algolia:reindex:categories</code> instead
+                        ]]>
+                    </comment>
+                </field>
+                <field id="pages" translate="label comment" type="select" sortOrder="30" showInDefault="1" showInWebsite="0" showInStore="0">
+                    <label>Enable Pages Indexer (algolia_pages)</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <comment>
+                        <![CDATA[
+                            We highly recommend to run <code>bin/magento algolia:reindex:pages</code> instead
+                        ]]>
+                    </comment>
+                </field>
+                <field id="suggestions" translate="label comment" type="select" sortOrder="40" showInDefault="1" showInWebsite="0" showInStore="0">
+                    <label>Enable Suggestions Indexer (algolia_suggestions)</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <comment>
+                        <![CDATA[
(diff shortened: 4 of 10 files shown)


Api/Builder/QueueBuilderInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Api\Builder;
+
+interface QueueBuilderInterface
```
cammonro, line 5: Like the idea of an interface for this - but I'm not sure of the name as it suggests that the class is responsible solely for constructing a queue. However, this service can handle batching, queuing, and potentially processing the items (depending on queue enablement configuration).

Perhaps something like `BatchQueueProcessor` might make more sense?

Console/Command/Indexer/IndexAllCommand.php
```diff
@@ -0,0 +1,59 @@
...
+        return [];
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $this->output = $output;
+        $this->input = $input;
+        $this->setAreaCode();
+
+        foreach ($this->commandsList as $commandName) {
```
cammonro, line 52: Should we call out that if the queue is enabled indexing will be processed asynchronously on these commands?

Console/Command/Indexer/ProcessQueueCommand.php
```diff
@@ -0,0 +1,68 @@
...
+        protected Queue $queue,
+        protected AlgoliaCredentialsManager $algoliaCredentialsManager,
+        ?string $name = null
+    ) {
+        parent::__construct($name);
+    }
+
+    protected function getCommandName(): string
+    {
+        return 'algolia:reindex:process_queue';
```
cammonro, line 32: This change will impact our [current guidelines](https://www.algolia.com/doc/integration/magento-2/how-it-works/indexing-queue/?client=php#process-the-queue). I'm thinking rather than just pointing to a new CLI, this is a good opportunity to finally abandon the explicit separate crontab and instead use Magento's `crontab.xml` setup. We have a task for this in the backlog.

Console/Command/Indexer/AbstractIndexerCommand.php
```diff
@@ -0,0 +1,41 @@
...
+        State $state,
+        StoreNameFetcher $storeNameFetcher,
+        ?string $name = null
+    ) {
+        parent::__construct($state, $storeNameFetcher, $name);
+    }
+
+    protected function getCommandPrefix(): string
+    {
+        return parent::getCommandPrefix() . 'reindex:';
```
cammonro, line 23: We're going to need a task to update our [documentation](https://www.algolia.com/doc/integration/magento-2/how-it-works/indexing/?client=php#manual-indexing) for these new CLIs.

Model/Indexer/Product.php
```diff
@@ -2,119 +2,34 @@
...
-                /** @uses ProductIndexBuilder::buildIndexFull() */
-                $this->queue->addToQueue(
-                    ProductIndexBuilder::class,
-                    'buildIndexFull',
-                    $data,
-                    $productsPerPage,
-                    true
-                );
-            }
+        if (!$this->configHelper->isProductsIndexerEnabled()) {
```
cammonro, line 27: We need to limit our scope on this disable operation to `executeFull` and not `execute` - otherwise this will break how materialized views are handled which is essential for delta updates. (Applies to all affected entity indexers)

Model/Indexer/Category.php
```diff
@@ -3,55 +3,29 @@
...
-            $this->rebuildAffectedProducts($storeId);
-
-            $categoriesPerPage = $this->configHelper->getNumberOfElementByPage();
-
-            if (is_array($categoryIds) && count($categoryIds) > 0) {
-                $this->processSpecificCategories($categoryIds, $categoriesPerPage, $storeId);
-
-                continue;
-            }
+        if (!$this->configHelper->isCategoriesIndexerEnabled()) {
```
cammonro, line 23: Restriction needs to move to `executeFull`.

Model/Indexer/Page.php
```diff
@@ -2,61 +2,34 @@
...
-
-                return;
-            }
-
-            if ($this->isPagesInAdditionalSections($storeId)) {
-                $data = ['storeId' => $storeId];
-                if (is_array($ids) && count($ids) > 0) {
-                    $data['options'] = ['entityIds' => $ids];
-                }
+        if (!$this->configHelper->isPagesIndexerEnabled()) {
```
cammonro, line 27: Restriction needs to move to `executeFull`.

etc/adminhtml/system.xml
```diff
@@ -984,6 +984,78 @@
                 </field>
             </group>
         </section>
+        <section id="algoliasearch_indexing_manager" translate="label" type="text" sortOrder="66" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Indexing Manager</label>
+            <tab>algolia</tab>
+            <resource>Algolia_AlgoliaSearch::algolia_algoliasearch</resource>
+            <group id="full_indexing" translate="label" type="text" sortOrder="10" showInDefault="1" showInWebsite="0" showInStore="0">
+                <label>Full indexing via Magento indexers</label>
+                <comment>Since version 3.16.0, the following Magento indexers are now disabled by default. We highly suggest to keep this configuration as it is and run the newly created CLI commands instead.</comment>
```
cammonro, line 993: I think we may need to clarify our language here that these configurations only disable "automatic full indexing" as we are not (or should not be) disabling indexers completely because we want to handle **delta** operations. We do mention "Full indexing via Magento indexers" in the title so that's good but reinforcing this idea further below would help eliminate potential confusion. 

Of course another option is to provide _even more granularity_ in the configuration in case they _want_ to disable delta indexers... 🤔  That would mean two sets of configurations... Thoughts?

view/adminhtml/templates/indexing-manager/overview.phtml
```diff
@@ -0,0 +1,14 @@
+<?php
+/** @var \Magento\Backend\Block\Template $block */
+?>
+<div class="algolia-indexing-manager-overview">
+    <div class="algolia_dashboard_warning">
+        <p>Here is a list of all the indices created by the extension on the Algolia Dashboard.</p>
```
cammonro, line 6: I really love this new feature - but we should clarify that the indices listed exclude replicas.

cammonro (changes requested):
HUGE work @damcou !! 🙌

I **love** the new Indexing Manager page. What a great idea!

I called out some things that need attention but this is just super. 

Most of these raised items can be separate tasks but the main immediate concern is the disablement piece as we have to make sure that delta indexing is not adversely impacted.

Lastly we'll also need to revisit the integration tests since they rely on the legacy indexer models. 

Thank you for the big lift here. Bravo! 👏

Api/Builder/QueueBuilderInterface.php
damcou: I agree that this service is supposed to do "more" than just building the queue indeed. This issue is that historically, this is the queue model which handles the queue enablement configuration so we are in a weird spot where we use `addToQueue` in any case,  even if the queue is disabled. I agree to rename it but I think this is more a architecture issue that may need to be tackled one day 👍

Console/Command/Indexer/AbstractIndexerCommand.php
damcou: task created (MAGE-1225)

Console/Command/Indexer/ProcessQueueCommand.php
damcou: I think that this is still better to have a "non Magento indexer" command for this, because we want to make sure that this isn't processed on setup:upgrade for example. I don't think that creating this would be bad but I'm ok to explore the `crontab.xml` path but it's scheduled for Phase 2. 

In the meantime, maybe `algolia_queue_runner` indexer shouldn't be disabled by default to avoid breaking changes on upgrade for users who already have setup their cron as in the doc ?

Model/Indexer/Product.php
damcou: Makes sense ! Sorry for that.

etc/adminhtml/system.xml
damcou: The goal of this section of the admin will be enhanced in the future during the different phases of the indexing improvements, and introducing more granularity can definitely be part of this enhancement :) .
In the meantime, I'll try to be more explicit.

view/adminhtml/templates/indexing-manager/overview.phtml
damcou: Done!

I think there is a lot of opportunities to improve the page here (thinking of displaying the number of replicas, number of records, the application each index belong to, last update, etc ... ) . Let's improve this all along the indexing improvement phases !

Api/Builder/QueueBuilderInterface.php
cammonro: Yes we often find ourselves in "weird spots"! 😂

cammonro (approved):
❤️

PR #1702, opened 2025-03-04
MAGE-973 Default renderingContent

cammonro:
**Summary**

This primary of objective of this PR is introduce default `renderingContent` that will help power `dynamicWidgets`. It also includes some refactoring:

- New `FacetBuilder` service
- Shared `rawFacets` between `renderingContent` and `attributesForFaceting`
- Separating logic into functions by area of concern: groups, categories,
- Introducing constants for Magento integration specific key conventions
- Added unit tests and reorganized existing

During implementation it was observed that our currency handling is flawed as it does not handle scoped currency config by store but this PR does not address that issue. That will be addressed in a separate ticket. 

**Result**

`renderingContent` persisted as Facet Display in Dashboard:
image

All tests passed (including reorganized tests)
image

### DIFF
--- a/Service/Product/FacetBuilder.php
+++ b/Service/Product/FacetBuilder.php
@@ -44,7 +44,7 @@ class FacetBuilder
         $facets = $this->configHelper->getFacets($storeId);
         foreach ($facets as $facet) {
             if ($facet['attribute'] === self::FACET_ATTRIBUTE_PRICE) {
-                $attributesForFaceting = array_merge($attributesForFaceting, $this->getPricingAttributesForFaceting($storeId));
+                $attributesForFaceting = array_merge($attributesForFaceting, $this->getPricingAttributes($storeId));
             } else {
                 $attributesForFaceting[] = $this->decorateAttributeForFaceting($facet);
             }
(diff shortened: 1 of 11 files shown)


Service/Product/FacetBuilder.php
```diff
@@ -0,0 +1,242 @@
...
+     * @param int $storeId
+     * @param string $currencyCode
+     * @return string[]
+     * @throws NoSuchEntityException
+     * @throws LocalizedException
+     */
+    protected function getGroupPricingAttributes(int $storeId, string $currencyCode): array
+    {
+        $groupPricingAttributes = [];
+        $websiteId = (int) $this->storeManager->getStore($storeId)->getWebsiteId();
```
damcou, line 209: Micro improvement here, but can it be under the if condition as well ?

Service/Product/FacetBuilder.php
```diff
@@ -0,0 +1,242 @@
...
+
+use Algolia\AlgoliaSearch\Helper\ConfigHelper;
+use Magento\Customer\Api\GroupExcludedWebsiteRepositoryInterface;
+use Magento\Customer\Model\ResourceModel\Group\Collection as GroupCollection;
+use Magento\Directory\Model\Currency as CurrencyHelper;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Exception\NoSuchEntityException;
+use Magento\Store\Model\StoreManagerInterface;
+
+class FacetBuilder
```
damcou, line 13: Wondering if it's worth adding a `FacetBuilderInterface` with a single `getAttributesForFaceting` method to implement, but I guess this will be only useful for products so that's not mandatory 🤷

Service/Product/FacetBuilder.php
```diff
@@ -0,0 +1,242 @@
...
+            array_map(
+                function(string $attribute) {
+                    if ($attribute === self::FACET_ATTRIBUTE_CATEGORIES) {
+                        $attribute = self::FACET_ATTRIBUTE_CATEGORIES . '.level0';
+                    }
+                    return $attribute;
+                },
+                $facets
+            ),
+            array_fill(0, count($facets), [ 'sortRemainingBy' => 'alpha' ])
```
damcou, line 107: Do we plan to make this configurable in the admin in the future or only with backend events ?

Test/Unit/Service/Product/FacetBuilderTestable.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Test\Unit\Service\Product;
+
+use Algolia\AlgoliaSearch\Service\Product\FacetBuilder;
+
+class FacetBuilderTestable extends FacetBuilder
```
damcou, line 7: Just wanted to check why is this needed ?

damcou (approved):
Very good work 🚀 

I just have questions more than changes requested, I think you can merge it afterwards

Service/Product/FacetBuilder.php
```diff
@@ -0,0 +1,242 @@
...
+     * @param string[] $facets
+     * @return array<string, array>
+     */
+    protected function getRenderingContentValues(array $facets): array
+    {
+        return array_combine(
+            array_map(
+                function(string $attribute) {
+                    if ($attribute === self::FACET_ATTRIBUTE_CATEGORIES) {
+                        $attribute = self::FACET_ATTRIBUTE_CATEGORIES . '.level0';
```
damcou, line 101: Is it normal that I don't get the `.level0` in my Algolia dashboard and that the `sortRemaingBy` attribute is not applied ? 

image

Service/Product/FacetBuilder.php
cammonro: Good call!

Service/Product/FacetBuilder.php
cammonro: Maybe... I considered it but couldn't imagine other implementations of it at this time. I also didn't want to commit to the public API just yet - at least until the `dynamicWidgets` work is fully implemented.

Service/Product/FacetBuilder.php
cammonro: Good question! We certainly could make it configurable. But yes as of right now it would require using our backend events (or a preference class if they really want to customize things). 

I'll make a separate task as we'd need to not only alter the backend but add to the admin grid UI for configuring the facets. It also doesn't really apply to everything (like pricing - the `sortRemainingBy` doesn't really seem to matter based on my tests). I also was trying to keep the new feature simple and less dependent on Magento. At some point I'd prefer to make Algolia the source of truth and not our local configuration for this kind of stuff but that's a big lift for a future time.

Test/Unit/Service/Product/FacetBuilderTestable.php
cammonro: Ah! This is purely so I could perform unit tests on `protected` methods that I did not want to expose in the API.

damcou: Ah makes sense ! Very smart !

Service/Product/FacetBuilder.php
cammonro: Fixed:

Thanks for catching this!

Will add unit test later.

damcou: Works on my side as well now :)

cammonro: Unit test added
image

PR #1709, opened 2025-03-13
MAGE-974 InstantSearch extension points 

cammonro:
**Summary**

This PR aims to clean up the monolithic JS implementation with its heavy use of broadly scoped variables and closures and scattered implementation. We historically have had hooks to help customize the way the JavaScript works but it's very limited. Now it will be possible to write a mixin on any function to avoid overwriting the entire JS file. 

- All logic has been split granularly into named, mixable functions 
- Functions have been named in order to distinguish between InstantSearch widget objects, special facet config objects and builders
- XSS vulnerabilities were identified and corrected 
- Behavior of conjunctive vs disjunctive facets has been clarified
- InstantSearch lib has been upgraded to v4.78
- Source maps have been added to InstantSearch to assist with debugging 
- jQuery dependency has been significantly reduced (still not there - more work to do here but the goal is to remove altogether) 
- Non relevant legacy code has been removed 
- A discovered JS bug with empty hits has been resolved (this was impacting the execution chain and creating some weird behaviors with other widgets) 
- Deprecation notices added - including UA sunsetted July 1, 2023
- Facet building operations have been refactored for clarity of intent
- A new front end hook called `beforeFacetInitialization` has been added which allows a user to extend the functionality by adding, removing or modifying "builder" functions which are used to define a facet config that will drive the rendering of facets 

**Result**

Unit tests need to be added for this large revamp but manual testing has been performed including (but not limited to):

- General functionality
image
- XSS (and suggestions custom widget)
image
- nodeValue undefined bug fix
image
image
- Conjunctive facets (facet count change)
image
image
- Disjunctive facets (no facet count change)
image
image
- URL redirect on add to cart
image


view/frontend/web/js/instantsearch.js
```diff
@@ -19,596 +19,694 @@ define([
...
 
         isStarted: false,
 
+        minQuerySuggestions: 4,
+
         /**
          * Initialize search results using Algolia's InstantSearch.js library v4
          * Docs: https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/
          */
         async buildInstantSearch() {
```
damcou, line 34: I like the fact that there are a lot more comments / JSdoc in the whole file for each method, I feel that the only thing missing is a full explanation on how `buildInstantSearch` works exactly. I could see a small comment before each of the methods used in it with what it does and which hooks it triggers, etc without having to navigate in the file since this is a pretty big one ... This would be the perfect entry point to understand everything at first sight WDTY ?

damcou (approved):
Wonderful work ! 🥳 💪 

I really like how the file looks now, this is a lot easier to understand, even for a non-js developer ;) .
Added a few comments but nothing blocking for you to go forward.

Out of curiosity, have you considered splitting the file into multiple ones since this is still a 1000+ lines one ? This isn't mandatory for sure, I guess this has its own pros.

view/frontend/web/js/instantsearch.js
```diff
@@ -19,596 +19,701 @@ define([
...
-        initialize(config, element) {
-            // console.log('IS initialized with', config, element);
-            this.buildInstantSearch();
+        initialize(_config, _element) {
+            this.buildInstantSearch().then(() => console.log("[Algolia] InstantSearch build complete"));
         },
 
         isStarted: false,
 
+        minQuerySuggestions: 4,
```
damcou, line 28: Is there a particular reason why this is the only value that is needed here ?

README.md
```diff
@@ -79,14 +79,15 @@ Please check our [Custom Extension](https://github.com/algolia/algoliasearch-cus
...
+| v3.14.x           | [1.6.3](https://github.com/algolia/autocomplete.js/tree/v1.6.3)*   | [4.63.0](https://github.com/algolia/instantsearch/tree/instantsearch.js%404.63.0)* | [2.11.0](https://github.com/algolia/search-insights.js/tree/v2.11.0) | [1.15.0](https://github.com/algolia/recommend/tree/v1.15.0) |
+| v3.15.x           | [1.17.9](https://github.com/algolia/autocomplete.js/tree/v1.17.9)  | [4.77.0](https://github.com/algolia/instantsearch/tree/instantsearch.js%404.77.0)  | [2.17.3](https://github.com/algolia/search-insights.js/tree/v2.17.3) | [1.16.0](https://github.com/algolia/recommend/tree/v1.16.0) |
+| v3.16.x           | [1.17.9](https://github.com/algolia/autocomplete.js/tree/v1.17.9)  | [4.78.0](https://github.com/algolia/instantsearch/tree/instantsearch.js%404.77.0)  | [2.17.3](https://github.com/algolia/search-insights.js/tree/v2.17.3) | [1.16.0](https://github.com/algolia/recommend/tree/v1.16.0) |
```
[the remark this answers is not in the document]
cammonro: Yeah removing the bundle makes this a lot easier.

etc/adminhtml/system.xml
```diff
@@ -404,9 +404,10 @@
                     <backend_model>Algolia\AlgoliaSearch\Model\Backend\Facets</backend_model>
                     <comment>
                         <![CDATA[
-                            Specify the filtering options you want to offer on a search results page.<br>
-                                &nbsp; - Disjunctive allows a selection of different values per facet (e.g. hotels with 4 OR 5 stars).<br>
-                                &nbsp; - Conjunctive allows a selection of only one value per facet (e.g. only "Size M").
+                            Specify the <a href="https://www.algolia.com/doc/guides/managing-results/refine-results/faceting/?utm_source=magento&utm_medium=extension&utm_campaign=magento_2&utm_term=shop-owner&utm_content=doc-link#conjunctive-and-disjunctive-faceting" target="_blank">filtering options</a> you want to offer on a search results page.<br>
```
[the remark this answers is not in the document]
cammonro: No it's not! I had to double check the docs to be sure of myself but the old description was definitely wrong.

view/frontend/web/js/instantsearch.js
cammonro: You're right. I should call out that this is the main entry point. Good call. I could also perhaps add an overview of the general concept.

My preference is to keep the comments within the body of a function at a minimum and instead document with JSDoc at the method level and try to use descriptive names for those functions. Let me see if I can spruce it up.

view/frontend/web/js/instantsearch.js
cammonro: It was hardcoded in the old version so I lifted the value in case someone wanted to write a simple mixin to override it without having to copy and hack the entire file. We could add an admin config later if we wish but I don't think it's that critical.

cammonro:
Thanks for the feedback! I'll pick up these observations on the next PR. 👍 

> Out of curiosity, have you considered splitting the file into multiple ones since this is still a 1000+ lines one ? This isn't mandatory for sure, I guess this has its own pros.

I did consider this! What's holding me back is that I would like to make these components available as UMD modules at some point for applications outside of Luma / RequireJS (like Hyvä etc.). Having too many dependencies can make that cumbersome unless we implement a build process for that (and we just ditched the build process necessary for `algoliaBundle` in the last release). 

For now, I'm keeping the monolithic approach but just adding a bit of structure and scoping because before it was extremely brittle. Happy to revisit this in the future - but hopefully this is a step in the right direction.

view/frontend/web/js/instantsearch.js
cammonro: JSDoc enhanced in https://github.com/algolia/algoliasearch-magento-2/pull/1710

cammonro: JSDoc enhanced in https://github.com/algolia/algoliasearch-magento-2/pull/1710

PR #1710, opened 2025-03-17
Feat/mage 975 dynamic widgets

cammonro:
**Summary**

This PR finally introduces "out-of-the-box" support for dynamic facets by means of InstantSearch's `dynamicWidgets` feature.

Magento remains the source of truth for default configuration but now merchandising rules can override the rendering of facets based on any condition supported by the Algolia rules framework. This does rely on default `renderingContent` which was introduced in https://github.com/algolia/algoliasearch-magento-2/pull/1702 

Because of the risk of possible impact to customized implementations this capability has been feature flagged via a new Magento admin configuration called "Enable dynamic facets":
image

**Result**

InstantSearch experience using default `renderingContent`:
image

Create a category merchandising rule with a single facet:
image
image


cammonro:
Rebasing against updated `3.16.0-dev`.

damcou:
The `renderingContent` change looks good to me :)

cammonro:
I observed a flaw in the DOM logic with the initial implementation of dynamic widgets so I needed to tweak things a bit more. 

Historically we rigidly add widgets to the DOM by assigning unique containers in the config. The widget is then configured in InstantSearch, and although you can later wrap it in a `dynamicWidgets` widget to control the visibility via rules, it is insufficient to support sorting - as the DOM has been already assigned. 

To make it truly dynamic the container must be assigned by the `dynamicWidgets` callback which passes the `container` as an argument. e.g.

Therefore I modified things to *defer* the configuring of widgets in InstantSearch if the widget is eligible to be applied as a dynamic widget. The container is then derived at runtime. 

In my testing, this achieved parity with the facets as configured in the rule when applied... eg.

image
image

damcou (approved):
LGTM ! 🚀

PR #1711, opened 2025-03-18
MAGE-1228 Add proxies to constructor to defer store ops

cammonro:
**Summary**

This addresses an install issue encountered on Community Edition when installing Algolia via Composer before Magento. It uses proxies to defer the dependency chain that attempts to access store config before it has been defined. 

The error resolved is: "In WebsiteRepository.php line 159:  The default website isn't defined. Set the website and try again."

**Result**

Before fix:

image

After fix:

image

### DIFF
--- a/etc/di.xml
+++ b/etc/di.xml
@@ -185,6 +185,13 @@
         </arguments>
     </type>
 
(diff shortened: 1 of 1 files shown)


damcou (approved):
Good catch ! 🚀 
Thanks for your investigation !

PR #1712, opened 2025-03-27
MAGE-1245 Fix category FPC

cammonro:
**Summary**

This PR tackles an issue with FPC not working on category PLPs when applying cache contexts with the backend rendering feature. 

- [An earlier fix was supplied](https://github.com/algolia/algoliasearch-magento-2/pull/1161) that updates the data that powers `getVaryString()` and ultimately `\Magento\Framework\App\PageCache\Identifier::getData()` but it was discovered to not be working reliably over the course of the request. 
- Additional checks have been added to ensure consistent behavior of the context data depending on how the method is invoked.
- Unit tests have been added
- Integration tests have been added to cover various scenarios:
    - With default backend render
    - No backend render
    - Allowed user agents for backend render
 - Integration test `Response` object did not reliably return relevant headers depending on how plugins were applied so a spy was added to ensure the right header is invoked to verify use cases
 - Removed source map refs from new JS lib files which were clogging up log files
 - As we plan on improving the backend rendering in the future, current functionality has been marked as deprecated

**Result**
Unit tests:
image

Integration tests:
image


Plugin/RenderingCacheContextPlugin.php
```diff
@@ -59,4 +58,52 @@ public function afterGetData(HttpContext $subject, $data)
...
+
+    /**
+     * This can be called in a variety of contexts - so this should always be called consistently
+     *
+     * @param int $storeId
+     * @return bool
+     */
+    protected function isCategoryPage(int $storeId): bool
+    {
+        return $this->request->getControllerName() === self::CATEGORY_CONTROLLER
```
damcou, line 95: So if I understood correctly, this is the main difference with the current implementation, getting the controller name can be not sufficient in some cases, that's why we also check the rewrites ?

Test/Integration/Category/CategoryCacheTest.php
```diff
@@ -0,0 +1,341 @@
...
+            "expected MISS on category {$name} id {$categoryId}"
+        );
+        $this->assertContains(
+            'FPC',
+            explode(',', $response->getHeader('X-Magento-Tags')->getFieldValue()),
+            "expected FPC tag on category {$name} id {$categoryId}"
+        );
+
+        if ($hasProducts) {
+            $this->assertDoesNotMatchRegularExpression('/<div.*class=.*products-grid.*>/', $response->getContent(), 'Backend content was rendered.');
```
damcou, line 163: As discussed on slack, this test fails for me for some reason, maybe because I only cache one version on the page ? This may be related to integration test config though

Test/Integration/Category/CategoryCacheTest.php
```diff
@@ -0,0 +1,341 @@
...
+     * @dataProvider getCategoryProvider
+     * @depends      testCategoryPlpMissBackendRenderOn
+     * @magentoConfigFixture current_store system/full_page_cache/caching_application 1
+     * @magentoConfigFixture current_store algoliasearch_advanced/advanced/prevent_backend_rendering 0
+     * @magentoConfigFixture current_store algoliasearch_instant/instant/replace_categories 1
+     * @magentoCache full_page enabled
+     */
+    public function testCategoryPlpHitBackendRenderOn(int $categoryId, string $name): void
+    {
+        $replace = $this->config->getValue('algoliasearch_instant/instant/replace_categories', ScopeInterface::SCOPE_STORE);
```
damcou, line 124: Maybe the two first lines of each test can be factorized ?

Test/Integration/Category/CategoryCacheTest.php
```diff
@@ -0,0 +1,341 @@
...
+    {
+        return $this->url . $categoryId;
+    }
+
+    protected function dispatchCategoryPlpRequest(int $categoryId): void
+    {
+        $this->dispatchHttpRequest($this->getCategoryUrl($categoryId));
+    }
+
+    /**
```
damcou, line 296: I didn't understood that the whole goal of this method was to ensure we get a HIT, maybe this is worth adding it in the description ?

Test/Integration/Category/CategoryCacheTest.php
```diff
@@ -0,0 +1,341 @@
...
+
+use Magento\Framework\App\Cache\Manager as CacheManager;
+use Magento\Framework\App\Config\ScopeConfigInterface;
+use Magento\Framework\App\Response\Http as ResponseHttp;
+use Magento\Framework\Indexer\IndexerRegistry;
+use Magento\Store\Model\ScopeInterface;
+use Magento\TestFramework\ObjectManager;
+use Magento\TestFramework\TestCase\AbstractController;
+
+class CategoryCacheTest extends AbstractController
```
damcou, line 13: I fear that if we start to mix every type of tests in our current directory structure, it's going to be hard to maintain. What do you think about creating folders `Test/Integration/Indexing` where we can put every existing test related to indexing , and `Test/Integration/Frontend` where we could put this test and the existing `SearchTest.php `for example ?

damcou (approved):
Very good job ! 👏 
Got some suggestions.
I tested on the frontend, I now have the HIT on second request and two different versions of the page as expected.
Unit tests are passing, Integration tests have a small issue as discussed (not sure if it's related to the test itself though)

Plugin/RenderingCacheContextPlugin.php
cammonro: Yes there were 3 scenarios I observed depending on who the request is dispatched 

1. With a controller (on MISS) 
or a URL (on HIT) which can either be 
1. The real path or
1. The rewrite path where we then need to check the original path

The objective in all 3 cases is to correctly discern if the data is being generated on a request for a category page for the entire lifecyle. If we don't derive that correctly then the context changes and the cache is broken.

Test/Integration/Category/CategoryCacheTest.php
cammonro: It's not failing for me so I suspect it's the test throwing off a false positive on the RegEx. To confirm we'd need to inspect the content you're seeing on your local. I'm suspecting it might be injecting a different type of `products-grid` that is not related to the "real" `products-grid` I'm looking for. Once we determine that we can fix the test but since I can't reproduce I'll leave it for now.

Test/Integration/Category/CategoryCacheTest.php
cammonro: Yup definitely some redundancy here we can streamline. I was mainly making sure an accidental copy/paste on an annotation doesn't cause the test to fail for the wrong reason.

Test/Integration/Category/CategoryCacheTest.php
cammonro: This is an excellent idea! I'll reorganize this a bit.

cammonro: I went and reorganized both the integration and unit tests...

`Integration\Indexing`
image

`Integration\Frontend`
image

`Unit`
image

PR #1716, opened 2025-03-31
MAGE-1244: Fix `$affectedProductIds` logic

damcou:
This PR contains:
- A fix to the way the Category indexer handles the products reindexing a category contains when the CategoryObserver is triggered (an issue was introduced with the refactoring of the indexers). 

Tests on the Category indexing will be part of an upcoming effort

### DIFF
--- a/Model/Indexer/Category.php
+++ b/Model/Indexer/Category.php
@@ -12,8 +12,6 @@ use Magento\Store\Model\StoreManagerInterface;
  */
 class Category implements \Magento\Framework\Indexer\ActionInterface, \Magento\Framework\Mview\ActionInterface
 {
-    public static $affectedProductIds = [];
-
     public function __construct(
         protected StoreManagerInterface $storeManager,
         protected ConfigHelper $configHelper,
@@ -43,9 +41,6 @@ class Category implements \Magento\Framework\Indexer\ActionInterface, \Magento\F
 
     public function executeRow($id)
     {
-        if (count(self::$affectedProductIds)) {
-            $this->categoryBatchQueueProcessor->setAffectedProductIds(self::$affectedProductIds);
-        }
         $this->execute([$id]);
     }
 }
--- a/Model/Indexer/Category.php
+++ b/Model/Indexer/Category.php
@@ -12,6 +12,8 @@ use Magento\Store\Model\StoreManagerInterface;
  */
 class Category implements \Magento\Framework\Indexer\ActionInterface, \Magento\Framework\Mview\ActionInterface
 {
+    public static $affectedProductIds = [];
+
     public function __construct(
         protected StoreManagerInterface $storeManager,
         protected ConfigHelper $configHelper,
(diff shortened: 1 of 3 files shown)


Model/Indexer/Category.php
```diff
@@ -41,6 +43,9 @@ public function executeList(array $ids)
 
     public function executeRow($id)
     {
+        if (count(self::$affectedProductIds)) {
+            $this->categoryBatchQueueProcessor->setAffectedProductIds(self::$affectedProductIds);
```
cammonro, line 47: Would it make sense to just relocate this to the `\Algolia\AlgoliaSearch\Service\Category\BatchQueueProcessor` with a static setter/getter? Is there any benefit to shuffling around objects in this way? 

Also any reason to limit this check to only `executeRow`?

Service/Category/BatchQueueProcessor.php
```diff
@@ -138,4 +138,13 @@ protected function processFullReindex($storeId, $categoriesPerPage)
             );
         }
     }
+
+    /**
+     * @param array $affectedProductIds
+     * @return void
+     */
+    public function setAffectedProductIds(array $affectedProductIds): void
```
cammonro, line 146: Having a getter/setter is much cleaner and follows good OOP principles. 👍 

I'm wondering if we should have the operation push ids onto an array rather than simply set the values. That seems safer so that one operation doesn't potentially overwrite the other?

Service/Category/BatchQueueProcessor.php
```diff
@@ -43,7 +44,9 @@ public function processBatch(int $storeId, ?array $entityIds = null): void
             return;
         }
 
-        $this->rebuildAffectedProducts($storeId);
+        if (count($this->affectedProductIds) > 0) {
+            $this->rebuildAffectedProducts($storeId);
```
cammonro, line 48: Do the `affectedProductIds` get reset after this? Also do we have to maintain two separate sets of IDs? In both this class and the class property in the `Algolia\AlgoliaSearch\Model\Indexer\Category`? If so a single source of truth might be better...

cammonro (changes requested):
Really like the shift to the abstraction of the `BatchQueueProcessor` and the use of a mutator method. I might suggest improving our encapsulation more along this vein and centralizing our source of truth by maintaining a single set of affected product ids that are also threadsafe (and having the source of truth live wherever it makes the most sense). 

Part of me wonders if by indexing products in the category `BatchQueueProcessor` we are violating SRP a little 🤔  - but these are things we can continue to iteratively improve. 

Also don't hate me but I think we need some integration tests for this... 😅 

And now I run for the door... 
🚪 🏃‍♂️

Model/Indexer/Category.php
damcou: I'm not fully sure if it's meaningful on a code POV indeed. For me this works because the products ids are added to the category indexer which is injected in the `CategoryObserver` so that we can make sure those are not "lost in the process" . I guess if we want to shunt the category indexer, we can inject directly the `BatchQueueProcessor` into the observer since this is the only place where this logic exists, I can try that.

cammonro (approved):
A MUCH better approach! Great job! 👏 
Let's ship it. 📦

PR #1718, opened 2025-04-01
MAGE-1218 Replica management enhancements - data patch fix / tests

cammonro:
**Summary**

The PR aims to add robustness to handling replica operations when the API behaves unexpectedly. 

- Added replica specific delete method to `ReplicaManager` with an optional pre-validation step to ensure indices are not still in an attached state before attempting deletion
- Enforced additional pre validation for data patch calls to delete operations
- Refactored patch to be non blocking to CI/CD
- Added a retry mechanism to the data patch to address potential latency and/or accuracy issues with underlying PHP API client `waitForTask` operations
- Revised enablement logic to take into account stores without backend render active but still using InstantSearch 
- Added unit tests to ReplicaManager
- Added new and refactored existing integration tests for applying data patches for replicas and verifying edge cases
- Implemented mock `ReplicaManager` objects for simulating failed operations 
- Added DI proxies to address occasional integration test setup failures

**Result**

Integration test results:
image

Unit test results:
image

### DIFF
--- a/Service/Product/ReplicaManager.php
+++ b/Service/Product/ReplicaManager.php
@@ -470,7 +470,7 @@ class ReplicaManager implements ReplicaManagerInterface
      */
     public function isReplicaSyncEnabled(int $storeId): bool
     {
-        return $this->configHelper->isInstantEnabled($storeId);
+        return $this->configHelper->isInstantEnabled($storeId) && $this->configHelper->isEnabledBackend($storeId);
     }
 
     /**
--- a/Service/Product/ReplicaManager.php
+++ b/Service/Product/ReplicaManager.php
@@ -41,6 +41,7 @@ use Magento\Store\Model\StoreManagerInterface;
 class ReplicaManager implements ReplicaManagerInterface
 {
     public const ALGOLIA_SETTINGS_KEY_REPLICAS = 'replicas';
+    public const ALGOLIA_SETTINGS_KEY_PRIMARY = 'primary';
 
     protected const _DEBUG = true;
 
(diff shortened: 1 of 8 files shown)


Test/Unit/Service/ReplicaManagerTestable.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Test\Unit\Service;
```
damcou, line 3: WDYT about adopting the habit to create a `Testable` subdirectory in each directory so we don't mix "real tests" with those utility classes ? Feels like we will be quickly overwhelmed otherwise.

Service/Product/ReplicaManager.php
```diff
@@ -354,21 +355,74 @@ protected function getBareIndexNameFromReplicaSetting(string $replicaSetting): s
...
+     * @throws AlgoliaException
+     */
+    protected function detachReplica(string $primaryIndexName, string $replicaIndexName): void
+    {
+        $settings = $this->algoliaHelper->getSettings($primaryIndexName);
+        if (!isset($settings[self::ALGOLIA_SETTINGS_KEY_REPLICAS])) {
+            return;
+        }
+        $newReplicas = $this->removeReplicaFromReplicaSetting($settings[self::ALGOLIA_SETTINGS_KEY_REPLICAS], $replicaIndexName);
+        $this->algoliaHelper->setSettings($primaryIndexName, [ self::ALGOLIA_SETTINGS_KEY_REPLICAS => $newReplicas]);
```
damcou, line 406: Is there a way to perform only one `setSettings`/`waitLastTask` in the process and deleting all replicas all at once ?

damcou (approved):
Great job ! 🚀 
Unit and Integration tests are passing !
Had two small comments but nothing blocking, you can merge if you like to move on !

Test/Unit/Service/ReplicaManagerTestable.php
cammonro: We could - I've been trying to stick to a naming convention of `Test` vs `Testable`. I wasn't troubled by pairs of these classes in the directory but it might be personal preference. Having it in the same directory is one less namespace import I guess but that's not a big deal. 

If we end up sharing Testables across tests I would definitely agree - but I can't imagine testing private/protected methods outside of a unit test. Integration tests should only ever test public methods IMO unless you need to mock an object to create some edge case but I haven't had need to import a Testable for that yet.

PR #1719, opened 2025-04-03
MAGE-1222 scoped currency codes

cammonro:
**Summary**

This PR adds scoped currency codes to the facet build operations. Previously we were relying on `Magento\Directory\Model\Currency::getConfigAllowCurrencies` which retrieves all currencies regardless of scope which unnecessarily inflates the facet configuration. 

**Result**

Before fix:
image

After fix:
image

Facet config shows only two currencies:
image

Unit test results:
image

### DIFF
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -1190,15 +1190,30 @@ class ConfigHelper
     }
 
     /**
-     * @param $storeId
-     * @return string
      * @throws Magento\Framework\Exception\NoSuchEntityException
      */
-    public function getCurrencyCode($storeId = null)
+    public function getCurrencyCode(int $storeId = null): string
     {
         /** @var Magento\Store\Model\Store $store */
         $store = $this->storeManager->getStore($storeId);
         return $store->getCurrentCurrencyCode();
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


Helper/ConfigHelper.php
```diff
@@ -1483,6 +1498,8 @@ public function showSuggestionsOnNoResultsPage($storeId = null)
     }
 
     /**
+     * NOTE: This method is currently only used in integration tests and was removed from the general implementation
```
damcou, line 1501: Out of curiosity, when was it removed from the general implementation ? :)

cammonro: I tracked it down here: https://github.com/algolia/algoliasearch-magento-2/pull/1376/files 

I was initially looking at it because it also harvests all possible currencies but realized it wasn't used anymore. My main concern is that by removing it do we expose "hidden" product attributes in certain scenarios. Need to do more investigation here.

damcou (approved):
LGTM 🚀

Helper/ConfigHelper.php
damcou: Oh I see, it was only used to generate secured api key... 🤔

PR #1721, opened 2025-04-03
MAGE-1226: Update indexing integration tests

damcou:
This PR contains:

- Added Proxy declaration in di.xml to avoid installation issue and provide ability to run the integration tests.
- Update all indexing integration tests so they use the introduced `BatchQueueProcessor` classes instead of the legacy indexers..
- Removed a test in the `ConfigTest` class which was testing the handling of the `categories` attribute. The newly created `FacetBuilder` class has now an applicative cache which can't be flushed, so changing the facets "on the fly" cannot be tested as before and resulted as an error. This could be removed because the `categories` attribute handling tests are now located in the `FacetBuilderTest` unit test. 
- All integration tests located in the `Test/Integration/Indexing` folder are now passing:
image

### DIFF
--- a/Test/Integration/Indexing/IndexingTestCase.php
+++ b/Test/Integration/Indexing/IndexingTestCase.php
@@ -2,6 +2,7 @@
 
 namespace Algolia\AlgoliaSearch\Test\Integration\Indexing;
 
+use Algolia\AlgoliaSearch\Api\Processor\BatchQueueProcessorInterface;
 use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
 use Algolia\AlgoliaSearch\Test\Integration\TestCase;
 use Magento\Framework\Indexer\ActionInterface;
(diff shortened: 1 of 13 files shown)


Test/Integration/Indexing/IndexingTestCase.php
```diff
@@ -15,16 +18,43 @@ protected function setUp(): void
...
+                $this->createMock(OutputInterface::class)
+            ]
+        );
         $this->algoliaHelper->waitLastTask();
 
-        $resultsDefault = $this->algoliaHelper->query($this->indexPrefix . 'default_' . $indexSuffix, '', []);
+        $this->assertNumberofHits($indexSuffix, $expectedNbHits);
+    }
 
+    protected function assertNumberofHits($indexSuffix, $expectedNbHits)
```
cammonro, line 55: Love self documenting method names 😄

cammonro (approved):
Great job! Question about the facet cache - do you think it would be worthwhile to provide a means to update on the fly like the integration test used to do?

damcou:
> Great job! Question about the facet cache - do you think it would be worthwhile to provide a means to update on the fly like the integration test used to do?

That's what I was wondering before removing the test... I guess adding a way to clean the applicative cache of the `FacetBuilder` class wouldn't hurt but at the same time this setting is not supposed to be updated several times at once, so I'm not sure it would be that useful except for this particular integration test.

PR #1722, opened 2025-04-07
MAGE-1228 Fix error on invalid creds when using manual SKU indexer

cammonro:
**Summary**
 Fix for ugly error when creds aren't confgured.

**Result**
Before:

After:
image
image


damcou (approved):
Great job 🚀

PR #1723, opened 2025-04-07
MAGE-1228 Fix error on invalid creds when using manual SKU indexer (3.14.x) 

cammonro:
Backport fix from https://github.com/algolia/algoliasearch-magento-2/pull/1722 for v3.14.x.


damcou (approved):
LGTM

PR #1727, opened 2025-04-09
MAGE-347: Configurable cronjob for the indexing queue

damcou:
This PR contains:
- Added a configurable cronjob so that the indexing queue is now handled by the native `bin/magento cron:run` command
- Updated configuration for the indexing queue, adding possibility to provide a specific cron expression.
- Added the related `Cron/ProcessQueue` class which is called each time the magento cron runs and finds a `algolia_queue_process` job to process.
- Removed old queue comment class which asked to the user to add the job manually to their cronjob
- Added a unit test


Cron/ProcessQueue.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function __construct(
+        protected ConfigHelper $configHelper,
+        protected Queue $queue,
+        protected AlgoliaCredentialsManager $algoliaCredentialsManager
+    ) {}
+
+    public function execute()
+    {
+        if (!$this->configHelper->isQueueIndexerEnabled()) {
```
cammonro, line 19: I'm wondering if we should add another check here to bridge the transition from legacy extra crontab to using our new built-in cron? Maybe we could introduce a flag that says "use built-in cron" or something like that? The reason I think it's important is so people can upgrade without making any changes. (Otherwise they can end up with two running crons if they don't remember to remove the old entry from their server crontab.)  

It might make sense Initially to make it so that the "built in cron" is _disabled_ by default for 3.16 and _enabled_ by default for 3.17.

Model/Config/EnableQueueComment.php
```diff
@@ -1,21 +0,0 @@
...
-{
-    public function __construct(
-        protected UrlInterface $urlInterface
-    ) { }
-
-    public function getCommentText($elementValue)
-    {
-        $url = $this->urlInterface->getUrl('algolia_algoliasearch/queue/index');
-
-        return 'If enabled, all indexing operations (add, remove & update operations) will be done asynchronously using the CRON mechanism.<br><br/> Go to <a href="' . $url . '">Indexing Queue</a>. <br/> <br/>To schedule the run you need to add this in your crontab:<br>
```
cammonro, line 18: I understand this comment is out of data but I'm thinking it still might be useful to provide a link to the indexing queue from this page no? I don't believe that is achievable in straight XML in Magento.

Test/Unit/Model/Backend/QueueCronTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $registry = $this->createMock(Registry::class);
+        $config = $this->createMock(ScopeConfigInterface::class);
+        $cacheTypeList = $this->createMock(TypeListInterface::class);
+
+        $this->queueCronModel = new QueueCron($context, $registry, $config, $cacheTypeList);
+    }
+
+    public function testInput(): void
+    {
+        $valuesToTest = [
```
cammonro, line 33: I like all of these test cases! I think it might be cleaner to introduce these as a [separate data provider](https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers).

etc/adminhtml/system.xml
```diff
@@ -958,8 +958,23 @@
...
                     <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                     <comment>
-                        <model>Algolia\AlgoliaSearch\Model\Config\EnableQueueComment</model>
+                        <![CDATA[
+                            If enabled, all indexing operations (add, remove & update operations) will be done asynchronously using the CRON mechanism.<br><br/>
+                            <span class="algolia-config-warning">&#9888;</span> Enabling this option is recommended in production or if your store has a lot of products.
+                        ]]>
+                    </comment>
+                </field>
+                <field id="cron_expr" translate="label comment" type="text" sortOrder="20" showInDefault="1" showInWebsite="0" showInStore="0">
```
cammonro, line 967: Perhaps we can make this field dependent on a Yes/No flag that enables/disabled "built in cron" per my earlier comment. We could make it so that leaving this blank disables the cron as well but that might break your validation and perhaps be less intuitive?

cammonro (changes requested):
Great job! This is long overdue! 🎉 
Added a few observations.

Cron/ProcessQueue.php
damcou: That a totally good point ! Adding it.

etc/adminhtml/system.xml
```diff
@@ -961,6 +961,31 @@
                         <model>Algolia\AlgoliaSearch\Model\Config\EnableQueueComment</model>
                     </comment>
                 </field>
+                <field id="use_built_in_cron" translate="label comment" type="select" sortOrder="15" showInDefault="1" showInWebsite="0" showInStore="0">
+                    <label>Use Magento built-in Cron</label>
```
cammonro, line 965:
```suggestion
                    <label>Use Magento built-in cron</label>
```

Model/Config/EnableQueueComment.php
```diff
@@ -15,7 +15,7 @@ public function getCommentText($elementValue)
     {
         $url = $this->urlInterface->getUrl('algolia_algoliasearch/queue/index');
 
-        return 'If enabled, all indexing operations (add, remove & update operations) will be done asynchronously using the CRON mechanism.<br><br/> Go to <a href="' . $url . '">Indexing Queue</a>. <br/> <br/>To schedule the run you need to add this in your crontab:<br>
-<code>*/5 * * * * php /absolute/path/to/magento/bin/magento indexer:reindex algolia_queue_runner</code> <br/><br/> <span class="algolia-config-warning">&#9888;</span> Enabling this option is recommended in production or if your store has a lot of products. ';
+        return 'If enabled, all indexing operations (add, remove & update operations) will be done asynchronously using the CRON mechanism.<br><br/> Go to <a href="' . $url . '">Indexing Queue</a>. <br/>
```
cammonro, line 18:
```suggestion
        return 'If enabled, all indexing operations (add, remove & update operations) will be done asynchronously using the cron mechanism.<br><br/> Go to <a href="' . $url . '">Indexing Queue</a>. <br/>
```

etc/crontab.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Cron:etc/crontab.xsd">
+    <group id="default">
```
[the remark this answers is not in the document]
cammonro: @damcou Just a reminder if we want to support a [custom group](https://github.com/algolia/algoliasearch-magento-2/pull/1727/commits/dbc869bbb8fc2f09555de5f1218842922ac84a83) configuration we are also going to need a `cron_groups.xml` file. ([Commerce docs](https://experienceleague.adobe.com/en/docs/commerce-operations/configuration-guide/crons/custom-cron-tutorial#step-7-optional-set-up-a-custom-cron-group))

PR #1731, opened 2025-04-10
MAGE-1131: Added unit tests for IndexingManager/Reindex controller

damcou:
This PR contains:
- Unit tests for `IndexingManager/Reindex` controller

### DIFF
--- a/Test/Unit/Controller/Adminhtml/IndexingManager/ReindexTest.php
+++ b/Test/Unit/Controller/Adminhtml/IndexingManager/ReindexTest.php
@@ -1,0 +1,247 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Test\Unit\Controller\Adminhtml\IndexingManager;
+
+use Algolia\AlgoliaSearch\Helper\ConfigHelper;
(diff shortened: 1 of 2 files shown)


Test/Integration/Indexing/IndexingTestCase.php
```diff
@@ -31,6 +32,16 @@ protected function processTest(
         $this->assertNumberofHits($indexSuffix, $expectedNbHits);
     }
 
+    protected function processOldIndexerTest(ActionInterface $indexer, $indexSuffix, $expectedNbHits)
+    {
+        $this->algoliaHelper->clearIndex($this->indexPrefix . 'default_' . $indexSuffix);
+
+        $indexer->executeFull();
+        $this->algoliaHelper->waitLastTask();
```
cammonro, line 40: Shouldn't we be removing references to `AlgoliaHelper` for 3.16?

cammonro (approved):
Great job! Seems like a reasonable set of checks to me and I like the use of data providers. 🙂  

We should strip out any references to `AlgoliaHelper` though - unless we are doing that on a separate task?

PR #1735, opened 2025-04-15
MAGE-1259 Remove InstantSearch enablement dependency in Magento admin

cammonro:
**Summary**

In version 3.14, in an effort to enhance UX and make configuration impact clearer, we introduced Magento admin config dependencies on InstantSearch elements such as "Faceting" and "Sorting". A side effect of this change is the way Magento handles hidden elements that use the backend model: `\Magento\Config\Model\Config\Backend\Serialized`. If InstantSearch is disabled, complex configuration like Faceting and Sorting will be lost. This may be undesirable behavior when experimenting with Algolia admin configurations. For this reason we are removing all dependencies on `algoliasearch_instant/instant/is_instant_enabled` in version 3.16. 

**Result**

### DIFF
--- a/etc/adminhtml/system.xml
+++ b/etc/adminhtml/system.xml
@@ -362,9 +362,6 @@
                             DOM selector of block in which the search results page will be displayed using JavaScript.
                         ]]>
                     </comment>
-                    <depends>
-                        <field id="is_instant_enabled">1</field>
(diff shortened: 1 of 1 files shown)


damcou (approved):
LGTM 🚀

PR #1732, opened 2025-04-10
MAGE-984 Autocomplete redirects with Algolia merch rules

cammonro:
**Summary**

This PR introduces redirects with Algolia merchandising rules (on top of earlier [work for dynamic facets](https://github.com/algolia/algoliasearch-magento-2/pull/1710)).

- Upgrades Autocomplete library to v 1.18.1 and adds sourcemap for debugging
- Installs the [Autocomplete redirect plugin](https://www.algolia.com/doc/ui-libraries/autocomplete/api-reference/autocomplete-plugin-redirect-url/createRedirectUrlPlugin/) 
- Groups all Autocomplete UI libs in a single directory
- Removal of jQuery DOM manipulation for Autocomplete desktop layout in favor of CSS Grid
- Removed legacy references from storefront facing CSS and reorganized by area of concern
- Fixed bug with footer render on empty product results in Autocomplete
- Reduced dependence on jQuery
- Provided [default `render` implementation](https://www.algolia.com/doc/ui-libraries/autocomplete/api-reference/autocomplete-js/autocomplete/#param-render) in Autocomplete options
- Added handling of Autocomplete render with a single source
- Modified products grid from CSS columns to CSS grid to support vertical scrolling 
- Implemented redirect hit (in addition to onSubmit handling)

**Result**

✅ Desktop full grid:
image

✅ Single source render (products):
image

❌  Before (without single source render):
image

✅ Single source render (categories):
image

✅ Selectable redirect hit:
image

✅ Mobile render:
image

🎥 Demo of redirect in action:
https://github.com/user-attachments/assets/e7697a07-1a19-435d-92fe-a6d2e3f3eebf

PR #1738, opened 2025-04-16
MAGE-985 InstantSearch redirects

cammonro:
**Summary**

This PR introduces out-of-the-box redirect support with InstantSearch on Magento - intended to be complimentary to Autocomplete redirects added on https://github.com/algolia/algoliasearch-magento-2/pull/1732.

- This functionality interacts with the `searchBox` widget but that widget is not required.
- It handles the following use cases:
  - When typing a matching query in the `searchBox` widget it will present a selectable link to the redirect. (This was to avoid potentially jarring behavior when redirecting with search as you type)
  - When InstantSearch matches a route it will automatically forward to any URL redirect. 
- The behavior will be made configurable on a future ticket
- Interactivity state is monitored via the containing `uiComponent`
- `instantsearch.v3.css` renamed to `instantsearch.css` (Not sure of original intent here but we've been on InstantSearch v4 since v3 of the Magento extension.)

**Result**

Sample rule:
image

Interactive display:
image

Demo: 
https://github.com/user-attachments/assets/b066be5e-b95a-4671-b5a5-097101b7769d


view/frontend/web/js/instantsearch.js
```diff
@@ -1052,6 +1061,26 @@ define([
...
+            return {
+                render({ results }) {
+                    let content = '';
+                    if (results && results.renderingContent) {
+                        if (results.renderingContent.redirect) {
+                            if (!uiComponnent.hasInteracted) {
+                                window.location.href = results.renderingContent.redirect.url;
+                            }
+                            content += `<div class="instant-redirect">`;
+                            content += `<a href="${results.renderingContent.redirect.url}">${algoliaConfig.translations.redirectSearchPrompt} "${results.query}"</a>`;
```
damcou, line 1075: For an absolute URL, I would expect to have a `target="_blank"` attribute here but I'm not sure what is Algolia's stance regarding this aspect of the feature. (Could be added in the future config maybe ?)  Not blocking anyway.

damcou (approved):
Good job ! 🚀 
Just added a small idea ;)

view/frontend/web/js/instantsearch.js
cammonro: I wondered this as well. Right now the merch rules offer just a URL option. I think it's a good to provide a configurable option in the meantime for the target in the Magento side (dependent on the desired UX for the storefront - since you can redirect to both internal and external links).

PR #1739, opened 2025-04-17
MSI compatibility while being backward compatible

thomas-kl1:
**Summary**

These changes fix the MSI compatibility. It also make this module unnecessary: https://github.com/algolia/algoliasearch-inventory-magento-2

algolia/algoliasearch-inventory-magento-2 is not really maintained anymore and many flaws exists as it does not really follow the changes introduced in the main module.
I've also noticed dead code.

The `$product->getIsSalable()` method use the alias added by MSI while it also fallback to simple inventory management when MSI is disabled.

I've also updated how the stock_qty is retrieved so it used the method that MSI have plugged, so if MSI is used, it's automatically compatible.

### DIFF
--- a/Service/Product/RecordBuilder.php
+++ b/Service/Product/RecordBuilder.php
@@ -268,8 +268,7 @@ class RecordBuilder implements RecordBuilderInterface
     public function addInStock($defaultData, $customData, Product $product)
     {
         if (isset($defaultData['in_stock']) === false) {
-            $stockItem = $this->stockRegistry->getStockItem($product->getId());
-            $customData['in_stock'] = $product->isSaleable() && $stockItem->getIsInStock();
+            $customData['in_stock'] = $this->productIsInStock($product, $product->getStoreId());
         }
 
         return $customData;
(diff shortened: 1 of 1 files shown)


cammonro:
Thank you for your contribution @thomas-kl1 ! It has been on the roadmap to retire the need for a separate module for MSI since it has long been the default in Magento 2.4. We'll be sure to review and incorporate your proposed changes in an upcoming release!

thomas-kl1:
@cammonro things would be less complicated if the changes I suggest were approved. It's a minimal change for a greater support overall

cammonro:
@thomas-kl1 

> @cammonro things would be less complicated if the changes I suggest were approved. It's a minimal change for a greater support overall

Thanks for your thoughts on this. 

A couple of things:

1. The patch we are proposing on the inventory module is immediately shippable to our customers whereas the core module has other constraints.
2. We are still intending to incorporate changes in the core module so these fixes are not mutually exclusive.
 
I also have a concern with your suggested approach as it relies on \Magento\InventoryCatalog\Plugin\CatalogInventory\Model\ResourceModel\Stock\Status\AdaptAddStockDataToCollectionPlugin::aroundAddStockDataToCollection to enforce MSI and we are still retaining deprecated APIs. 

While the change is elegant, I personally prefer that we be more explicit here - as code that appears to use legacy logic like the `Magento_CatalogInventory` module (but is secretly MSI-aware via plugins) is in IMO ambiguous, misleading even, and could lead to maintenance issues down the road.

thomas-kl1:
Your concern are legitimate and I share them. However we didn't see any major changes in the Magento core and we don't expect any soon. Actually many parts of the Magento code is based on this similar mechanic. I don't see Adobe doing this refactor as it's really big. That's why I ended up doing this, would work for both cases (MSI or not) even if it use a ghost plugin (intended by MSI package to add coverage support on vanilla features)

cammonro:
Fair points! 

Meanwhile, accurate stock quantities by warehouse were also a concern which the [referenced PR](https://github.com/algolia/algoliasearch-inventory-magento-2/pull/31) aims to address.

Please note although we don't have a timeline we do intend to revisit this in the main module and your feedback will be incorporated. Marking this on hold for now. Thanks again for your help! We really appreciate it!

PR #1750, opened 2025-05-06
MAGE-1201 fix recommend model validation

cammonro:
**Summary**

This PR:
- Adds friendlier error messaging for 404 errors including "no index found"
- Removes dependency on `renderingContent` - not a good measure of valid model
- Adds soft fail warning on missing hits on configured model
- Fixes validation on Looking Similar 

TODO:
- Address store scoped model validation 

**Result**

image

### DIFF
--- a/Model/RecommendManagement.php
+++ b/Model/RecommendManagement.php
@@ -75,7 +75,7 @@ class RecommendManagement implements RecommendManagementInterface
      */
     public function getLookingSimilarRecommendation(string $productId): array
     {
-        return $this->getRecommendations($productId, 'bought-together');
+        return $this->getRecommendations($productId, 'looking-similar');
     }
 
     /**
--- a/Observer/RecommendSettings.php
+++ b/Observer/RecommendSettings.php
@@ -9,9 +9,11 @@ use Magento\Catalog\Api\ProductRepositoryInterface;
 use Magento\Catalog\Model\Product\Visibility;
 use Magento\Framework\Api\SearchCriteriaBuilder;
 use Magento\Framework\App\Config\Storage\WriterInterface;
+use Magento\Framework\App\State;
 use Magento\Framework\Event\Observer;
 use Magento\Framework\Event\ObserverInterface;
 use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Message\ManagerInterface as MessageManagerInterface;
 
 class RecommendSettings implements ObserverInterface
 {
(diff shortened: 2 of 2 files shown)


Observer/RecommendSettings.php
```diff
@@ -154,12 +159,46 @@ protected function validateRecommendation(string $changedPath, string $recommend
...
 
+    /**
+     * If API does not return a hits response the model may not be configured correctly.
+     * Do not hard fail but alert the end user.
+     * @throws LocalizedException
+     */
+    protected function shouldDisplayWarning(array $recommendationResponse): bool
+    {
+        return
+            $this->appState->getAreaCode() === \Magento\Framework\App\Area::AREA_ADMINHTML
```
damcou, line 176: Why do we need to check this ? Any reasons why the area code could be different ?

Model/RecommendManagement.php
```diff
@@ -75,7 +75,7 @@ public function getTrendingItemsRecommendation(): array
      */
     public function getLookingSimilarRecommendation(string $productId): array
     {
-        return $this->getRecommendations($productId, 'bought-together');
+        return $this->getRecommendations($productId, 'looking-similar');
```
damcou, line 78: 🙄

Observer/RecommendSettings.php
```diff
@@ -143,9 +147,10 @@ protected function validateRecommendation(string $changedPath, string $recommend
     {
         try {
             $recommendations = $this->recommendManagement->$recommendationMethod($this->getProductId());
-            if (empty($recommendations['renderingContent'])) {
-                throw new LocalizedException(__(
-                    "It appears that there is no trained model available for Algolia application ID %1.",
+            if ($this->shouldDisplayWarning($recommendations)) {
+                $this->messageManager->addWarningMessage(__(
```
damcou, line 151: It's way better like this, I guess the models can take a while to be trained so displaying a warning instead of an error makes definitely more sense.

damcou (approved):
LGTM ✅ 
As discussed on JIRA, this is still not ideal and needs additional love, but at least this should unblock some users.

Observer/RecommendSettings.php
cammonro: The event observed by this class is dispatched by `Magento\Config\Model\Config::save` so it's possible it might not be triggered from the admin panel (perhaps CLI scripts, web API's, other customizations etc.). I wanted to play it safe so as not to introduce any unwanted side effects.

Model/RecommendManagement.php
cammonro: LOL

Observer/RecommendSettings.php
damcou: ok makes sense :)

cammonro:
@damcou Thanks for the feedback here! Until we can implement proper store scoping on this validation I've gone and removed the hard fail across the board and switched to issuing warnings only.

PR #1753, opened 2025-05-06
MAGE-1201 Concrete injectable logger

cammonro:
**Summary**

This PR is a proposed change while implementing https://github.com/algolia/algoliasearch-magento-2/pull/1752 that switches from a virtual type logger to a concrete logger which can be referenced in code (unlike the previous virtual type). It is intended as a drop in replacement for `Psr\Log\LoggerInterface` and to provide a simple way to localize Algolia specific logging without needing to inject the heavier `DiagnosticsLogger` or `TimedLogger`.

**Result**

### DIFF
--- a/Api/LoggerInterface.php
+++ b/Api/LoggerInterface.php
@@ -1,0 +1,4 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Api;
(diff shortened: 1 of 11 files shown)


cammonro:
Moving to separate ticket on https://github.com/algolia/algoliasearch-magento-2/pull/1756

PR #1756, opened 2025-05-07
MAGE-1272 Concrete injectable logger

cammonro:
**Summary**

This PR is a proposed change while implementing https://github.com/algolia/algoliasearch-magento-2/pull/1752 that switches from a virtual type logger to a concrete logger which can be referenced in code (unlike the previous virtual type). It is intended as a drop in replacement for `Psr\Log\LoggerInterface` and to provide a simple way to localize Algolia specific logging without needing to inject the heavier `DiagnosticsLogger` or `TimedLogger`.

**Result**

### DIFF
--- a/Test/Unit/Logger/AlgoliaLoggerTest.php
+++ b/Test/Unit/Logger/AlgoliaLoggerTest.php
@@ -10,9 +10,9 @@ use Psr\Log\LoggerInterface;
 
 class AlgoliaLoggerTest extends TestCase
 {
-    protected LoggerInterface $algoliaLogger;
-    protected SystemLoggerHandler $systemLoggerHandler;
-    protected AlgoliaLoggerHandler $algoliaLoggerHandler;
+    protected ?LoggerInterface $algoliaLogger;
+    protected ?SystemLoggerHandler $systemLoggerHandler;
+    protected ?AlgoliaLoggerHandler $algoliaLoggerHandler;
 
     protected function setUp(): void
     {
--- a/Test/Unit/Logger/AlgoliaLoggerTest.php
+++ b/Test/Unit/Logger/AlgoliaLoggerTest.php
@@ -2,9 +2,9 @@
 
 namespace Algolia\AlgoliaSearch\Test\Unit\Logger;
 
+use Algolia\AlgoliaSearch\Logger\AlgoliaLogger;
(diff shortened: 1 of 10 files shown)


Test/Unit/Logger/AlgoliaLoggerTest.php
```diff
@@ -19,7 +19,7 @@ protected function setUp(): void
         $this->systemLoggerHandler = $this->createMock(SystemLoggerHandler::class);
         $this->algoliaLoggerHandler = $this->createMock(AlgoliaLoggerHandler::class);
 
-        $this->algoliaLogger = new Logger(
+        $this->algoliaLogger = new AlgoliaLogger(
```
damcou, line 22: This is passing, but I still need to set the class attributes as nullable to make it work 😅

etc/di.xml
```diff
@@ -252,24 +252,14 @@
     </type>
     <!-- custom indexers END -->
 
-    <virtualType name="Algolia\AlgoliaSearch\Logger\AlgoliaLogger" type="Monolog\Logger">
+    <type name="Algolia\AlgoliaSearch\Logger\AlgoliaLogger">
```
damcou, line 255: I checked if no proxy was needed here 😄 with a quick integration test, this looks ok ✅

damcou (approved):
This looks alright to me ! 🚀 

I saw that there are still some occurrences of `Psr\Log\LoggerInterface` in the codebase, should we keep them as they are ?

Test/Unit/Logger/AlgoliaLoggerTest.php
cammonro: Ah! Yes I need to remember to make all of these nullable going forward. I've gone and tweaked the unit test for this.

cammonro:
Thanks for the review!

> I saw that there are still some occurrences of `Psr\Log\LoggerInterface` in the codebase, should we keep them as they are ?

There are times where I think we want the logger behavior to perform "as is" (for classes that extend `\Magento\Framework\Model\ResourceModel\Db\Collection\AbstractCollection` for instance) so I decided to leave those - but we can always revisit. 

My goal here was to ensure that we can discretely log activities directly related to the extension without polluting the main logs and that it could be done easily without needlessly using DI.

PR #1757, opened 2025-05-20
MAGE-1260: Remove deprecated methods/classes for 3.16

damcou:
WIP


Helper/Entity/ProductHelper.php
```diff
@@ -440,7 +417,7 @@ protected function getCompositeTypes(): array
      * @param Product $product
      * @return mixed
      *
-     * @deprecated (will be removed in v3.16.0)
+     * @deprecated (will be removed in once MSI compatibility module will be added to this module)
```
cammonro, line 420: Good call out. 👍

Service/AdditionalSection/IndexBuilder.php
```diff
@@ -66,25 +75,23 @@ public function buildIndex(int $storeId, ?array $entityIds, ?array $options): vo
 
             $indexName = $this->additionalSectionHelper->getIndexName($storeId);
             $indexName = $indexName . '_' . $section['name'];
+            $tempIndexName = $indexName . IndexNameFetcher::INDEX_TEMP_SUFFIX;
 
             $attributeValues = $this->additionalSectionHelper->getAttributeValues($storeId, $section);
 
-            $tempIndexName = $indexName . IndexNameFetcher::INDEX_TEMP_SUFFIX;
+            $indexOptions = $this->indexOptionsBuilder->buildWithEnforcedIndex($indexName, $storeId);
```
cammonro, line 82: I think this code would also benefit from an implementation of `EntityIndexOptionsBuilderInterface`.

Service/AlgoliaConnector.php
```diff
@@ -311,13 +311,13 @@ function ($id) {
...
-     * @param IndexOptions $fromIndexOptions
-     * @param IndexOptions $toIndexOptions
+     * @param IndexOptionsInterface $fromIndexOptions
+     * @param IndexOptionsInterface $toIndexOptions
      * @return void
      * @throws AlgoliaException
      * @throws NoSuchEntityException
      */
-    public function moveIndex(IndexOptions $fromIndexOptions, IndexOptions $toIndexOptions): void
+    public function moveIndex(IndexOptionsInterface $fromIndexOptions, IndexOptionsInterface $toIndexOptions): void
```
cammonro, line 320: Interfaces are always preferred - good choice!

Service/Category/IndexBuilder.php
```diff
@@ -140,20 +151,20 @@ protected function buildIndexPage($storeId, $collection, $page, $pageSize, $cate
         }
         $collection->setCurPage($page)->setPageSize($pageSize);
         $collection->load();
-        $indexName = $this->categoryHelper->getIndexName($storeId);
+        $indexOptions = $this->indexOptionsBuilder->buildWithComputedIndex(CategoryHelper::INDEX_NAME_SUFFIX, $storeId);
```
cammonro, line 154: Another instance where `EntityIndexOptionsBuilderInterface` can be used.

Service/Page/IndexBuilder.php
```diff
@@ -61,7 +69,7 @@ public function buildIndex($storeId, ?array $entityIds, ?array $options): void
             return;
         }
 
-        $indexName = $this->pageHelper->getIndexName($storeId);
+        $indexOptions = $this->indexOptionsBuilder->buildWithComputedIndex(PageHelper::INDEX_NAME_SUFFIX, $storeId);
```
cammonro, line 72: Can use an `EntityIndexOptionsBuilderInterface` implementation for page entities.

Service/Product/BackendSearch.php
```diff
@@ -59,7 +60,8 @@ public function getSearchResult(string $query, int $storeId, ?array $searchParam
             $params = array_merge($params, $searchParams);
         }
 
-        $response = $this->algoliaHelper->query($indexName, $query, $params, $storeId);
+        $indexOptions = $this->indexOptionsBuilder->buildWithEnforcedIndex($indexName, $storeId);
```
cammonro, line 63: In general I would prefer we use computed index options as much as possible.

Service/Suggestion/IndexBuilder.php
```diff
@@ -102,19 +113,22 @@ protected function rebuildStoreSuggestionIndexPage(int $storeId, QueryCollection
         $collection = clone $collectionDefault;
         $collection->setCurPage($page)->setPageSize($pageSize);
         $collection->load();
-        $indexName = $this->suggestionHelper->getTempIndexName($storeId);
+        $indexOptions = $this->indexOptionsBuilder->buildWithComputedIndex(
```
cammonro, line 116: Can use an `EntityIndexOptionsBuilderInterface` implementation for suggestion entities.

Test/Integration/AssertValues/Magento24CE.php
```diff
@@ -9,7 +9,7 @@ abstract class Magento24CE
     public $productsCountWithoutGiftcards = 181;
     public $lastJobDataSize = 13;
     public $expectedCategory = 16;
-    public $attributesForFaceting = 5;
+    public $attributesForFaceting = 6;
```
cammonro, line 12: Is this related to recent dynamic facet changes?

Test/Integration/Indexing/Config/ConfigTest.php
```diff
@@ -136,17 +137,18 @@ private function replicaCreationTest($withCustomerGroups = false)
...
+        $indices = $this->algoliaConnector->listIndexes();
         $indicesNames = array_map(function ($indexData) {
             return $indexData['name'];
         }, $indices['items']);
 
         foreach ($sortingIndicesWithRankingWhichShouldBeCreated as $indexName => $firstRanking) {
             $this->assertContains($indexName, $indicesNames);
 
-            $settings = $this->algoliaHelper->getSettings($indexName);
+            $indexOptions = $this->indexOptionsBuilder->buildWithEnforcedIndex($indexName);
```
cammonro, line 150: Can we use computed index options here?

Test/Integration/Indexing/Config/MultiStoreConfigTest.php
```diff
@@ -88,15 +88,17 @@ public function testMultiStoreIndicesCreation()
 
         $this->indicesConfigurator->saveConfigurationToAlgolia($fixtureSecondStore->getId());
 
-        $defaultCategoryIndexSettings = $this->algoliaHelper->getSettings(
+        $defaultCategoryIndexOptions = $this->indexOptionsBuilder->buildWithEnforcedIndex(
```
cammonro, line 91: Can we compute here as well? Or would that not work with these fixtures?

cammonro (changes requested):
Great work! Thanks for this huge effort. You're a hero! 🦸‍♂️

I added a few thoughts but in an effort to be thorough realized I was getting redundant especially once I hit the testing classes. 

My main feedback is really that we should just lean towards computing indices as much as possible and use entity specific builders to separate the logic even further. 

If there's any chance we can clean that up then I think this looks good!

Helper/AlgoliaHelper.php
cammonro, line 1: Can we consider retaining this class until 3.17 in case anyone is still using it?

Block/Algolia.php
```diff
@@ -107,9 +107,9 @@ public function getCatalogSearchHelper(): CatalogSearchHelper
         return $this->catalogSearchHelper;
     }
 
-    public function getAlgoliaHelper(): AlgoliaHelper
+    public function getAlgoliaConnector(): AlgoliaConnector
     {
-        return $this->algoliaHelper;
+        return $this->algoliaConnector;
```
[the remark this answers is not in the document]
damcou: From what I can see, its only purpose is to be used in the monolithic `Configuration::getConfiguration()` method. I wouldn't be against refactoring it a little but I guess this would need an additional ticket 😅

Helper/AlgoliaHelper.php
damcou: Just brought back the class with a new deprecation comment.

Test/Integration/Indexing/Config/MultiStoreConfigTest.php
damcou: I think the rework of index names in the integration tests should be done on a separate ticket since it needs some tidying, I've created a ticket for it (MAGE-1298)

cammonro (changes requested):
Thank you Damien! Looking good. I just had a concern about the public method change on `ProductHelper`. If we can fix that I think we're ready to go. I also agree with your rationale on updating the tests under a separate ticket.

Helper/Entity/ProductHelper.php
```diff
@@ -314,66 +301,70 @@ public function getIndexSettings(?int $storeId = null): array
...
      * @param bool $saveToTmpIndicesToo
      * @return void
      * @throws AlgoliaException
      * @throws LocalizedException
      * @throws NoSuchEntityException
      */
-    public function setSettings(string $indexName, string $indexNameTmp, int $storeId, bool $saveToTmpIndicesToo = false): void
-    {
+    public function setSettings(
+        IndexOptionsInterface $indexOptions,
```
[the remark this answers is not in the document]
damcou: Sorry I don't understand how we can keep the same signature without using an enforced name in the method (maybe I misunderstood the original feedback : https://github.com/algolia/algoliasearch-magento-2/pull/1757#discussion_r2102439334)

I feel that it's ok to change method signatures on a minor release (we can specify it in the changelogs)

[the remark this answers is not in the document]
damcou: Changelog updated ! :)

PR #1758, opened 2025-05-23
MAGE-1205: Fix Revenue analytics' discount calculation with configured VAT

damcou:
This PR contains:
- A fix for revenue calculation on `place order` events where discounts weren't properly calculated in case website has a setup with configured VAT. 
- Units tests for this precise feature.


cammonro (approved):
I'm good with a configuration aware approach here. Let's follow up with integration tests to make sure we're not missing anything. 🙂

PR #1761, opened 2025-05-30
MAGE-1270: Move indexing configuration to the indexing manager

damcou:
This PR contains:
- Moved indexing configuration from "Credentials and Basic setup" section to the newly created "Indexing Manager" section
- Renamed "Enable Backend" occurrences to "Enable Indexing" to be more meaningful and to avoid confusion with "backend rendering" statements in the extension.
- Fixed tests according to these changes
- Added a data patch to ensure the configuration doesn't get lost in the process.

### DIFF
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -21,9 +21,6 @@ use Magento\Store\Model\StoreManagerInterface;
 class ConfigHelper
 {
     public const ENABLE_FRONTEND = 'algoliasearch_credentials/credentials/enable_frontend';
-    public const ENABLE_BACKEND = 'algoliasearch_credentials/credentials/enable_backend';
-    public const ENABLE_QUERY_SUGGESTIONS_INDEX = 'algoliasearch_credentials/credentials/enable_query_suggestions_index';
-    public const ENABLE_PAGES_INDEX = 'algoliasearch_credentials/credentials/enable_pages_index';
     public const LOGGING_ENABLED = 'algoliasearch_credentials/credentials/debug';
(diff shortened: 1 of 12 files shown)


Helper/ConfigHelper.php
```diff
@@ -21,9 +21,6 @@
 class ConfigHelper
 {
     public const ENABLE_FRONTEND = 'algoliasearch_credentials/credentials/enable_frontend';
-    public const ENABLE_BACKEND = 'algoliasearch_credentials/credentials/enable_backend';
```
cammonro, line 24: Can we add these to the "Deprecated Constants & Methods" section?

Helper/ConfigHelper.php
```diff
@@ -247,9 +247,9 @@ public function isEnabledFrontEnd($storeId = null)
      * @param $storeId
      * @return bool
      */
-    public function isEnabledBackend($storeId = null)
```
cammonro, line 250: Can we have this method call the new `isIndexingEnabled` method and then move it to the deprecated section?

etc/adminhtml/system.xml
```diff
@@ -53,43 +53,6 @@
                         ]]>
                     </comment>
                 </field>
-                <field id="enable_backend" translate="label comment" type="select" sortOrder="40" showInDefault="1" showInWebsite="1" showInStore="1">
-                    <label>Enable Indexing</label>
```
cammonro, line 57: I'm wondering if we should add a notice to this section that says that the indexing settings have been relocated to the Indexing Manager tab?

PR #1762, opened 2025-06-05
MAGE-1269: Change default behaviour for new full indexing feature

damcou:
This PR contains:
- Changed default configuration for full indexing (keeps the legacy Magento indexers enabled)
- Changed wording in the config section
- Added a note in the changelogs to suggest to disable them


etc/adminhtml/system.xml
```diff
@@ -1093,9 +1093,9 @@
                 <label>Full indexing via Magento indexers</label>
                 <comment>
                     <![CDATA[
-                         <p><span class="algolia-config-warning">&#9888;</span> Since version 3.16.0, the following Magento indexers are now disabled by default for <strong>full indexing</strong>.
-                         However the extension continues to rely on them for reindexing specific entities with Magento materialized views or entity save events.</p>
-                         <p>If you want to perform full reindexing, we highly suggest to keep this configuration as it is and run the newly created CLI commands instead.</p>
```
cammonro, line 1098: Can we add a comment that this will become the default in a future version?

PR #1764, opened 2025-06-10
MAGE-1083 Collection size caching

cammonro:
**Summary**

This PR aims to reduce overhead related to calculating collection size. As catalogs grow in size the query becomes increasingly less performant. By eliminating redundant `COUNT()` operations we can shave some processing time here.

This PR includes the following changes: 

- Removed `rebuildEntityIds()` as a relic of `Algolia\AlgoliaSearch\Helper\Data::rebuildStoreProductIndex()`[here](https://github.com/algolia/algoliasearch-magento-2/blob/3.14.5/Helper/Data.php#L369) 
  -  It is not necessary to re-page an already chunked array of `entityIds`
  - Product set on the delta will never be higher than the page size max because of potential parent aggregation for configurables etc.
- Removed collection cloning - was only needed for in-process collection paging on deltas which we are no longer doing
- Removed collection reset ops for memory management - collection use is now limited to function scope
- Paging information is now included on deltas for observability - previously logs misreported page 1 for all jobs on a large delta

- Removed `areParentsLoaded` member boolean - this internal state was observed to be problematic when processing jobs across stores (parents would not be loaded for subsequent stores) - now should be consistent

- Moved emulation to encompass entire collection build to ensure full alignment with storefront behavior.
- Introduced an optional cache to store collection size in Redis
- Added a plugin to purge the collection size when products are added or deleted or edited in a way that impacts visibility on the storefront
- Added unit tests

Please note that I chose not to address all stock scenarios for cache invalidation until we more properly address MSI. That can be reevaluated at a later date. 

**Result**

Unit tests: 
image

Caching of collection size in Redis: 
image

New indexing cache:

Full index job gen - EXTRA LARGE fixture - uncached:
image

Full index job gen - EXTRA LARGE fixture - CACHED:
image


Model/Cache/IndexCollectionSize.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Model\Cache;
+
+use Algolia\AlgoliaSearch\Model\Cache\Type\Indexer;
+use Magento\Framework\App\Cache\StateInterface;
+use Magento\Framework\App\Cache\TypeListInterface;
+use Magento\Framework\App\CacheInterface;
+
+class IndexCollectionSize
```
damcou: If we plan to use this type of cache for categories as well in the future, maybe this is worth adding a Product subdirectory for this ?

Model/Cache/Type/Indexer.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Model\Cache\Type;
+
+use Magento\Framework\App\Cache\Type\FrontendPool;
+use Magento\Framework\Cache\Frontend\Decorator\TagScope;
+
+class Indexer extends TagScope {
+    const TYPE_IDENTIFIER = 'algolia_indexer';
```
damcou, line 9: Same here, rename it to `algolia_product_index` so we can use `algolia_category_index` later ?

etc/di.xml
```diff
@@ -2,6 +2,7 @@
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
     <type name="Magento\Catalog\Model\ResourceModel\Product">
         <plugin name="algoliaProducts" type="Algolia\AlgoliaSearch\Model\Indexer\ProductObserver"/>
+        <plugin name="algoliaProductCacheClean" type="Algolia\AlgoliaSearch\Plugin\Cache\CacheCleanProductPlugin"/>
```
damcou, line 5: I'm not fully sure that relying on product events to clean this cache is enough. Does it cover mass actions from product grid for example ? 
We'll probably have some issues with direct DB changes from ERP I guess. 
I think that at least we should warn the user to make sure to clean the cache after any action like this ? Or maybe we should set a short TTL ?

Service/Product/IndexBuilder.php
```diff
@@ -86,71 +87,23 @@ public function buildIndexList(int $storeId, array $entityIds = null, array $opt
...
         $this->buildIndexPage(
             $storeId,
             $collection,
-            $options['page'],
-            $options['pageSize'],
-            null,
-            $options['entityIds'],
-            $options['useTmpIndex']
+            $options['page'] ?? 1,
+            $options['pageSize'] ?? $this->configHelper->getNumberOfElementByPage(),
```
damcou: Should we store scope `getNumberOfElementByPage()` here  ?

Service/Product/BatchQueueProcessor.php
```diff
@@ -39,60 +40,104 @@ public function __construct(
...
-                /** @uses ProductIndexBuilder::buildIndexList() */
-                $this->queue->addToQueue(
-                    ProductIndexBuilder::class,
-                    'buildIndexList',
-                    ['storeId' => $storeId, 'entityIds' => $chunk],
-                    count($chunk)
-                );
-            }
-
+        if (!empty($entityIds)) {
```
damcou, line 54: Nice usage of short protected methods here, this is way more readable ! :)

etc/cache.xml
```diff
@@ -0,0 +1,6 @@
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Cache/etc/cache.xsd">
+    <type name="algolia_indexer" translate="label,description" instance="Algolia\AlgoliaSearch\Model\Cache\Type\Indexer">
```
damcou, line 2: The cache was disabled by default when I setup upgraded with the branch, is there a way to make it enabled by default ?

Test/Unit/Service/Product/BatchQueueProcessorTest.php
```diff
@@ -0,0 +1,281 @@
...
+{
+    protected ?Data $dataHelper;
+    protected ?ConfigHelper $configHelper;
+    protected ?ProductHelper $productHelper;
+    protected ?Queue $queue;
+    protected ?DiagnosticsLogger $diag;
+    protected ?AlgoliaCredentialsManager $algoliaCredentialsManager;
+    protected ?IndexBuilder $indexBuilder;
+    protected ?IndexCollectionSize $indexCollectionSizeCache;
+    protected BatchQueueProcessor $processor;
```
damcou:
```suggestion
    protected ?BatchQueueProcessor $processor;
```

Still have my good old nullable issue :D

Service/Product/BatchQueueProcessor.php
```diff
@@ -9,25 +9,26 @@
...
 use Algolia\AlgoliaSearch\Model\IndicesConfigurator;
 use Algolia\AlgoliaSearch\Model\Queue;
 use Algolia\AlgoliaSearch\Service\AlgoliaCredentialsManager;
 use Algolia\AlgoliaSearch\Service\Product\IndexBuilder as ProductIndexBuilder;
+use Magento\Catalog\Model\ResourceModel\Product\Collection;
 use Magento\Framework\Exception\NoSuchEntityException;
 
 class BatchQueueProcessor implements BatchQueueProcessorInterface
 {
-    protected bool $areParentsLoaded = false;
```
damcou, line 21: I guess this should have been an array of values with all different stores so we don't have the issue you're mentionning in the PR's description no ?

Service/Product/BatchQueueProcessor.php
```diff
@@ -39,60 +40,104 @@ public function __construct(
...
+        foreach (array_chunk($entityIds, $productsPerPage) as $i => $chunk) {
+            /** @uses ProductIndexBuilder::buildIndexList() */
+            $this->queue->addToQueue(
+                ProductIndexBuilder::class,
+                'buildIndexList',
+                [
+                    'storeId'   => $storeId,
+                    'entityIds' => $chunk,
+                    'options'   => [
+                        'page'        => $i + 1,
```
damcou, line 118: Adding `page` and `pageSize` could be confusing with full indexing jobs, do you feel this is useful/mandatory ?

Model/Cache/IndexCollectionSize.php
cammonro: Yes agree. This is a very good idea. Even if we don't need this kind of optimization right away I like the idea of preemptively structuring for it.

Model/Cache/Type/Indexer.php
cammonro: I like idea of distinguishing - but for now I want to keep under a single manageable cache in the front end. I think I will alter the Redis key save so that can go in the cache model rather than the type.

etc/di.xml
cammonro: No you're right. It's probably not sufficient but I figured this can be a starting point. I need to verify on mass action. I know it worked on mass action delete but there are other scenarios. I'm also concerned about stock events but I would like to reassess when we address MSI.

Service/Product/IndexBuilder.php
cammonro: Yes! Good idea.

etc/cache.xml
cammonro: 🤔  I know that `app/etc/env.php` controls this behavior but as to how to set the default I'll need to inspect how Magento does that in its core implementation. I also wasn't sure whether I wanted to enable by default due to invalidation scenarios. If a catalog is really dynamic (i.e. the size changes frequently) this cache probably won't help much.

Test/Unit/Service/Product/BatchQueueProcessorTest.php
cammonro: 🤦 Oops! That's a miss on my part.

Service/Product/BatchQueueProcessor.php
cammonro: Yes that's possible but I also couldn't see any benefit to trying to manage this since it was only executing once in my tests. Is there a scenario where you've seen otherwise? If so then yes it makes sense to do a keyed array here.

Service/Product/BatchQueueProcessor.php
cammonro: Well the invoked class name should distinguish the job type and (as of right now) we don't explicitly specify IDs in the job for full reindex (which might change as we've discussed). My goal here was improving observability. If a large change set of IDs occurs that exceeds the page size limit and we _don't_ indicate the page then there's no way of knowing that a job was part of a single change event. It will instead look like a separate event (as page 1).

etc/di.xml
cammonro: Actually I think I'm going to add a configurable TTL here. Then the user can set it to whatever they want...

cammonro:
Thanks for the great feedback @damcou ! I will implement the changes you've proposed (though some might go on a different ticket).

Model/Cache/Type/Indexer.php
damcou: Ok makes sense :)

etc/cache.xml
damcou: Enabling this by default is not mandatory for use, at least we'll need to this to our documentation so the users know where to enable it if needed.

Service/Product/BatchQueueProcessor.php
damcou: No you're right, I think this is another relic that can be removed.

Service/Product/BatchQueueProcessor.php
damcou: Ok I see, as long as there's an easy way to distinguish deltas and full indexings, I'm good with it !

etc/cache.xml
cammonro: For sure - I'll add a documentation task to follow up here.

etc/di.xml
cammonro: Created a new task for this

damcou (approved):
Tests are passing, LGTM 🚀

PR #1769, opened 2025-06-23
Fix store id for queue job

pikulsky:
**Summary**
The “algoliasearch_queue” table keeps job entries. Job's "data" field is a JSON string with job details.
Starting from 3.6.0 the field names in "data" JSON got a new format: from snake case they were changed to camel case:

How job was added to the table before version 3.6.0: field names with snake case:
https://github.com/algolia/algoliasearch-magento-2/blob/338a9433d9a89e98dca7ad316c284186c897baa9/Model/Indexer/Product.php#L85
```
  $this->queue->addToQueue(
      Data::class,
      'rebuildStoreProductIndex',
      ['store_id' => $storeId, 'product_ids' => $chunk],
      count($chunk)
  );
```
in 3.6.0 the format was changed to camel case:
https://github.com/algolia/algoliasearch-magento-2/blob/5554a79abd04e93331ffcf1d7769270253da0a90/Model/Indexer/Product.php#L85
```
  $this->queue->addToQueue(
      Data::class,
      'rebuildStoreProductIndex',
      ['storeId' => $storeId, 'productIds' => $chunk],
      count($chunk)
  );
```

But there are two places where code works with the old snake case format:
* while loading a job record and decoding "data" values
* while sorting jobs

### DIFF
--- a/Model/Job.php
+++ b/Model/Job.php
@@ -94,8 +94,8 @@ class Job extends \Magento\Framework\Model\AbstractModel implements JobInterface
 
             $this->setDecodedData($decodedData);
 
-            if (isset($decodedData['store_id'])) {
-                $this->setStoreId($decodedData['store_id']);
+            if (isset($decodedData['storeId'])) {
+                $this->setStoreId($decodedData['storeId']);
             }
         }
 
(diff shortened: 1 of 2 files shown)


damcou:
Thanks a lot for your contribution ! 
Your fix will be included in our next releases (3.15.2, 3.16.1 and 3.17.0)

PR #1773, opened 2025-06-26
MAGE-1341 Fix dropped products during full reindex via queue

cammonro:
**Summary**

This PR addresses an issue where products were written to the wrong index during a full reindex with indexing queue enabled. 

This issue only affects the 3.16.0 beta and was identified during testing.

**Result**

Before:
image
After:
image


cammonro:
Resubmitting on separate ticket.

PR #1779, opened 2025-07-01
MAGE-1138: Configurable temporary index usage

damcou:
This PR contains:
- Added "Use temporary index for full products reindex" config to the indexing queue section
- Updated to product batch processing to take this new config into
- Added a small improvement to the Indexing Queue Data column so it now indicates clearly if a temporary index is used in the job or not.
- Added an integration test which ensure that no temporary index is created in case this config is set to "No"
- Recent change on this PR for 3.17.0 introduced `$tmpTmpIndex` to `Product/IndexBuilder::indexPage()` . This changes worked for 3.16.0 but another PR for 3.17.0 removed the parameter: https://github.com/algolia/algoliasearch-magento-2/pull/1764/files#diff-0251b76d832bc7f8f1205481909f582d196147fa77188441b39a86a7885ac3a3L107 => This PR also fixes this issue so it works for 3.17.0 as well. 

WIP

### DIFF
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -5,6 +5,7 @@ namespace Algolia\AlgoliaSearch\Helper;
 use Algolia\AlgoliaSearch\Api\Product\ReplicaManagerInterface;
 use Algolia\AlgoliaSearch\Helper\Configuration\AutocompleteHelper;
 use Algolia\AlgoliaSearch\Helper\Configuration\InstantSearchHelper;
+use Algolia\AlgoliaSearch\Helper\Configuration\QueueHelper;
 use Algolia\AlgoliaSearch\Service\AlgoliaConnector;
 use Algolia\AlgoliaSearch\Service\Serializer;
 use Magento\Cookie\Helper\Cookie as CookieHelper;
@@ -41,11 +42,6 @@ class ConfigHelper
     public const SHOW_CATS_NOT_INCLUDED_IN_NAV = 'algoliasearch_categories/categories/show_cats_not_included_in_navigation';
     public const INDEX_EMPTY_CATEGORIES = 'algoliasearch_categories/categories/index_empty_categories';
     public const CATEGORY_SEPARATOR = 'algoliasearch_categories/categories/category_separator';
-
-    public const IS_ACTIVE = 'algoliasearch_queue/queue/active';
-    public const USE_BUILT_IN_CRON = 'algoliasearch_queue/queue/use_built_in_cron';
-    public const NUMBER_OF_JOB_TO_RUN = 'algoliasearch_queue/queue/number_of_job_to_run';
-    public const RETRY_LIMIT = 'algoliasearch_queue/queue/number_of_retries';
 
     public const XML_PATH_IMAGE_WIDTH = 'algoliasearch_images/image/width';
     public const XML_PATH_IMAGE_HEIGHT = 'algoliasearch_images/image/height';
@@ -156,7 +152,8 @@ class ConfigHelper
         protected GroupExcludedWebsiteRepositoryInterface               $groupExcludedWebsiteRepository,
         protected CookieHelper                                          $cookieHelper,
         protected AutocompleteHelper                                    $autocompleteConfig,
-        protected InstantSearchHelper                                   $instantSearchConfig
+        protected InstantSearchHelper                                   $instantSearchConfig,
+        protected QueueHelper                                           $queueHelper
     )
     {}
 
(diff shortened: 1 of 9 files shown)


Helper/ConfigHelper.php
```diff
@@ -156,7 +152,8 @@ public function __construct(
         protected GroupExcludedWebsiteRepositoryInterface               $groupExcludedWebsiteRepository,
         protected CookieHelper                                          $cookieHelper,
         protected AutocompleteHelper                                    $autocompleteConfig,
-        protected InstantSearchHelper                                   $instantSearchConfig
+        protected InstantSearchHelper                                   $instantSearchConfig,
+        protected QueueHelper                                           $queueHelper
```
cammonro, line 156: Thanks for taking this opportunity to further break down this monolithic helper class!

Helper/Configuration/QueueHelper.php
```diff
@@ -0,0 +1,67 @@
...
+    public function getRetryLimit($storeId = null)
+    {
+        return (int)$this->configInterface->getValue(self::RETRY_LIMIT, ScopeInterface::SCOPE_STORE, $storeId);
+    }
+
+    /**
+     * @param $storeId
+     * @return bool
+     */
+    public function useTmpIndex($storeId = null)
```
cammonro, line 62: Good abstraction of the temp index logic here. 👍

Service/Product/IndexBuilder.php
```diff
@@ -101,7 +101,8 @@ public function buildIndex(int $storeId, ?array $entityIds, ?array $options): vo
             $collection,
             $options['page'] ?? 1,
             $options['pageSize'] ?? $this->configHelper->getNumberOfElementByPage($storeId),
-            $entityIds
+            $entityIds,
+            $options['useTmpIndex'] ?? false
```
cammonro, line 105: Thank you for catching the temp index param! 🙏 Now I remember why I removed it. At the time it wasn’t used in the code but I should have caught this when porting the temp index fix from 3.16 to 3.17.

Setup/Patch/Schema/ConfigPatch.php
```diff
@@ -68,6 +68,7 @@ class ConfigPatch implements SchemaPatchInterface
         'algoliasearch_queue/queue/active' => '0',
         'algoliasearch_queue/queue/number_of_job_to_run' => '5',
         'algoliasearch_queue/queue/number_of_retries' => '3',
+        'algoliasearch_queue/queue/use_tmp_index' => '1',
```
cammonro, line 71: Now that we are more consistently using `config.xml` is it still necessary to update this old data patch?

Test/Integration/Indexing/Queue/QueueTest.php
```diff
@@ -136,16 +137,57 @@ public function testExecute()
...
+        $indices = $this->algoliaConnector->listIndexes();
+
+        $existsDefaultTmpIndex = false;
+        foreach ($indices['items'] as $index) {
+            if ($index['name'] === $this->indexPrefix . 'default_products_tmp') {
+                $existsDefaultTmpIndex = true;
+            }
+        }
+        // Checking if the temporary index hasn't been created
+        $this->assertFalse($existsDefaultTmpIndex, 'Temporary index exists and it should not');
```
cammonro, line 177: Solid! 😎

etc/adminhtml/system.xml
```diff
@@ -1038,6 +1038,23 @@
...
+                    <depends>
+                        <field id="active">1</field>
+                    </depends>
+                </field>
+                <field id="use_tmp_index" translate="label comment" type="select" sortOrder="50" showInDefault="1" showInWebsite="0" showInStore="0">
+                    <label>Use a temporary index for full products reindex</label>
+                    <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
+                    <comment>
+                        <![CDATA[
+                            This configuration defines if a temporary index suffixed by <code>_tmp</code> will be used for products full reindex. If you set it to "No", product updates will be sent directly to your production index.
```
cammonro, line 1055: Should we put a placeholder here linking to the official docs on why you may or may not want to do this?

cammonro (approved):
Great job! 👏 

Had a couple of questions but nothing blocking. Approved!

Setup/Patch/Schema/ConfigPatch.php
damcou: Maybe not, but it doesn't hurt I guess :D

etc/adminhtml/system.xml
damcou: I guess we can add the link afterwards indeed.

PR #1781, opened 2025-07-09
MAGE-1112: Make `price` attribute optional during product indexing

damcou:
This PR contains:
- Make the `price` attribute optional => now it shouldn't be indexed if the price attribute is **not** present in the **product attributes**, the **sortings**, the **facets** and the **custom rankings** lists.
- Display a notice in the admin when the user removes the `price` attribute from the attributes list but not from another list (which would make the `price` attribute still indexed)
- Updated the product indexing integration test to ensure this new logic is applied.

### DIFF
--- a/Helper/Configuration/NoticeHelper.php
+++ b/Helper/Configuration/NoticeHelper.php
@@ -23,6 +23,7 @@ class NoticeHelper extends \Magento\Framework\App\Helper\AbstractHelper
         'getRecommendNotice',
         'getCookieConfigurationNotice',
         'getMultiApplicationIDsNotice',
+        'getPriceIndexingNotice',
     ];
 
     /** @var array[] */
@@ -189,8 +190,8 @@ class NoticeHelper extends \Magento\Framework\App\Helper\AbstractHelper
         ];
     }
 
-   protected function getCookieConfigurationNotice()
-   {
+    protected function getCookieConfigurationNotice()
+    {
         $noticeContent = '';
         $selector = '';
         $method = 'after';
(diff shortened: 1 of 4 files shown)


Helper/Configuration/NoticeHelper.php
```diff
@@ -363,4 +364,32 @@ protected function getMultiApplicationIDsNotice(): void
...
+
+        // we want to display the warning only if price is not present in the attribute list but is present somewhere else
+        if (!($this->configHelper->isAttributeInList($attributesToIndex, 'price')
+            && !$this->configHelper->isAttributeInList($attributesList, 'price'))
+        ) {
+            return;
+        }
+
+        $noticeTitle = 'Price attribute indexing';
+        $noticeContent = '<p>Price attribute has been removed from the product attributes list but is still present in the facets, sortings or custom rankings lists.</p>
```
cammonro, line 386: Based on the logic, couldn't this warning be raised if the opposite was true? i.e. added as a facet but not a product attribute?

cammonro (commented):
Looking good to me so far. I just had a question about our validation logic for displaying user warnings.

Helper/Configuration/NoticeHelper.php
damcou: To me, removing the price attribute from the product attributes list marks the user's intent not to see it in the records anymore. 
On the other hand, I don't see the case where user would like to use it as a facet without having the will for it to be present in the product's attribute. 

Not sure if I'm clear here, but I think the most important thing is to warn the user that there's an misconfiguration between the two lists for the price attribute specifically since this is what this is about in this ticket. 

...

I guess I can add an additional sentence in case the user wants it to be indexed.

cammonro: Thanks for this clarification @damcou - on second look at the boolean logic I think we're actually good!

cammonro (approved):
🚀

PR #1782, opened 2025-07-10
MAGE-991: Update CircleCI build jobs

omar-rahman-dev:
**Summary**

- Update build process for CircleCI to setup magento enterprise + matrix build the extension for magento 2.4.6 and 2.4.7 on PHP 8.2

**Result**
image

image

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
           name: Configure composer authentication
           command: |
             composer config --global http-basic.repo.magento.com ${MAGENTO_AUTH_USERNAME} ${MAGENTO_AUTH_PASSWORD}
-            composer config repositories.algolia vcs https://github.com/algolia/algoliasearch-magento-2
+            composer config --global repositories.algolia vcs https://github.com/algolia/algoliasearch-magento-2
       - run:
           name: Setup folder structure and override file
           command: |
@@ -43,22 +43,13 @@ jobs:
           working_directory: ~/Sites
           command: |
             curl -s https://raw.githubusercontent.com/markshust/docker-magento/master/lib/template | bash
-            docker compose -f compose.yaml -f compose.override.yml up -d --remove-orphans 
+            docker compose -f compose.yaml -f compose.override.yml up -d --remove-orphans
+            bin/cli rm -rf composer.lock vendor composer.json
             bin/fixowns
             bin/setup-composer-auth
             bin/cli git clone git@github.com:magento/magento2.git .
             bin/cli git checkout tags/<< parameters.magento-version >>
-            bin/composer require "algolia/algoliasearch-magento-2:${CIRCLE_BRANCH}"
-            bin/composer install
-      # - run:
-      #     name: Configure AlgoliaSearch extension
-      #     working_directory: ~/Sites
-      #     command: |
-      #       cp -r ~/project/ AlgoliaSearch
-      #       bin/cli mkdir -p app/code/Algolia
-      #       docker cp AlgoliaSearch/ "$(bin/docker-compose ps -q phpfpm|awk '{print $1}')":/var/www/html/app/code/Algolia
-      #       bin/fixowns
-      #       bin/fixperms
+            bin/composer require "algolia/algoliasearch-magento-2:dev-${CIRCLE_BRANCH}"
       - run:
           name: Enable AlgoliaSearch extension
           working_directory: ~/Sites


.circleci/config.yml
```diff
@@ -1,112 +1,80 @@
...
+            cat \<<EOF > compose.override.yml
+            services:
+              phpfpm:
+                image: markoshust/magento-php:<< parameters.php-version >>-fpm
+            EOF
+      - run:
+          name: Setup magento enterprise
+          working_directory: ~/Sites
+          command: |
+            curl -s https://raw.githubusercontent.com/markshust/docker-magento/master/lib/template | bash
```
damcou, line 45: I think that having this process working with markshust's setup is already a big win for us \o/ . However I fear that we'll need to make it work with the official Adobe components at the end of the day (see https://hub.docker.com/u/magento).

Not sure how mandatory it is for now for the MVP, I'll let @cammonro give his thoughts about it.

.circleci/config.yml
```diff
@@ -1,112 +1,80 @@
...
+            docker compose -f compose.yaml -f compose.override.yml up -d --remove-orphans 
+            bin/fixowns
+            bin/setup-composer-auth
+            bin/cli git clone git@github.com:magento/magento2.git .
+            bin/cli git checkout tags/<< parameters.magento-version >>-p6
```
damcou: I think the patch version should be part of the variable if possible so we can easily change it.

.circleci/config.yml
```diff
@@ -1,112 +1,80 @@
...
+      - run:
+          name: Enable AlgoliaSearch extension
+          working_directory: ~/Sites
+          command: |
+            bin/magento module:enable Algolia_AlgoliaSearch
+      - run:
+          name: Compile AlgoliaSearch extension
+          working_directory: ~/Sites
+          command: |
+            bin/composer require algolia/algoliasearch-client-php "^4.0" -W
```
damcou: Not sure if it's intended not to install the actual extension for now but the package you're installing is the PHP client, the magento2 extension package is `algolia/algoliasearch-magento-2` .
The PHP client is present in the dependencies of the extension so it will be added automatically.

.circleci/config.yml
```diff
@@ -1,112 +1,80 @@
...
+            bin/setup-composer-auth
+            bin/cli git clone git@github.com:magento/magento2.git .
+            bin/cli git checkout tags/<< parameters.magento-version >>-p6
+            bin/composer install
+      - run:
+          name: Configure AlgoliaSearch extension
+          working_directory: ~/Sites
+          command: |
+            cp -r ~/project/ AlgoliaSearch
+            bin/cli mkdir -p app/code/Algolia
```
damcou: Normally if the extension is installed with composer, the code will be located into the` vendor/algolia `directory, so you don't need to move it. the `app/code/Algolia` directory is more when you're using the github repo to work locally (at least for me).

.circleci/config.yml
omar-rahman-dev: We could have another set of jobs work with the official components if that works? 👀

.circleci/config.yml
omar-rahman-dev: > I think the patch version should be part of the variable if possible so we can easily change it.

It should be possible, do you want p6 as the default?

.circleci/config.yml
omar-rahman-dev: > Normally if the extension is installed with composer, the code will be located into the`vendor/algolia`directory, so you don't need to move it. the `app/code/Algolia` directory is more when you're using the github repo to work locally (at least for me).

Ah that makes sense! I think then it would make sense to remove the composer step and just use the repo (this way changes in PRs are verified and compiled)

.circleci/config.yml
cammonro: I've found the Magento Cloud Docker tooling to be pretty cumbersome and worry it would burn too much compute time. Also OpenSearch tends to fail quite often so my preference would be to possibly revisit when we remove that dependency. However, if we can reliably run our integration tests (and install Algolia via Composer) then we probably can catch almost all (if not all) of the issues that tooling would typically reveal. 

Integration tests are also process intensive as well so I'd imagine those could be performed on merges to the `release/*` branches?

.circleci/config.yml
cammonro: Agree here. 

These are the versions we can use to start:

- [2.4.6-p11](https://experienceleague.adobe.com/en/docs/commerce-operations/release/notes/security-patches/2-4-6-patches?lang=en#p11)
- [2.4.7-p6](https://experienceleague.adobe.com/en/docs/commerce-operations/release/notes/security-patches/2-4-7-patches?lang=en#p6)

.circleci/config.yml
cammonro: Agree - this library should be driven by our extension's dependency requirements rather than be hard coded in the pipeline.

.circleci/config.yml
cammonro: It would be my preference if we could install via Composer as well - and will eliminate the need to explicitly require `algolisearch-client-php` as mentioned.

.circleci/config.yml
omar-rahman-dev: @cammonro correct, its not part of this PR yet (this is a base line) however you can run it on PRs that contain a label such as "run-integration-test" and by default will run on main branch.

cammonro (dismissed):

cammonro (approved):
LGTM!

PR #1783, opened 2025-07-10
MAGE-1363 PHPUnit 10 prep

cammonro:
**Summary**

This PR addresses concerns related to compatibility with PHPUnit 10 in preparation for Magento 2.4.8 as recommended here: https://experienceleague.adobe.com/en/docs/commerce-operations/release/notes/adobe-commerce/2-4-8#platform

**Result**

Unit tests work as expected on PHPUnit 10
image
Integration tests will need to be refactored further for compatibility.

### DIFF
--- a/Test/Integration/Frontend/Category/CategoryCacheTest.php
+++ b/Test/Integration/Frontend/Category/CategoryCacheTest.php
@@ -23,12 +23,12 @@ class CategoryCacheTest extends AbstractController
     public static function getCategoryProvider(): array
     {
         return [
-            ['categoryId' => 20, 'name' => 'Women', 'hasProducts' => false],
-            ['categoryId' => 21, 'name' => 'Women > Tops', 'hasProducts' => true],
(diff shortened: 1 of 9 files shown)


Test/Integration/Frontend/Category/CategoryCacheTest.php
```diff
@@ -23,12 +23,12 @@ class CategoryCacheTest extends AbstractController
     public static function getCategoryProvider(): array
     {
         return [
```
damcou, line 25: Does it mean that we need to change all data providers for PHPunit 10 ?

damcou (approved):
LGTM 🚀 
Additional comments on JIRA

Test/Integration/Frontend/Category/CategoryCacheTest.php
cammonro: I believe I updated all the ones needing attention so we should be good here. 👍

PR #1784, opened 2025-07-14
MAGE-1367: Add unit tests to CircleCI

omar-rahman-dev:
# MAGE-1367: Add unit tests to CircleCI

**Summary**

- Add unit testing steps to CircleCI
- Add unit coverage steps to CircleCI
- Store unit tests results in "test" tab
- Store artifacts available for download

**Result**

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,13 +73,13 @@ jobs:
       - store_test_results:
           path: ~/Sites/unit-tests
       - run:
-          name: Run Unit Coverage
+          name: Run Unit Test Coverage and copy results to CircleCI
           working_directory: ~/Sites
           command: |
-            bin/test/unit-coverage vendor/algolia/algoliasearch-magento-2/Test/Unit
-            docker cp $(docker compose ps -q phpfpm):/var/www/html/dev/tests/unit/report ~/Sites/unit-coverage/
+            bin/cli php -d xdebug.mode=coverage vendor/bin/phpunit --log-junit /var/www/html/dev/tests/unit/report/junit.xml --coverage-html /var/www/html/dev/tests/unit/report --coverage-filter /var/www/html/vendor/algolia/algoliasearch-magento-2/ /var/www/html/vendor/algolia/algoliasearch-magento-2/Test/Unit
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -50,6 +52,7 @@ jobs:
             bin/cli git clone git@github.com:magento/magento2.git .
             bin/cli git checkout tags/<< parameters.magento-version >>
             bin/composer require "algolia/algoliasearch-magento-2:dev-${CIRCLE_BRANCH}"
+            bin/composer global require --dev phpunit/phpunit
```
cammonro, line 55: This appears to install PHPUnit 9 by default based on dependency restrictions I assume. For now this is fine but I'd be curious to see how this performs on Magento 2.4.8 and whether we might need a matrix here.

.circleci/config.yml
```diff
@@ -61,9 +64,26 @@ jobs:
           command: |
             bin/magento setup:di:compile
             bin/magento module:status Algolia_AlgoliaSearch
+      - run:
+          name: Run Unit Tests and copy results to CircleCI
+          working_directory: ~/Sites
+          command: |
+            bin/cli vendor/bin/phpunit --log-junit /var/www/html/dev/tests/unit/report/junit.xml /var/www/html/vendor/algolia/algoliasearch-magento-2/Test/Unit
```
cammonro, line 71: Is this falling back to PHPUnit's default configuration? I'm wondering if we should be explicitly using Magento's `phpunit.xml.dist` (at `dev/tests/unit`)? Can we confirm here?

cammonro:
> This PR looks good to me, I just have a follow-up question (on JIRA). I'm also wondering if your PRs shouldn't be targeting the `release/3.17.0-dev branch` instead of the main one (unless it's mandatory ?) . WDYT @cammonro ?

@damcou I'm actually fine with running these tests on every PR and activating it across the board. But perhaps we can have different matrices depending on what we're targeting? Like run minimal tests when targeting `release/*-dev` and run more comprehensive testing when targeting `release/*` (not `-dev`)? Could that be doable @omar-rahman-dev ?

.circleci/config.yml
omar-rahman-dev: Yes! It should pick that up automatically.

PR #1786, opened 2025-07-18
MAGE-1281 Replica settings forwarding

cammonro:
**Summary**

This PR aims to provide support for forwarding settings to replicas which historically the integration has never done. 

It is enabled by default but can be disabled in the admin under the "Advanced" tab. 

This now enables more reliable function of dynamic facets across alternate sorts (for instance if the primary index `attributesForFaceting` get out of sync with one or more replicas).

It also includes some clean up of `system.xml` and the mega ConfigHelper class (tried to apply some logical grouping here so future additions can be added more sensibly) along with a couple of deprecations. 

**Result**

Unit tests:

image

Rule config:
image

Consistent application of dynamic facet by rule across sorts:

### DIFF
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -308,7 +308,7 @@ class ConfigHelper
     }
 
     // --- Products --- //
-    
+
     /**
      * @param $storeId
      * @return array
@@ -2038,15 +2038,12 @@ class ConfigHelper
     /**
      * @param $storeId
      * @return bool
-     * @deprecated This method will be moved to the InstantSearch config helper
+     * @deprecated This method has been moved to the InstantSearch config helper and will be removed in a future version
+     * @see \Algolia\AlgoliaSearch\Helper\Configuration\InstantSearchHelper::shouldShowSuggestionsOnNoResultsPage()
      */
     public function showSuggestionsOnNoResultsPage($storeId = null)
     {
-        return $this->configInterface->isSetFlag(
-            self::SHOW_SUGGESTIONS_NO_RESULTS,
-            ScopeInterface::SCOPE_STORE,
-            $storeId
-        );
+        return $this->instantSearchConfig->shouldShowSuggestionsOnNoResultsPage($storeId);
     }
 
     /**
--- a/Helper/ConfigHelper.php
+++ b/Helper/ConfigHelper.php
@@ -21,6 +21,7 @@ use Magento\Store\Model\StoreManagerInterface;
 class ConfigHelper
 {
     // --- Credentials & Basic Setup --- //
+
     public const ENABLE_FRONTEND = 'algoliasearch_credentials/credentials/enable_frontend';
     public const LOGGING_ENABLED = 'algoliasearch_credentials/credentials/debug';
     public const APPLICATION_ID = 'algoliasearch_credentials/credentials/application_id';
@@ -32,6 +33,7 @@ class ConfigHelper
     public const ALGOLIA_COOKIE_DURATION = 'algoliasearch_credentials/algolia_cookie_configuration/cookie_duration';
 
     // --- Products --- //
+
     public const PRODUCT_ATTRIBUTES = 'algoliasearch_products/products/product_additional_attributes';
     public const PRODUCT_CUSTOM_RANKING = 'algoliasearch_products/products/custom_ranking_product_attributes';
     public const USE_ADAPTIVE_IMAGE = 'algoliasearch_products/products/use_adaptive_image';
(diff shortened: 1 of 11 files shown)


Helper/ConfigHelper.php
```diff
@@ -20,6 +20,8 @@
 
 class ConfigHelper
```
damcou, line 21: The changes are quite impossible to read, I assume this is just some code reorg which is not well displayed by Github, some forgotten deprecation and the introduction of the `shouldForwardPrimaryIndexSettingsToReplicas` method, no additional logic, right ?

etc/adminhtml/system.xml
```diff
@@ -1518,15 +1522,23 @@
                 <field id="write_timeout" translate="label comment" type="text" sortOrder="110" showInDefault="1">
                     <label>Write Timeout (In Seconds)</label>
                 </field>
+                <field id="forward_to_replicas" translate="label comment" type="select" sortOrder="115" showInDefault="1" showInWebsite="1" showInStore="1">
```
damcou, line 1525: This advanced tab is becoming very messy, maybe we could add a task to reorganize it a little bit, WDYT ?

Service/ReplicaSettingsHandler.php
```diff
@@ -0,0 +1,82 @@
...
+use Algolia\AlgoliaSearch\Api\Data\IndexOptionsInterface;
+use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
+use Algolia\AlgoliaSearch\Helper\ConfigHelper;
+use Magento\Framework\Exception\NoSuchEntityException;
+
+/**
+ *  This class will proxy the settings save operations and forward to replicas based on user configuration
+ *  excluding attributes which should never be forwarded
+ */
+class ReplicaSettingsHandler
```
damcou: Since this class is responsible for applying settings on primary AND replicas, I would assume it to be called SettingsHandler (not blocking though, I'm just picky)

Service/ReplicaSettingsHandler.php
```diff
@@ -0,0 +1,82 @@
...
+            if ($noForward) {
+                $this->connector->setSettings(
+                    $indexOptions,
+                    $noForward,
+                    false,
+                    true,
+                    $mergeSettingsFrom
+                );
+            }
+        } else {
```
damcou: I think this method would be simpler if the `if` condition was inverted and if this was an early return.

Test/Unit/Service/ReplicaSettingsHandlerTest.php
```diff
@@ -0,0 +1,173 @@
...
+        $this->config->method('shouldForwardPrimaryIndexSettingsToReplicas')
+            ->with($storeId)
+            ->willReturn(true);
+
+        $invocations = $this->exactly(2);
+        $this->connector->expects($invocations)
+            ->method('setSettings')
+            ->willReturnCallback(
+                function($indexOptions, $indexSettings, $forwardToReplicas, $mergeSettings, $mergeFrom) use ($invocations) {
+                    switch ($invocations->numberOfInvocations()) {
```
damcou: This doesn't seem to work with PHPUnit 9, don't we want to keep compatibility with it until 2.4.7 support ends ?

damcou (commented):
Nice job 🚀 
Got a few comments, I'll let you share your thoughts about them.

Helper/ConfigHelper.php
cammonro: I'd say that's a good summary. 

But yeah reading in Github is not going to be very beneficial. The commit history shows the sequence of the refactor starting with constants, then methods, then the forgotten deprecation then finally the new config. 

To really evaluate, I would recommend just comparing the class to the actual `system.xml` or UI. 

The goal here was to make sense of this because settings were plopped all over the place with no real pattern - making it hard to find a config without a search or hard to know where to add a new config. Now everything is logically grouped based on where it appears in the config (and should make it easier to split into separate config classes down the road which we started already doing.) I also grouped Magento core config operations. 

In the future if we add a config - I'd like it if we can add it in the helper in the corrresponding position with the actual config.

etc/adminhtml/system.xml
cammonro: I'm all for that. Will spin up in Jira.

Service/ReplicaSettingsHandler.php
cammonro: Yeah I struggled with naming this because SettingsHandler seemed too generic and I wanted to communicate the business use case in that it is specifically needed to handle replicas. What if I called it `ReplicaAwareSettingsHandler`? Do you like that better or too verbose?

Service/ReplicaSettingsHandler.php
cammonro: Yeah good callout. Will tweak that. Thanks!

Test/Unit/Service/ReplicaSettingsHandlerTest.php
cammonro: Yeah this is one of the real pains of migrating to Magento 2.4.8.
`numberOfInvocations()` only exists in PHPUnit 10.
`getInvocationCount()` only exists in PHPUnit 9.
I think I need to track my invocations differently here on this test.

Service/ReplicaSettingsHandler.php
damcou: Yeah, I'm not a huge fan to add more and more words to the name 😅 

What about `IndexSettingsHandlers` ?

Test/Unit/Service/ReplicaSettingsHandlerTest.php
damcou: All green now :)

Service/ReplicaSettingsHandler.php
cammonro: I like it! (But will drop the extra `s`) Going with `IndexSettingsHandler`. Thanks!

PR #1796, opened 2025-08-06
Fix product getUrl method by aligning with Magento core behavior

fasimana:
This commit adds the UrlRewrite::REDIRECT_TYPE => 0 parameter to the product URL filter data, ensuring that only non-redirect URL rewrites are  considered when generating product URLs.
This aligns the Algolia URL generation  with Magento's core URL handling behavior and prevents potential issues with  redirect loops or incorrect URL generation.

**Adding filter redirect_type = 0
This parameter is important because: in the core Magento Catalog/Model/Product/Url.php class, it is included in the filter data, but it's missing in the Algolia extension's version.**

**Without this parameter, the URL finder might return redirect URLs instead of direct URLs, which could lead to redirect chains or incorrect URL generation.**


cammonro:
@fasimana Thanks for calling this out. You're absolutely right and we're going to be retiring this preference in a future release. In the meantime, happy to apply your fix!

PR #1797, opened 2025-08-07
MAGE-1109: Add Batching Optimizer feature

damcou:
This PR contains:
- The Batching Optimizer CLI command

WIP...

### DIFF
--- a/Console/Command/BatchingOptimizerCommand.php
+++ b/Console/Command/BatchingOptimizerCommand.php
@@ -1,0 +1,245 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Console\Command;
+
+use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
+use Algolia\AlgoliaSearch\Model\IndexOptions;
+use Algolia\AlgoliaSearch\Service\AlgoliaConnector;
+use Algolia\AlgoliaSearch\Service\Product\IndexOptionsBuilder;
+use Algolia\AlgoliaSearch\Service\StoreNameFetcher;
+use Magento\Catalog\Model\ResourceModel\Product\CollectionFactory;
+use Magento\Framework\App\State;
+use Magento\Framework\Console\Cli;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Exception\NoSuchEntityException;
+use Magento\Store\Model\StoreManagerInterface;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class BatchingOptimizerCommand extends AbstractStoreCommand
+{
+    /**
+     * @var array|null
+     */
+    protected ?array $indices = null;
+
+    /**
+     * @var array|null
+     */
+    protected ?array $configurablePercentile = [];
+
+    /**
+     * Recommended Max batch size
+     * https://www.algolia.com/doc/guides/sending-and-managing-data/send-and-update-your-data/how-to/sending-records-in-batches/
+     */
+    const int MAX_BATCH_SIZE = 10_000_000;
+
+    /**
+     * Arbitrary default margin to ensure not to exceed recommended batch size
+     */
+    const int DEFAULT_MARGIN = 25;
+
+    /**
+     * Arbitrary increased margin to ensure not to exceed recommended batch size when catalog is a mix between configurables and other product types
+     * (i.e. with a lot of record sizes variations)
+     */
+    const int INCREASED_MARGIN = 50;
+
+    /**
+     * Arbitrary lower boundary where percentile of configurable products is considered "low enough"
+     */
+    const int CONFIGURABLE_PERCENTILE_LOWER_BOUNDARY = 10;
+
+    /**
+     * Arbitrary upper boundary where percentile of configurable products is considered "high enough"
+     */
+    const int CONFIGURABLE_PERCENTILE_UPPER_BOUNDARY = 90;
+
+    public function __construct(
+        protected AlgoliaConnector      $algoliaConnector,
+        protected State                 $state,
+        protected StoreNameFetcher      $storeNameFetcher,
+        protected StoreManagerInterface $storeManager,
+        protected IndexOptionsBuilder   $indexOptionsBuilder,
+        protected CollectionFactory     $productCollectionFactory,
+        ?string                         $name = null
+    ) {
+        parent::__construct($state, $storeNameFetcher, $name);
+    }
+
+    protected function getCommandPrefix(): string
+    {
+        return parent::getCommandPrefix() . 'batching:';
+    }
+
+    protected function getCommandName(): string
+    {
+        return 'optimizer';
+    }
+
+    protected function getCommandDescription(): string
+    {
+        return "Optimizes the batching size configuration under \"Configuration > Algolia Search > Advanced > Indexing Queue > Maximum number of records processed per indexing job\" according to various configurations.";
+    }
+
+    protected function getStoreArgumentDescription(): string
+    {
+        return 'ID(s) for store(s) to optimize (optional), if no store is specified, all stores will be taken into account.';
+    }
+
+    protected function getAdditionalDefinition(): array
+    {
+        return [];
+    }
+
+    /**
+     * @throws NoSuchEntityException|LocalizedException
+     */
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $this->input = $input;
+        $this->output = $output;
+        $this->setAreaCode();
+
+        $storeIds = $this->getStoreIds($input);
+
+        try {
+            $this->optimizeBatchingConfiguration($storeIds);
+        } catch (\Exception $e) {
+            $this->output->writeln('<error>' . $e->getMessage() . '</error>');
+            return CLI::RETURN_FAILURE;
+        }
+
+        return Cli::RETURN_SUCCESS;
+    }
+
+    /**
+     * @param array $storeIds
+     * @return void
+     */
+    protected function optimizeBatchingConfiguration(array $storeIds = []): void
+    {
+        if (count($storeIds)) {
+            foreach ($storeIds as $storeId) {
+                $this->optimizeBatchingForStore($storeId);
+            }
+        } else {
+            $this->optimizeBatchingForAllStores();
+        }
+    }
+
+    /**
+     * @return void
+     */
+    protected function optimizeBatchingForAllStores(): void
+    {
+        $storeIds = array_keys($this->storeManager->getStores());
+
+        foreach ($storeIds as $storeId) {
+            $this->optimizeBatchingForStore($storeId);
+        }
+    }
+
+    /**
+     * @param int $storeId
+     * @return void
+     * @throws AlgoliaException
+     * @throws NoSuchEntityException
+     */
+    protected function optimizeBatchingForStore(int $storeId): void
+    {
+        $indexOptions = $this->indexOptionsBuilder->buildEntityIndexOptions($storeId);
+        $indexData = $this->getIndexData($indexOptions);
+        $configurablePercentile = $this->getConfigurablePercentile($indexData['entries'], $storeId);
+
+        $this->output->writeln('<info> ====== ' . $this->storeNameFetcher->getStoreName($storeId) . ' ====== </info>');
+        $this->output->writeln('<comment>Index</comment>:               ' . $indexOptions->getIndexName());
+        $this->output->writeln('<comment>Number of records</comment>:   ' . $indexData['entries']
+            . ' (' . round($configurablePercentile) . '% of configurable products)');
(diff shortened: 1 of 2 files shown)


Console/Command/BatchingOptimizerCommand.php
```diff
@@ -0,0 +1,245 @@
...
+    /**
+     * @var array|null
+     */
+    protected ?array $configurablePercentile = [];
+
+    /**
+     * Recommended Max batch size
+     * https://www.algolia.com/doc/guides/sending-and-managing-data/send-and-update-your-data/how-to/sending-records-in-batches/
+     */
+    const int MAX_BATCH_SIZE = 10_000_000;
```
cammonro: I prefer the typing here but unfortunately as of 3.16 we [still support PHP 8.2](https://github.com/algolia/algoliasearch-magento-2/blob/3.16.0-rc2/composer.json#L8) to be backward compatible with Magento 2.4.7 and PHP 8.2 doesn't support this syntax. Can we switch to PHPDoc?

Console/Command/BatchingOptimizerCommand.php
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Fetches index data from the Algolia Dashboard
+     *
+     * @param IndexOptions $indexOptions
+     * @return array
+     * @throws AlgoliaException
+     */
+    protected function getIndexData(IndexOptions $indexOptions): array
```
cammonro:
```suggestion
    protected function getIndexData(\Algolia\AlgoliaSearch\Api\Data\IndexOptionsInterface $indexOptions): array
```

... or qualified with import.

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class BatchingOptimizeCommand extends AbstractStoreCommand
+{
+    /**
+     * Recommended Max batch size
+     * https://www.algolia.com/doc/guides/sending-and-managing-data/send-and-update-your-data/how-to/sending-records-in-batches/
+     */
+    const MAX_BATCH_SIZE = 10000000; //10MB
```
cammonro: I actually liked your earlier use of the [underscore numeric separator](https://php.watch/versions/7.4/underscore_numeric_separator). Any reason for abandoning it? I think rather than a comment we could just label it something like `MAX_BATCH_SIZE_IN_BYTES` for clarity.

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+    const DEFAULT_MARGIN = 25;
+
+    /**
+     * Arbitrary increased margin to ensure not to exceed recommended batch size when catalog is a mix between complex and other product types
+     * (i.e. with a lot of record sizes variations)
+     */
+    const INCREASED_MARGIN = 50;
+
+    const DEFAULT_SAMPLE_SIZE = 20;
+    const LARGE_SAMPLE_SIZE = 100;
```
cammonro: I think we should keep the default but allow the user to explicitly set the sample size to whatever they want.

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+    /**
+     * Arbitrary default margin to ensure not to exceed recommended batch size
+     */
+    const DEFAULT_MARGIN = 25;
+
+    /**
+     * Arbitrary increased margin to ensure not to exceed recommended batch size when catalog is a mix between complex and other product types
+     * (i.e. with a lot of record sizes variations)
+     */
+    const INCREASED_MARGIN = 50;
```
cammonro: Just like with sample size I think having an option to set the margin makes sense.

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+    protected function getStoreArgumentDescription(): string
+    {
+        return 'ID(s) for store(s) to optimize (optional), if no store is specified, all stores will be taken into account.';
+    }
+
+    protected function getAdditionalDefinition(): array
+    {
+        return [
+            new InputOption(
+                self::LARGE_SAMPLE_OPTION,
```
cammonro: I would recommend replacing this with two options: margin and sample size

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+        $simpleSampleSize = (int)round($sampleSize * ($this->storeCounts[$storeId]['simple_percentage'] / 100));
+        $complexSampleSize = (int)round($sampleSize * ($this->storeCounts[$storeId]['complex_percentage'] / 100));
+
+        $this->storeCounts[$storeId]['simple_sample_size'] = $simpleSampleSize;
+        $this->storeCounts[$storeId]['complex_sample_size'] = $complexSampleSize;
+
+        $this->storeCounts[$storeId]['sample'] = array_merge(
+            $this->getProductsSizes($simpleProducts, $simpleSampleSize),
+            $this->getProductsSizes($complexProducts, $complexSampleSize)
+        );
```
cammonro, line 348: Could we add a max product object size somehwere here? This stat could serve to inform the absolute floor for batching (for a worst case scenario).

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+    /**
+     * @param array $sizes
+     * @param int $averageSize
+     * @return float
+     */
+    protected function getStandardDeviation(array $sizes, int $averageSize): float
+    {
+        $sum = 0;
+        foreach ($sizes as $size) {
+            $sum += ($size - abs($averageSize)) * ($size - abs($averageSize));
```
cammonro: Can we use `pow` function here instead? I think it would make the formula implementation clearer.

Also why are we applying `abs` -  I think that will throw off the calculations too much?

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+        }
+
+        return round(sqrt($sum / count($sizes)), 2);
```
cammonro: This is actually the calculation for standard deviation on a _population_ - we need standard deviation on a **sample**:
```
s = √(Σ(x - x̄)² / (n-1))
```
```suggestion
        return round(sqrt($sum / (count($sizes) - 1)), 2);
```

Console/Command/BatchingOptimizeCommand.php
cammonro:
```suggestion
            $sum += pow($size - $averageSize, 2);
```

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+        return (int) round(self::MAX_BATCH_SIZE / $averageSize);
+    }
+
+    /**
+     * @param array $sizes
+     * @param int $averageSize
+     * @return float
+     */
+    protected function getStandardDeviation(array $sizes, int $averageSize): float
+    {
```
cammonro: I'm concerned about possible division by zero errors.
Can we guard these methods with something like?
```
if (count($sizes) <= 1) {
   return 0.0;
}
```

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+        return $stats;
+    }
+
+    /**
+     * @param array $sizes
+     * @return int
+     */
+    protected function getSizeAverage(array $sizes): int
+    {
+        return (int) round(array_sum(array_values($sizes)) / count($sizes));
```
cammonro: Possible division by zero errors.

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+        $this->storeCounts[$storeId] = [
+            'simple' => $simpleProducts->count(),
+            'complex' => $complexProducts->count()
+        ];
+
+        $this->storeCounts[$storeId]['total'] =
+            (int) $this->storeCounts[$storeId]['simple'] + (int) $this->storeCounts[$storeId]['complex'];
+
+        $this->storeCounts[$storeId]['simple_percentage'] =
+            ($this->storeCounts[$storeId]['simple'] * 100) / $this->storeCounts[$storeId]['total'];
```
cammonro: Possible division by zero errors.

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+    {
+        $storeName = $this->storeNameFetcher->getStoreName($storeId);
+
+        if (!$this->configHelper->isIndexingEnabled($storeId)) {
+            $this->output->writeln('<info>Indexing is disabled for store ' . $storeName . '</info>');
+            return;
+        }
+
+        if (!isset($this->storeCounts[$storeId])) {
+            $this->setStoreCounts($storeId);
```
cammonro, line 251: This method can throw exceptions but we don't catch or call out - maybe add some `@throws` statements in the calling method?

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,366 @@
...
+        $sizeAverage = $this->getSizeAverage($sample);
+        $this->output->writeln('<comment>Average record size</comment>         : ' . $sizeAverage . 'B');
+
+        $estimatedBatchCount = $this->getEstimatedMaxBatchCount($sizeAverage);
+        $this->output->writeln('<comment>Estimated Max batch count</comment>   : ' . $estimatedBatchCount . ' records');
+
+        $standardDeviation = $this->getStandardDeviation($sample, $sizeAverage);
+        $this->output->writeln('<comment>Standard Deviation</comment>          : ' . $standardDeviation);
+
+        $recommendedBatchCountLow = $this->getRecommendedBatchCount($sizeAverage, $standardDeviation, self::INCREASED_MARGIN);
```
cammonro: If we know the max possible size in the sample we can recommend an absolute floor that would be below the standard deviation.

cammonro (changes requested):
This is just fantastic work @damcou !! 🙌 

I do see some issues if you wouldn't mind taking a look and had a few suggestions. 

I also think we should add some language that says something along the lines that these numbers are estimates only and that indexing activity should be monitored after making changes to ensure batches are not exceeding the recommended size of 10 MB.

Console/Command/BatchingOptimizeCommand.php
damcou: Oh ok my bad, I missread your earlier comment, I thought there was a compatibility issue with the underscore separators. Re-adding it.

Console/Command/BatchingOptimizeCommand.php
damcou: This was my first idea, but for some reason, I'm not allowed to do it, apparently it's because the store_id parameter is an array ?
image

Console/Command/BatchingOptimizeCommand.php
damcou: Forgot about `pow` , thanks 
and `abs` was useless, indeed !

Console/Command/BatchingOptimizeCommand.php
damcou: Not sure to understand this maths thing, but the results seem accurate 😄

Console/Command/BatchingOptimizeCommand.php
damcou: Good catch !

Console/Command/BatchingOptimizeCommand.php
cammonro: I'm not sure what was causing that but I just added a [POC](https://github.com/algolia/algoliasearch-magento-2/pull/1797/commits/4d06c8c2efd9ee6b093c44f558bc550bcfa7cc49) for a sample size option (did not add for margin). Is this what you were going for or something different?
Sample size

Console/Command/BatchingOptimizeCommand.php
```diff
@@ -0,0 +1,393 @@
...
+        return $stats;
+    }
+
+    /**
+     * @param array $sizes
+     * @return int
+     */
+    protected function getSizeAverage(array $sizes): int
+    {
+        if (count($sizes) <= 1) {
```
cammonro:
```suggestion
        if (empty($sizes)) {
```

That earlier suggestion was intended for the standard deviation `n-1` adjustment which doesn't apply here. Here we can just check for `0`.

cammonro (changes requested):
I added a POC for the issue you mentioned - if you think this can work then we could do something for margin as well.

Also noted one issue on the division by zero check. 

Additional comments in Jira.

Console/Command/BatchingOptimizeCommand.php
damcou: Ok it's working, thanks ! 💯 
Not sure what I did wrong, maybe I was confused between options and arguments 🤔

PR #1798, opened 2025-08-12
MAGE-1383 Address Codacy complaints

cammonro:
**Summary**

Certain rulesets are triggering on Codacy that may be false positives. This PR aims to address them.

Adding unit tests to ensure parity with these changes.

**Result**

image


cammonro:
These static rules seem pretty antiquated and switching to things like imperative for loops would not be the right way to go here IMO. Declarative style is the right approach for PHP 8.2+. Putting this on hold for now.

PR #1799, opened 2025-08-19
MAGE-1396: Fix queue merging

damcou:
The PR contains:
- Fixed the merge mechanism inside the `Job` model, the merge now only depends on the `entitityIds` property, the legacy ones have been removed.
- Fixed the `fetchJobs` method from the `Queue` model. Jobs which are part of a run can now merge up to a total equal to the number set on  "Advanced > Maximum number of records processed per indexing job" .
- Updated Indexing Queue notice in the configuration to specify that users can now expect higher performances depending on those aspects.
- Updated integration tests in `QueueTest.php` class

### DIFF
--- a/Model/Queue.php
+++ b/Model/Queue.php
@@ -493,9 +493,10 @@ class Queue
             foreach ($rawJobs as $job) {
                 $jobSize = (int) $job->getDataSize();
                 $jobSizes[$job->getId()] = $jobSize;
-                $actualBatchSize = array_sum($jobSizes);
                 $jobs[] = $job;
             }
+
+            $actualBatchSize = array_sum($jobSizes);
         }
 
         return $jobs;


Model/Job.php
```diff
@@ -26,24 +32,24 @@ class Job extends \Magento\Framework\Model\AbstractModel implements JobInterface
 {
     protected $_eventPrefix = 'algoliasearch_queue_job';
 
-    /** @var \Magento\Framework\ObjectManagerInterface */
-    protected $objectManager;
+    /** @var ObjectManagerInterface */
```
cammonro, line 35: I like that we're steadily introducing stronger typing. With these changes I think the old PHPDoc becomes redundant and can be removed.

Model/Job.php
```diff
@@ -97,6 +103,10 @@ public function prepare()
             if (isset($decodedData['store_id'])) {
                 $this->setStoreId($decodedData['store_id']);
             }
+
+            if (isset($decodedData['storeId'])) {
```
cammonro, line 107: Looks like some duplication here?

Model/Job.php
```diff
@@ -124,44 +134,17 @@ public function canMerge(Job $job, $maxJobDataSize)
...
-        // @todo Remove on 3.16.0
-        if (isset($decodedData['product_ids'])
-            && count($decodedData['product_ids']) + count($candidateDecodedData['product_ids']) > $maxJobDataSize) {
-            return false;
-        }
-
-        // @todo Remove on 3.16.0
-        if (isset($decodedData['category_ids'])
-            && count($decodedData['category_ids']) + count($candidateDecodedData['category_ids']) > $maxJobDataSize) {
+        if (!isset($candidateDecodedData['entityIds']) || count($candidateDecodedData['entityIds']) <= 0) {
```
cammonro, line 143: This looks like it may have been breaking the merging of newer style job records so it's good we're removing the old stuff. 
Was this also affecting 3.16 or only 3.17?

Model/Queue.php
```diff
@@ -470,28 +473,30 @@ protected function fetchJobs($jobsLimit, $fetchFullReindexJobs = false, $lastJob
...
+            // At this point, if this condition is true, this means that no merge was possible in the last iteration
+            if (count($rawJobs) === $jobsLimit) {
                 $jobs = $rawJobs;
 
                 break;
             }
 
+            // Introduced an array of job sizes to determine the total batch size currently processed (sum of all jobs contained in the run)
+            // This will determine if we can continue to loop over the jobs
+            $jobSizes = [];
```
cammonro, line 491: Not sure why the earlier accumulation was failing but this seems more straightforward to understand and I like that we're getting rid of `break 2` 😅

cammonro (approved):
I just had a couple of questions / observations but no blockers that I can see. Approved on my end. 👍

Model/Job.php
damcou: Good question, I'm pretty sure the property was `entity_ids` and not `entityIds` at some point. but I re-tested on 3.16.0 and it was `entityIds` so it's possible that this was already broken there ... :(

PR #1800, opened 2025-08-20
MAGE-1394 queue clear CLI

cammonro:
**Summary**

To help support advanced scripting for deployments and testing this PR provides an "algolia:queue:clear" command.

- Supports non-interactive mode for scripting
  e.g.
  ```
  bin/magento algolia:queue:clear -n 
  ```

- Supports store specific queue clear operations
- Reorganizes existing CLIs for 3.17
- Provides an alias for `algolia:reindex:process_queue`

**Result**
image

**Todo**

Supply unit tests

### DIFF
--- a/Console/Command/AbstractStoreCommand.php
+++ b/Console/Command/AbstractStoreCommand.php
@@ -127,10 +127,10 @@ abstract class AbstractStoreCommand extends Command
             : "<info>$msg</info>";
     }
 
-    protected function confirmOperation(string $okMessage = '', string $cancelMessage = 'Operation cancelled'): bool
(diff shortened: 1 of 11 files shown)


damcou (approved):
This looks perfect ! Great job ! 🎉 
Tested on a single store and for all stores, it works like a charm and it's super quick!
All unit tests passed !

I'd see an improvement where you could pass an "entity" option which would only delete related jobs (for example for removing all "non products" jobs to speed up the merging and the queue process). But it's not a must have, it's already super cool like this !

I also like the new structure of the commands directory !

PR #1802, opened 2025-08-26
MAGE-1110: Indexing Queue updates

damcou:
This PR contains:
- Some cleaning on the `Queue` and `Job` models
- A new way of calculating maximum batch size in the queue, now we take into account the values defined at store level instead of the default one. This allows to take the values set by the batching optimizer into consideration.
- Added some missing store scopings
- Fixed a merge test

### DIFF
--- a/Model/Job.php
+++ b/Model/Job.php
@@ -33,11 +33,11 @@ class Job extends \Magento\Framework\Model\AbstractModel implements JobInterface
     protected $_eventPrefix = 'algoliasearch_queue_job';
 
     public function __construct(
-        protected Context                $context,
-        protected Registry               $registry,
+        Context                          $context,
+        Registry                         $registry,
         protected ObjectManagerInterface $objectManager,
-        protected ?AbstractResource      $resource = null,
-        protected ?AbstractDb            $resourceCollection = null,
+        ?AbstractResource                $resource = null,
+        ?AbstractDb                      $resourceCollection = null,
         array                            $data = []
     ) {
         parent::__construct($context, $registry, $resource, $resourceCollection, $data);
--- a/Model/Job.php
+++ b/Model/Job.php
@@ -32,28 +32,15 @@ class Job extends \Magento\Framework\Model\AbstractModel implements JobInterface
 {
     protected $_eventPrefix = 'algoliasearch_queue_job';
 
-    /** @var ObjectManagerInterface */
-    protected ObjectManagerInterface $objectManager;
-
-    /**
-     * @param Context                $context
-     * @param Registry               $registry
-     * @param ObjectManagerInterface $objectManager
-     * @param AbstractResource|null  $resource
-     * @param AbstractDb|null        $resourceCollection
-     * @param array                  $data
-     */
     public function __construct(
-        Context                $context,
-        Registry               $registry,
-        ObjectManagerInterface $objectManager,
-        ?AbstractResource      $resource = null,
-        ?AbstractDb            $resourceCollection = null,
-        array                  $data = []
+        protected Context                $context,
+        protected Registry               $registry,
+        protected ObjectManagerInterface $objectManager,
+        protected ?AbstractResource      $resource = null,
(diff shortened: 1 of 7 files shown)


Model/Job.php
```diff
@@ -32,36 +32,23 @@ class Job extends \Magento\Framework\Model\AbstractModel implements JobInterface
...
-        Registry               $registry,
-        ObjectManagerInterface $objectManager,
-        ?AbstractResource      $resource = null,
-        ?AbstractDb            $resourceCollection = null,
-        array                  $data = []
+        protected Context                $context,
+        protected Registry               $registry,
+        protected ObjectManagerInterface $objectManager,
+        protected ?AbstractResource      $resource = null,
+        protected ?AbstractDb            $resourceCollection = null,
```
cammonro: I like that we're steadily cleaning up our constructors but I think we need to take care here with property promotion since the properties being passed to the parent constructor are already declared in the parent. 

We are in effect creating duplicate properties which wastes memory and may create confusion. Instead [we can mix](https://php.watch/versions/8.0/constructor-property-promotion#constructor-property-mixing) the two syntaxes provided we keep optional params last. 

Not sure if this helps but this is how I addressed this issue earlier:
https://github.com/algolia/algoliasearch-magento-2/blob/main/Block/Algolia.php#L65-L66

Model/Queue.php
```diff
@@ -439,15 +401,13 @@ protected function fetchJobs(int $jobsLimit, bool $fetchFullReindexJobs = false,
...
-        $maxBatchSize = $this->configHelper->getNumberOfElementByPage();
-
+        $actualBatchSize = -1;
         $limit = $jobsLimit;
         $offset = 0;
 
         $fetchFullReindexJobs = $fetchFullReindexJobs ? 1 : 0;
 
-        while ($actualBatchSize < $maxBatchSize) {
+        while ($actualBatchSize < $this->maxBatchSize) {
```
cammonro: At first I wondered why we are introducing a stateful variable here as it does not appear to have scope outside of this function. But after further review it looks to me like a mutable high water mark that needs to persist across function calls. Was that the intent? In general I prefer if functions can be as stateless as possible. If we must do this I think we should have a separate function to calculate the average batch.

Model/Queue.php
```diff
@@ -490,18 +450,41 @@ protected function fetchJobs(int $jobsLimit, bool $fetchFullReindexJobs = false,
...
             foreach ($rawJobs as $job) {
                 $jobSize = (int) $job->getDataSize();
                 $jobSizes[$job->getId()] = $jobSize;
                 $jobs[] = $job;
+                $this->maxBatchSize += $this->getStoreMaxBatchSize($job->getStoreId());
             }
 
+            // Final calculation for the loop
             $actualBatchSize = array_sum($jobSizes);
+            $this->maxBatchSize = round($this->maxBatchSize / count($jobSizes));
```
cammonro: I like that we're store scoping the max batch size but I have to admit I found this particular code a little confusing - it seems we're accounting for jobs *across* store indices by determining an average? 

I'm wondering if we should instead *not* be merging jobs across store indices? Perhaps we can revisit `\Algolia\AlgoliaSearch\Model\Job::canMerge` for this?

Model/Queue.php
damcou: yes you're right, that was the intent, I changed the calculation (I put it on a separate function and I removed the stateful variable)

Model/Queue.php
damcou: Gonna answer this one on Jira

Model/Queue.php
cammonro: Thank you! I like the new separate averaging function you’ve created that still keeps the `$maxBatchSize` dynamic for your use case (but not as risky as the stateful approach).

Model/Queue.php
cammonro: I get it now! Thank you for explaining.

PR #1803, opened 2025-08-28
MAGE-1391: Update maximum batch size config

damcou:
This PR contains:
- Moved "Maximum number of records sent per indexing request" config
- Updated `ConfigHelper` class
- Set default value to 1000 
- Added config patch to keep existing config

### DIFF
--- a/Setup/Patch/Schema/ConfigPatch.php
+++ b/Setup/Patch/Schema/ConfigPatch.php
@@ -81,7 +81,7 @@ class ConfigPatch implements SchemaPatchInterface
 
         'algoliasearch_synonyms/synonyms_group/enable_synonyms' => '0',
 
-        'algoliasearch_advanced/queue/number_of_element_by_page' => '300',
+        'algoliasearch_advanced/advanced/number_of_element_by_page' => '1000',
         'algoliasearch_advanced/advanced/remove_words_if_no_result' => 'allOptional',
         'algoliasearch_advanced/advanced/partial_update' => '0',
         'algoliasearch_advanced/advanced/customer_groups_enable' => '0',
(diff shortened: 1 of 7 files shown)


Setup/Patch/Schema/ConfigPatch.php
```diff
@@ -81,7 +81,7 @@ class ConfigPatch implements SchemaPatchInterface
 
         'algoliasearch_synonyms/synonyms_group/enable_synonyms' => '0',
 
-        'algoliasearch_advanced/queue/number_of_element_by_page' => '300',
+        'algoliasearch_advanced/advanced/number_of_element_by_page' => '1000',
```
cammonro: Unless there's a use case I'm missing, I'm not a fan of setting these simple values via this legacy data patch rather than just relying on `config.xml` as there is a risk of SQL integrity constraint violations depending on the install history of the module. We can revisit but I'm wondering for this instance if we should just remove this line since it's redudant with `config.xml` anyway.

damcou: I've added this because this class is still used in our tests. I agree this is not ideal and this would need some KTLO work for sure. Regarding the constraint violations, I think if people upgrade the module more often than once every 5 years, they should be good 😅

cammonro: Indeed! But it can't hurt to just remove the line altogether since `config.xml` will do the job and it's one less potential conflict. I can see possibly using `ConfigPatch` for complex config setup but the simple stuff we should unload.

damcou: Please look at the `TestCase::resetConfig()` method which uses the `ConfigPatch` class. I think that if we remove the line, this will break the tests in the `QueueTest.php` file.

And now that I'm thinking of it, we would probably keep this value to 300 for testing purposes before revamping this class 😅

cammonro: Good callout. Yeah that's some old code. I think that might benefit from a proper test fixture but we can defer this until then. Another approach might be to add a guard clause on the new patch just to be safe. Have we confirmed this passes fresh install? If so I'll stop worrying about it for now... 😆

cammonro (approved):
As discussed in Jira let's move this forward and revisit any related `ConfigPatch` concerns on a separate ticket. Approved!

PR #1804, opened 2025-08-29
MAGE-1407: fixed indexing queue notices display in phtml

damcou:
This PR contains:
- Fixed Indexing Queue notice escaping in phtml

Before:

image

After:

image


cammonro (approved):
Works for me! I just wonder if we should add a comment that these dynamic values are internally generated and not unsanitized user input?

PR #1817, opened 2025-09-18
MAGE-1404: Add checks on configuration migration processed on data patches

damcou:
This PR contains:
- Added a `DataMigrationTrait.php` class which handles the data migration in the `core_config_data` table in the data patches.
- Introduced a check before performing the update statement, which should prevent some "Integrity constraint violation" issues which can occur in some circumstances.


cammonro (approved):
Great solution! 🙌

PR #1818, opened 2025-09-18
MAGE-1404: Refactored config.xml

damcou:
This PR contains:
- Refactored `config.xml` files with values contained in `ConfigPatch.php`
- Introduced `DefaultConfigProvider.php` to be used by integration tests
- Updated `TestCase.php` to use this new provider
- Removed `ConfigPatch.php`

### DIFF
--- a/Setup/Patch/Data/MigrateLegacyConfigPatch.php
+++ b/Setup/Patch/Data/MigrateLegacyConfigPatch.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Setup\Patch\Data;
+
+use Algolia\AlgoliaSearch\Setup\Patch\DataMigrationTrait;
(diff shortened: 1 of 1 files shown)


Setup/Patch/Schema/ConfigPatch.php
```diff
@@ -257,108 +235,34 @@ class ConfigPatch implements SchemaPatchInterface
...
-    {
-        return [];
-    }
-
-    /**
-     * @return ConfigPatch|void
-     */
-    public function apply()
-    {
-        $movedConfigDirectives = [
```
cammonro, line 299: I'm not sure during which version of the extension these moved settings were introduced - but I'm wondering if we should add a migrate patch for these as well?

damcou: Looks like very old config from 2022, but I'll create a data patch for it anyway 👍

PR #1821, opened 2025-09-19
MAGE-1426 dynamic faceting fixes (also MAGE-1427)

cammonro:
**Summary**

This PR aims to correct two issues:

1. Replica forwarding necessary to support alternate sorts with dynamic faceting can abort if async operation fails to complete in time
2. Local InstantSearch widgets can override facet sorting behavior in merchandising rules

**Result**

Unit tests and integration tests have been added to catch the replica forwarded settings issue in the future.

Unit tests with missing wait op (fails as expected):
image
Unit tests with wait op:
image

Integration tests on PHPUnit 10 (For Magento 2.4.8):
image

For sorting via dynamic facet merch rules:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ vendor/
 .php_cs.cache
 .vscode
 .idea
+node_modules
--- a/Service/IndexSettingsHandler.php
+++ b/Service/IndexSettingsHandler.php
@@ -55,6 +55,7 @@ class IndexSettingsHandler
                 true,
                 false
             );
+            $this->connector->waitLastTask($indexOptions->getStoreId());
         }
         if ($noForward) {
             $this->connector->setSettings(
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ vendor/
 .php_cs.cache
 .vscode
 .idea
+node_modules
```
damcou, line 8: Is there a reason why it appears now ?

damcou (approved):
LGTM 🚀 
Run the integration and unit tests, everything is green ✅ 
Good catch regarding the missing waitLastTask call !

.gitignore
cammonro: We've never ignored `node_modules` it seems. I needed to use eslint to verify Codacy false positives and doing so introduced a ton of JS modules. We want to be sure to never include any JS packages like that by accident.

PR #1828, opened 2025-09-29
MAGE-1419 Fix floating point precision bug in EventProcessor for v3.15 (also MAGE-1434)

cammonro:
**Summary**

This PR follows up on the issue raised on https://github.com/algolia/algoliasearch-magento-2/pull/1780 by applying rounding to all potentially problematic floating point values. 

Unit tests have been added for scenarios vulnerable to floating point decimal inaccuracies resulting from binary math. 

A 3 decimal scale is used for precision to accommodate currencies such as KWD (Kuwaiti Dinar), BHD (Bahraini Dinar), and JOD (Jordanian Dinar) all of which have 1,000 fils per unit and not cents such as used in currencies like dollars or euros. 

**Result**

Before:
image
image

After:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug fixes
 - Ensure that only non-redirect URL rewrites are considered when generating product URLs - thank you @fasimana
 - Apply rounding to insight events revenue values to avoid floating point precision errors - thank you @PromInc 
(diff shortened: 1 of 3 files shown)


damcou (approved):
🚀

damcou (approved):
🚀

PR #1829, opened 2025-09-29
MAGE-1434 Fix floating point precision bug in EventProcessor for v3.16

cammonro:
**Summary**

This PR:
- Ports https://github.com/algolia/algoliasearch-magento-2/pull/1780 to version 3.16.
- Merges 3.15 unit tests for EventProcessor into 3.16 and relocates to match namespace pattern.

**Result**

image


damcou (approved):
🚀

PR #1830, opened 2025-09-29
MAGE-1434 Fix floating point precision bug in EventProcessor for v3.17

cammonro:
**Summary**

This PR:
- Ports https://github.com/algolia/algoliasearch-magento-2/pull/1829 to version 3.17.

**Result**

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,12 +28,12 @@
 ## 3.16.1
 
 ### Updates
-
 - `EventProcessor` now calculates decimal precision on currency based on `Magento\Framework\Locale\FormatInterface`
 
 ### Bug fixes 
 - Ensure that only non-redirect URL rewrites are considered when generating product URLs - thank you @fasimana
 - Apply rounding to insight events revenue values to avoid floating point precision errors - thank you @PromInc
+- Fix store id for queue jobs sorting - thank you @pikulsky
 
 ## 3.16.0
 
--- a/Model/Queue.php
+++ b/Model/Queue.php
@@ -578,7 +578,7 @@ class Queue
                 SORT_ASC,
                 'method',
                 SORT_ASC,
-                'store_id',
+                'storeId',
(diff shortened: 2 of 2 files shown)


Service/Insights/EventProcessor.php
```diff
@@ -18,13 +20,24 @@ class EventProcessor implements EventProcessorInterface
...
+        protected ?InsightsClient       $client = null,
+        protected ?string               $userToken = null,
+        protected ?string               $authenticatedUserToken = null,
+    ) {
+        $this->initDecimalPrecision();
+    }
+
+    protected function initDecimalPrecision(): void
+    {
+        $this->decimalPrecision = $this->localeFormat->getPriceFormat()['requiredPrecision']
```
damcou, line 38: I can see that `getPriceFormat` can take `$localeCode` and `$currencyCode` as parameters (that we don't pass here), and I'm wondering how it would behave in case of a multicurrency store with different currencies and precisions (USD and KWD for example) ... But I guess this is an extremely unlikely scenario which wouldn't hurt that much at Insights level in any case.

cammonro: When I looked at the implementation of `\Magento\Framework\Locale\Format::getPriceFormat` I noted that it fell back to resolvers on both values. I decided that rather try to get too fancy I would let Magento decide the locale and currency based on its own internals. 

If that doesn't work for someone I did separate the initialization of the precision via protected method `\Algolia\AlgoliaSearch\Service\Insights\EventProcessor::initDecimalPrecision`. Worst case you could write a preference on that method and handle that method however you wish. 

Also with the plugin available for `\Algolia\AlgoliaSearch\Service\Insights\EventProcessor::applyPrecision` I figured the implementation should be flexible enough for others. We shall see! 😅

damcou: Yes ! No problem for me :)

PR #1836, opened 2025-09-30
MAGE-1423: update version files and changelog (3.15.2)

damcou:
Prepping for 3.15.2 release

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Algolia Search & Discovery extension for Magento 2
 ==================================================
 
-![Latest version](https://img.shields.io/badge/latest-3.15.2-green)
+![Latest version](https://img.shields.io/badge/latest-3.16.0-green)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 Algolia Search & Discovery extension for Magento 2
 ==================================================
 
-![Latest version](https://img.shields.io/badge/latest-3.15.1-green)
+![Latest version](https://img.shields.io/badge/latest-3.15.2-green)
```
cammonro: I'm wondering should we indicate the latest is 3.16.0 here (at time of release)?

cammonro (commented):
I'm feeling like for patch releases we should start making it clear that there's a newer version in the README. WDYT?

README.md
damcou: Makes sense to me !

Updated the file :)

PR #1845, opened 2025-10-08
3.17.0-beta.2 release

damcou:
### Updates
- Removed all `is_null` occurrences

### Bug fixes
- Fixed 3.17 setup:upgrade on PHP 8.4
- Fixed many Codacy issues

⚠️ **This also contains changes from 3.16.1.**


CHANGELOG.md
```diff
@@ -19,6 +28,23 @@
 ### Bug fixes
 - Fixed indexing queue templates escaping.
 
+## 3.16.1
+
+### Updates
```
cammonro, line 33: @damcou I realized we missed notating some updates on the original change log - which I've added here in Github:  https://github.com/algolia/algoliasearch-magento-2/releases/tag/3.16.1 

I'm not sure if we'll have a 3.16.2 but we should be sure to update the 3.17 release I would think?

cammonro (approved):
Please see my query regarding the change log for 3.16.1 but otherwise approved on my end!

CHANGELOG.md
damcou: I updated the changelog so it reflects the changelog from the 3.16.1 release :)

PR #1921, opened 2026-03-21
MAGE-1550 agent context and linting

cammonro:
**Summary**

_Agent & Human Context Architecture_
- Adopted `AGENTS.md` as the open cross-tool standard for agent and contributor context, replacing vendor-specific CLAUDE.md inline context
- Created `doc/ARCHITECTURE.md` covering bird's eye view, codemap, indexing pipeline (delta vs full), queue system, multi-store architecture, database tables, key decisions, architectural invariants, and contributor patterns
- Reduced `CLAUDE.md` to a single-line `@AGENTS.md` shim - all substantive context now lives in tool-agnostic files

_PHP-CS-Fixer Modernization_

- Replaced brittle `$_SERVER['argv'][3]` config loading (dating to 2018) with `getcwd()` + `file_exists()` graceful fallback -- fixer now works standalone without Magento root
- Added `@PHP82Migration` ruleset to anchor minimum PHP compatibility
- Added parallel execution support 
- Various rule updates

_PHPStan Configuration_

- Simplified `phpstan.neon` from 14 hardcoded vendor paths to `%currentWorkingDirectory%` with `dev` exclusion
- Designed to work with the new `magento2-analyse` utility in `magento2-tools` which handles Magento root detection and autoloader bootstrapping

_Contributor Guidelines (`CONTRIBUTING.md`)_

- Replaced abandoned `magento/marketplace-eqp` with `magento/magento-coding-standard` (already in Magento 2.4.x dev dependencies)
- Updated all PHP-CS-Fixer references from `.php_cs` to `.php-cs-fixer.php` 
- Updated Magento dist config reference from `.php_cs.dist` to `.php-cs-fixer.dist.php`
- Documented actual PHPCS commands with `Magento2` standard and proper flags
- Updated `magento2-tools` command reference: `magento2-types` replaced by `magento2-analyse`, added vendor bin auto-detection docs

> NOTE: Requires updates to tooling at: https://github.com/algolia/magento2-tools/pull/10


doc/ARCHITECTURE.md
```diff
@@ -0,0 +1,327 @@
...
+    -> buildIndexList queue jobs
+    -> IndexBuilder -> RecordBuilder -> AlgoliaConnector
+    -> records updated in-place on the live index
+
+Full (complete rebuild):
+  CLI command (algolia:reindex:*) or executeFull() if enabled
+    -> BatchQueueProcessor.handleFullIndex()
+    -> buildIndexFull queue jobs
+    -> IndexBuilder -> RecordBuilder -> AlgoliaConnector
+    -> records written to a temporary index
```
damcou: Shouldn't we mention that this is now optional ?

doc/ARCHITECTURE.md
```diff
@@ -0,0 +1,327 @@
...
+the correct store context during indexing.
+
+### `Model/`
+
+Magento models, the indexer layer, and the queue system.
+
+`Model/Indexer/` contains 7 Magento indexers (registered in `etc/indexer.xml`): Products,
+Categories, Pages, Suggestions, AdditionalSections, QueueRunner, and DeleteProduct. It
+also contains three observer-style classes (ProductObserver, CategoryObserver,
+PageObserver) that intercept entity save/delete via Magento's plugin system.
```
damcou: Maybe it would be useful to say that those observers are miss-named from the beginning of the extension and probably will be renamed in the future ?

doc/ARCHITECTURE.md
```diff
@@ -0,0 +1,327 @@
...
+Each job references a class and method that must appear in `Job::ALLOWED_HANDLERS` - a
+hardcoded whitelist that prevents arbitrary code execution if the queue table is
+compromised. Execution logs go to `algoliasearch_queue_log`; completed or failed jobs
+are archived to `algoliasearch_queue_archive`.
+
+The queue processes full reindex and delta jobs in a mixed ratio
+(`FULL_REINDEX_TO_REALTIME_JOBS_RATIO = 0.33`), ensuring delta updates get at least 67%
+of each processing cycle.
+
+## Multi-Store Architecture
```
damcou: Why not rename this paragraph into something like "Index name conventions" and also add a small sentence about customer group indices as well ? (I don't think they are mentioned in the doc and they are worth being mentioned IMO)

AGENTS.md
```diff
@@ -0,0 +1,68 @@
+# AGENTS.md
+
+## Project Overview
+
+Algolia Search & Discovery extension for Magento 2 (`Algolia_AlgoliaSearch`). Indexes Magento entities (products, categories, pages, suggestions, additional sections) into Algolia, manages sorting replicas, captures analytics events (clicks, conversions) and personalization signals, and provides frontend search via autocomplete, InstantSearch, faceted navigation, and Recommend widgets. Includes built-in merchandising (landing pages) — likely to be deprecated or reworked. Uses the Algolia PHP API Client v4.18.3.
+
+**Requirements:** PHP 8.2-8.4, Magento 2.4.6+, `algolia/algoliasearch-client-php` 4.18.3
```
damcou: Additional piece of documentation that needs to be checked on version release from now :)

damcou (approved):
Nice cleanup on the documentation and on (very) old files 🎉 
Had some small non-blocking suggestions, feel free to disregard if you want ✅

doc/ARCHITECTURE.md
cammonro: Good idea! Will tweak this.

doc/ARCHITECTURE.md
cammonro: I carefully chose to call them "observer-style classes" and not observers for this very reason! But a note here wouldn't hurt. Let's just follow through... 😛

doc/ARCHITECTURE.md
cammonro: Multi store orientation is intrinsic to the entire integration which is why I didn't want to limit in scope to index name convention only. 

However, you do make a good point about customer group pricing. These are not true indices but replicas with design intention that should be called out. Right now replicas are mentioned for sorting - but not in great detail. 

Going to rework this. It's important context for agents and humans alike!

AGENTS.md
cammonro: Agreed! Definitely want to revisit this with every minor release - but I think I'll relax the PHP version to reduce the maintenance surface.

cammonro:
These were really good suggestions @damcou . Updated!

damcou (approved):
Look perfect ✅ 

Great job !

PR #1923, opened 2026-03-25
MAGE-1531 Apply PHPCSF linting rules implemented in updated magento2-…

cammonro:
**Summary**
The updated magento2-tools package ships an opinionated fixer config aligned with PSR-12 and modern PHP (8.2+). Running it once against the full codebase means all future diffs will be style-clean, making PRs easier to review and CI linting pass on first push.

- Applies the PHP-CS-Fixer ruleset from `algolia/magento2-tools` v0.4.0 across all 294 files to establish a consistent code style baseline for contributions
- Key fixes include: explicit `public const` visibility, trailing commas in multi-line arrays, blank line before return statements, `mb_rtrim`/`mb_strlen` over legacy string functions, removal of redundant PHPDoc blocks (where type info is already in the signature), and consistent docblock spacing
- Net reduction of ~1,700 lines, almost entirely removed boilerplate comments and redundant annotations

**Result**
Unit tests
image
Integration tests
image

PR #1924, opened 2026-03-26
MAGE-1531 Introduce SendStrategy pattern to AlgoliaConnector 

cammonro:
**Summary**

This PR introduces a "strategy pattern" for record delivery in `AlgoliaConnector`, enabling satellite modules (e.g., upcoming `Algolia_Ingestion` module) to replace the default `SearchClient::batch()` call without modifying core extension code.

(Targeting the linting PR https://github.com/algolia/algoliasearch-magento-2/pull/1923 to cleanly separate these change sets)

Note that there is ***no behavioral change***. `DirectSendStrategy` produces the same `batch()` call with the same payload shape as before.

- **`Api/SendStrategyInterface`** (public API) - defines interface for external implementations
- **`Service/DirectSendStrategy`** - default implementation; wraps the existing `SearchClient::batch()` call extracted from `performBatchOperation()`. Behavior should be identical to before.
- **`Service/SendStrategyResolver`** - iterates optional strategies (injected by satellite modules via `di.xml`); falls back to `DirectSendStrategy`
- **`AlgoliaConnector::performBatchOperation()`** - delegates to the resolved strategy instead of calling the client directly.
- **`di.xml`** - and uses `AlgoliaConnector\Proxy` on `DirectSendStrategy` to break the resulting circular dependency. 

> [!NOTE]
The circular dependency is not ideal even with the proxy. I have scoped a new task to break down the composition of `AlgoliaConnector` to address this issue. 

**Result**

`AlgoliaConnectorTest` was introduced for baseline testing and then via a red-green-refactor, the mock injection point was moved from `SearchClient` (in `clients` array) to `SendStrategyInterface` (via mock resolver). All existing payload assertions are preserved.

Added a new `DirectSendStrategyTest` to cover the extracted `batch()` delegation directly.

image

### DIFF
--- a/Api/SendStrategyInterface.php
+++ b/Api/SendStrategyInterface.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Api;
+
+use Algolia\AlgoliaSearch\Api\Data\IndexOptionsInterface;
+
+interface SendStrategyInterface
+{
+    /**
+     * Determine if this strategy should handle send operations for the given store.
+     * Only called on optional strategies - not the default fallback.
+     *
+     * @param int $storeId
+     * @return bool
+     */
+    public function isApplicable(int $storeId): bool;
+
+    /**
+     * @param IndexOptionsInterface $indexOptions
+     * @param array $requests Batch requests [{action, body}, ...]
+     * @return array Response from the send operation
+     */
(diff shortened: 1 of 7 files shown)


damcou (approved):
Looks good ! 🙌 ✅ 

More feedback on the ticket.

PR #1926, opened 2026-03-26
MAGE-1556 Extract SearchClientProvider

cammonro:
**Summary**

This PR decomposes `AlgoliaConnector` so that search client provisioning can be invoked independently and relieves the original connector of this responsibility. This is especially useful for supporting the strategy pattern introduced on https://github.com/algolia/algoliasearch-magento-2/pull/1924. 

- Lifecycle management (creation, caching, credentials, user agent) has been moved from `AlgoliaConnector` into a new `SearchClientProvider` class behind a `SearchClientProviderInterface`)
- Eliminates the circular dependency discovered during introduction of the send strategy (`AlgoliaConnector -> SendStrategyResolver -> DirectSendStrategy -> AlgoliaConnector`) that required a Magento proxy as a temporary workaround in `di.xml`
- `DirectSendStrategy` now depends on the narrow `SearchClientProviderInterface` instead of the full `AlgoliaConnector`
- Updates `ARCHITECTURE.md` with new design pattern
- Unit tests updated / added 

The change is intended to be purely architectural and should represent no behavioral change.

**Result**
Unit tests:
image
Integration tests:
image

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,5 +64,6 @@ CircleCI runs on branches matching `^(feat|fix|chore)/MAGE.*`. Tests against PHP
 
 - PSR-2 base with additional rules in `.php-cs-fixer.php`
 - Comments should be rare and only when logic isn't self-descriptive — prefer renaming classes/methods over adding comments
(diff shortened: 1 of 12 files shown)


damcou (approved):
LGTM ! Good job ! 🙌 🚀

PR #1928, opened 2026-03-31
MAGE-1534: introduce AbstractClientProvider class

damcou:
This PR contains:
- Introduction of the `AbstractClientProvider`
- Refactoring of the  `SearchClientProvider`

### DIFF
--- a/Service/AbstractClientProvider.php
+++ b/Service/AbstractClientProvider.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Service;
+
+use Algolia\AlgoliaSearch\Api\ClientProviderInterface;
+use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
+use Algolia\AlgoliaSearch\Support\AlgoliaAgent;
+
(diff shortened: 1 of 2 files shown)


Service/AbstractClientProvider.php
```diff
@@ -0,0 +1,56 @@
...
+     * @throws AlgoliaException
+     */
+    protected function createClientByStore(?int $storeId): void
+    {
+        if ($storeId === null) {
+            $storeId = ClientProviderInterface::ALGOLIA_DEFAULT_SCOPE;
+        }
+
+        if (!isset($this->clients[$storeId])) {
+            $this->createClient($storeId);
```
cammonro, line 42: We should also define `createClient` as an abstract method in this class.

cammonro (changes requested):
Looks good! Just one change needed.

Service/AbstractClientProvider.php
damcou: Done ✅

cammonro (approved):
Parfait! Allons-y! 🚀

PR #1929, opened 2026-04-01
MAGE-1557: Introduce Claude testing guides 

damcou:
This PR contains:
- Introduced Claude testing guides:
  - `testing-core.md `which contains universal rules for all tests
  - `types/testing-*.md `for all type related rules
  - `specifics/testing-<ClassName>.md` for class specific rules
- Added /`generate-unit-tests <ClassName> `skill to automatically generate tests for a given class.
- Generated `AlgoliaConnectorTest.php` class as an example.

### DIFF
--- a/.claude/skills/generate-unit-tests/SKILL.md
+++ b/.claude/skills/generate-unit-tests/SKILL.md
@@ -1,0 +1,65 @@
+---
+name: generate-unit-tests
+description: Generates focused PHPUnit unit tests for a given PHP class in the AlgoliaSearch module, following the project's 3-layer testing guidelines (core rules → type guide → class-specific rules).
+---
+
+Generate focused PHPUnit unit tests for the file: $ARGUMENTS
+
+Follow these steps exactly:
+
+## Step 1 — Load foundation rules
+Read `.claude/testing-core.md`. These rules are non-negotiable and apply to every test you generate.
+
+## Step 2 — Load the type-specific guide
+Determine which category the target file belongs to based on its path and class name:
+
+| If the path contains…          | Read this guide                              |
+|---------------------------------|----------------------------------------------|
+| `Observer/`                     | `.claude/types/testing-observers.md`         |
+| `Service/`                      | `.claude/types/testing-services.md`          |
+| `Helper/`                       | `.claude/types/testing-helpers.md`           |
+| `Model/Indexer/` or `Indexer/`  | `.claude/types/testing-indexers.md`          |
+| anything else                   | skip — use core rules only                   |
+
+Read the matching guide before generating any test code.
+
+## Step 3 — Check for class-specific rules
+Extract the class name from the file path (e.g. `AlgoliaConnector` from `Service/AlgoliaConnector.php`) and check whether `.claude/specifics/testing-<ClassName>.md` exists. If it does, read it — its rules take precedence over the type guide for this class.
+
+## Step 4 — Read the target file
+Read the full source file at the path provided in $ARGUMENTS. Understand:
+- Constructor dependencies (these will all be mocked)
+- Public methods and their signatures
+- Guard clauses and early exits
+- Any use of `addCommitCallback`, event dispatch, or external service calls
+
+## Step 5 — Determine output path
+Mirror the module structure under `Test/Unit/`:
+- `Model/Indexer/CategoryObserver.php` → `Test/Unit/Model/Indexer/CategoryObserverTest.php`
+- `Service/Product/RecordBuilder.php` → `Test/Unit/Service/Product/RecordBuilderTest.php`
+- `Helper/ConfigHelper.php` → `Test/Unit/Helper/ConfigHelperTest.php`
+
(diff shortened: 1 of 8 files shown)


.claude/skills/generate-unit-tests/SKILL.md
```diff
@@ -0,0 +1,65 @@
...
+- `Model/Indexer/CategoryObserver.php` → `Test/Unit/Model/Indexer/CategoryObserverTest.php`
+- `Service/Product/RecordBuilder.php` → `Test/Unit/Service/Product/RecordBuilderTest.php`
+- `Helper/ConfigHelper.php` → `Test/Unit/Helper/ConfigHelperTest.php`
+
+Check if a test file already exists at that path. If it does, read it first to avoid duplicating existing tests.
+
+## Step 6 — Generate tests
+
+Rules:
+- **Maximum 10 tests** per file
```
cammonro: I can see how this limit can help prevent the model from generating dozens of low-value tests but at the same time I'm concerned about this seemingly arbitrary value. We lift it necessarily for `AlgoliaConnector` but in general we should be targeting code coverage quality vs quantity and I don't think there's a magic number for that. `ConfigHelper` for instance is massive and will easily need more than 10 tests. 

Perhaps we can add a section that instructs the model on how to budget test creation. e.g. 1 (or maybe 2?) tests per code path through a method (i.e. driven by cyclomatic complexity)

Ultimately every public method should probably be tested at least once (with certain exceptions like delegating methods, etc.)

.claude/skills/generate-unit-tests/SKILL.md
```diff
@@ -0,0 +1,65 @@
...
+- **Focus on user-observable behavior**: what the method returns, what exception it throws, which collaborator it calls with which arguments
+- **Do not test implementation details**: internal variable names, call counts on low-level helpers, private method logic
+- **Use `@dataProvider`** (or `#[DataProvider]` attribute) when testing the same behavior across multiple input variants
```
cammonro, line 51: This is great! I'm a fan of using data providers - it will also help collapse the number of tests generated.

.claude/skills/generate-unit-tests/SKILL.md
```diff
@@ -0,0 +1,65 @@
...
+- `Helper/ConfigHelper.php` → `Test/Unit/Helper/ConfigHelperTest.php`
+
+Check if a test file already exists at that path. If it does, read it first to avoid duplicating existing tests.
+
+## Step 6 — Generate tests
+
+Rules:
+- **Maximum 10 tests** per file
+- **Focus on user-observable behavior**: what the method returns, what exception it throws, which collaborator it calls with which arguments
+- **Do not test implementation details**: internal variable names, call counts on low-level helpers, private method logic
```
cammonro, line 50: Maybe we can have sections on what to test vs what not to test?

e.g. we shouldn't test...
- delegation methods (this wrappers that just call something else) 
- `@deprecated` methods
- Don't write a test that just return what we mock (I've seen AI do this and it's totally useless)

Test/Unit/Service/AlgoliaConnectorTest.php
cammonro, line 1: If this is just a POC ignore my feedback - but if it's the output of the skill I think we should reevaluate and iterate a little more. I say this mainly because there are still key tests we're missing like `waitLastTask()` which is a really important method with some different paths - maybe a data provider could be used? Also `castProductObject`, and the $mergeSettings flag on setSettings as there's a bit logic there on that path as well.

cammonro (commented):
Nice work here. This is a solid approach! I like how it's spec-driven and helps prevent generating useless tests. I think this is the way to go. 

I do think we can make it work even smarter with some tighter prompting and guardrails and added some comments to this effect. We can also add additional specs for the other namespaces like Controller etc. but this is a great starting point.

.claude/skills/generate-unit-tests/SKILL.md
damcou: You're right, for `ConfigHelper` for example, this is not really meaningful to keep this kind of limit, and this would be a perfect candidate to have its own md file in the `specifics` directory.

I'll remove this explicit limit and try what you're suggesting

cammonro (approved):
This is a good foundation and now we have coverage of all primary public methods in `AlgoliaConnector` (at least one test per the updated context). 

Approving as a net improvement over the current state (no tests).

A few observations for follow-up:

- `waitLastTask` has probably 4 or more distinct paths (store-specific, global fallback, explicit args, null storeId default) but only 2 are covered. I think the "cyclomatic complexity" guidance in the skill context could be more actionable here. We probably need to be more explicit for the agent. 
- `handleTooBigRecord` has three outcomes (pass-through, truncate, skip) - only skip is exercised through the `saveObjects` tests. The truncation path is where silent data loss can occur so this is worth covering.
- The class-specific guide's "focus on client-calling methods" may have inadvertently deprioritized path depth within those methods. 

These issues can be revisited in a future tooling update.

PR #1930, opened 2026-04-02
Fix suggestions reindex always results in empty index

its-leoeff511:
## Problem

Running `bin/magento algolia:reindex:suggestions` completes without error but the `_suggestions` index in Algolia always ends up empty.

## Steps to Reproduce

1. Ensure `algoliasearch_indexing_manager/algolia_indexing/enable_query_suggestions_index = 1`
2. Ensure `search_query` table has records meeting the minimum popularity/results thresholds
3. Run `bin/magento algolia:reindex:suggestions` - pass in store_id if preferred
4. Check the `_suggestions` index in Algolia — 0 records

## Root Cause

`rebuildStoreSuggestionIndexPage` saves records directly to the final index (e.g. `magento2_default_suggestions`) instead of the temporary index. 

The subsequent `moveStoreSuggestionIndex` call then:

1. Copies query rules from the final index -> `_tmp` (creating an empty `_tmp`)
2. Moves the empty `_tmp` over the final index essentially **wiping all records just saved**

This breaks the intended pattern used by the product `IndexBuilder`, where records are built in `_tmp` and moved to final in one operation.

## Fix

Pass `$useTmpIndex` to `buildEntityIndexOptions` in `rebuildStoreSuggestionIndexPage`.

To align with the product builder's pattern, `$useTmpIndex` is passed as a parameter from `buildIndex` ->  `rebuildStoreSuggestionIndexPage` rather than hardcoding `true`.

## Affected Versions

Confirmed in v3.16.1 and I believe through v3.17.3.

### DIFF
--- a/Service/Suggestion/IndexBuilder.php
+++ b/Service/Suggestion/IndexBuilder.php
@@ -84,7 +84,8 @@ class IndexBuilder extends AbstractIndexBuilder implements IndexBuilderInterface
                     $storeId,
                     $collection,
                     $page,
-                    $this->configHelper->getNumberOfElementByPage($storeId)
+                    $this->configHelper->getNumberOfElementByPage($storeId),
+                    true
                 );
                 $page++;
             }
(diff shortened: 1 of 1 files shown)


damcou:
Hi @its-leoeff511  , thanks a lot for having raised this issue, this will be part of a larger initiative to better manage temporary index for all entities including pages, suggestions and additional sections in this PR: https://github.com/algolia/algoliasearch-magento-2/pull/1960 . We will make sure to give you credit for it in our changelog.

Closing this PR, thanks again !

PR #1931, opened 2026-04-15
Prevent blocking other cron job execution on the mainline process

thomas-kl1:
**Summary**

**Result**

### DIFF
--- a/etc/cron_groups.xml
+++ b/etc/cron_groups.xml
@@ -7,6 +7,6 @@
         <history_cleanup_every>10</history_cleanup_every>
         <history_success_lifetime>60</history_success_lifetime>
         <history_failure_lifetime>4320</history_failure_lifetime>
-        <use_separate_process>0</use_separate_process>
+        <use_separate_process>1</use_separate_process>
     </group>
 </config>


cammonro:
@thomas-kl1 

Thanks for the PR, and for raising this. Heavy queue runs crowding out other cron groups can be a real problem, but after evaluating we’ve decided not to change the shipped default. 

There are three reasons for this:

~~**This setting does not actually unblock the mainline process.** `Magento\Framework\Shell::execute()` is a blocking `exec()`, and the observer awaits it before moving to the next group. You do get process isolation (a fatal or OOM error kills only the child) with the proposed setting but not concurrency. So a four minute queue run will still delay every other group by four minutes, and the `uksort()` in `ProcessCronQueueObserver::execute()` will actually run it *before* the other jobs - rather than after - so that seems like it could be worse to me? Is this different than what you are seeing?~~

> REDACTED (per thread): DI changes this assumed behavior. `ProcessCronQueueObserver` gets the `shellBackgroundCron` virtualType, not the default Shell, `CommandRendererBackground::render()` appends a trailing `&`. The rendered command looks something like:
`'/usr/bin/php' '/app/bin/magento' cron:run --group='algolia' --bootstrap=standaloneProcessStarted=1 >> '.../var/log/magento.cron.algolia.log' 2>&1 &`.
That `&` matters here. The wrapping shell exits immediately, so `exec()` returns in milliseconds instead of waiting on the child. So groups do run in parallel. 👍 

**This is already a merchant-owned configuration.** Magento exposes this per group at **Stores > Configuration > Advanced > System > Cron** ("Cron configuration options for group: algolia" - `core_config_data` path:`system/cron/algolia/use_separate_process`). 

So anyone who wants this can set it today with no release needed. `cron_groups.xml` only supplies the default for merchants who never touch it.

**Changing that default risks making things worse for other merchants.** ~~`Shell::execute()` throws on any non-zero child exit, and nothing up the stack catches it. Since any separate-process groups are sorted first, a failing Algolia child process could abort the whole cron run and skip every group after it. On top of that,~~  the fork happens before any check that a job is due, so every cron tick would pay a full Magento bootstrap even though we ship this feature disabled by default.

**If it’s parallelism you’re after** … how about a dedicated crontab line, so the scheduler starts the processes independently? Maybe something like:

```
*/5 * * * * php bin/magento cron:run --group=algolia
*/5 * * * * php bin/magento cron:run --exclude-group=algolia
```

I would leave `use_separate_process` at 0 with this setup or you could end up forking a redundant process. 

If the run itself is too long or too memory hungry, I would take a look at  `algoliasearch_queue/queue/number_of_job_to_run` and `number_of_element_by_page` and try tweaking those knobs. We shipped a batching optimizer in 3.17 to help with this.

There wasn’t much detail provided on this, so if you are seeing something we’re not addressing please let us know and we’ll be happy to look into this deeper. As always we really appreciate contributions from the community! Thanks again.

cc @damcou

thomas-kl1:
While I agree with the configuration part, I disagree with your understanding of the background task.
‘use_separate_process‘ does run the task in background and by doing so the group jobs are running in //

Check the shell instance injected in the ProcessCronQueueObserver:

```xml
<virtualType name="shellBackgroundCron" type="Magento\Framework\Shell">
        <arguments>
            <argument name="commandRenderer" xsi:type="object">Magento\Cron\Shell\CommandRendererBackground</argument>
        </arguments>
    </virtualType>
```

`\Magento\Cron\Shell\CommandRendererBackground::render`

You can check and play with xdebug to see this.

cammonro:
@thomas-kl1 You're right to push back on that parallelism claim of mine and I appreciate you challenging my assumptions on how I read that blocking exec. The core mechanic you cite does indeed seem to counter that argument:

I traced it properly this time. `ProcessCronQueueObserver` gets the `shellBackgroundCron` virtualType, **not the default Shell**. I was evaluating `Shell::execute()` in isolation and missed the DI override entirely. 🤦 Redacting that from my original argument.

I'm still reluctant to ship as a default because cron is disabled by default and the extra bootstrap still happens regardless. (I don't see a way to change that behavior on the fly based on a config.) Maybe that's a marginal cost - I'm not really sure and would need to explore this further. However, as mentioned anyone who wants this can enable it today without waiting on a release from us.

What would change my mind here are some real numbers. Could you share which cron group was being starved, roughly how long the Algolia group was running, and any relevant details on the environment or cron config? Before / after metrics would be good as well.

thomas-kl1:
Actually in some scenarios, you can have the queue being filled with small jobs (like 1 to 5 items to reindex max) to heavy jobs (jobs for full reindex, so it's divided by batch of hundreds of items).

But because the queue is configured to execute a maximum number of jobs, regardless the actual number of items to reindex it can fill up quickly and depending of the queue configuration it can stuck the other cron job to be executed at time and some can be missed.

I agree that it can be configured per installation so this is totally fine to me.
Also since I've opened this PR I've also faced the side of the coin, when having too much jobs with the `use_separate_process` setting on, can overload the CPU of your instance (ex PSH where the cron are on the same node as the backend server).

**TLDR**; Keep this PR closed 😄 As you said, change the setting directly in config regarding the needs of the project!

PR #1932, opened 2026-04-16
MAGE-1536 unit tests

cammonro:
**Summary**

- Add `SendStrategyResolver` unit tests covering resolution order, default fallback, store ID forwarding, and edge cases 
- Add `IndexNameFetcher::getOriginalIndexName()` helper to strip the `_tmp` suffix from temporary index names, needed by send strategy logic to identify the target index 
- Configure `AlgoliaLogger` with a `PsrLogMessageProcessor` virtual type that removes used context fields, enabling clean PSR-3 placeholder substitution in log messages 

**Result**
Unit tests:
image

### DIFF
--- a/Service/IndexNameFetcher.php
+++ b/Service/IndexNameFetcher.php
@@ -54,6 +54,18 @@ class IndexNameFetcher
     }
 
     /**
+     * If the index is a temporary index, return the original index name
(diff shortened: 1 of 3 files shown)


damcou (approved):
Good job ! Reviewed and tests are passing ✅

PR #1933, opened 2026-04-24
MAGE-1568: Introduced IndexSettingsComparator service

damcou:
This PR contains:
- Introduced `IndexSettingsComparator` service which check if the sent settings are the same as in the Algolia Dashboard to prevent unnecessary `setSettings` operations. 
- Changed `typoTolerance` typing for the suggestions settings (`typoTolerance` is now an enum: https://github.com/algolia/api-clients-automation/blob/main/specs/common/schemas/SearchEnums.yml#L22 (previous boolean value broke the comparison between settings coming from the extension and the dashboard)
- Adding logging when `setSettings` occurrences are skipped when no diff is found (MAGE-1572)
- `IndexSettingsHandler::setSettings()` now return a boolean (`false` when `setSettings` is skipped, `true` otherwise)
- Updated `IndexSettingsHandlerTest` as a result


Helper/Entity/SuggestionHelper.php
```diff
@@ -41,7 +41,7 @@ public function getIndexSettings(?int $storeId = null): array
         $indexSettings = [
             'searchableAttributes' => ['query'],
             'customRanking'        => ['desc(popularity)', 'desc(number_of_results)', 'asc(date)'],
-            'typoTolerance'        => false,
+            'typoTolerance'        => 'false',
```
cammonro, line 44: This is a great catch. Have we done a sweep of our defaults to ensure no other schema mismatches exist? If so we'll get false positives on our comparisons.

Service/IndexSettingsDiffChecker.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Service;
+
+use Algolia\AlgoliaSearch\Api\Data\IndexOptionsInterface;
+use Algolia\AlgoliaSearch\Exceptions\AlgoliaException;
+use Magento\Framework\Exception\NoSuchEntityException;
+
+class IndexSettingsDiffChecker
```
cammonro: We're not actually returning a diff anywhere - it's only used internally to this class - I'm wondering if a name like `IndexSettingsComparator` might be cleaner?

Service/IndexSettingsDiffChecker.php
```diff
@@ -0,0 +1,42 @@
...
+        $this->recursiveKSort($algoliaSettings);
+
+        if (hash('sha256', json_encode($indexSettings)) === hash('sha256', json_encode($algoliaSettings))) {
+            return true;
+        }
+
+        return false;
+    }
+
+    protected function recursiveKSort(array &$array): void
```
cammonro: Small nitpick but could we rename this to something like `canonicalize` to make the intent of the function clearer?

Service/IndexSettingsDiffChecker.php
```diff
@@ -0,0 +1,42 @@
...
+     * @throws NoSuchEntityException
+     * @throws AlgoliaException
+     */
+    public function matchAlgoliaSettings(IndexOptionsInterface $indexOptions, array $indexSettings): bool
+    {
+        $this->recursiveKSort($indexSettings);
+
+        $algoliaSettings = $this->connector->getSettings($indexOptions);
+        $algoliaSettings = array_intersect_key($algoliaSettings, $indexSettings);
+        $this->recursiveKSort($algoliaSettings);
```
cammonro: I think we should combine the canonicalization (ksort) and hashing (sha256) into a single function that produces a comparable artifact. This would provide a cleaner comparison to reason about.

Service/IndexSettingsDiffChecker.php
```diff
@@ -0,0 +1,42 @@
...
+
+        if (hash('sha256', json_encode($indexSettings)) === hash('sha256', json_encode($algoliaSettings))) {
```
cammonro: I'm concerned about relying on `json_encode` defaults here. There is risk of silent failure. 

For instance, `json_encode` can return false on something like an encode error. If both sides return false for whatever reason that would hash out equally and produce a false positive ... which means a settings change would get skipped. 

I think we should at least pass `JSON_THROW_ON_ERROR` so encoding failures surface instead of masquerading as "no diff". Can we mock this as well in our unit tests?

Service/IndexSettingsDiffChecker.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    public function __construct(
+        protected AlgoliaConnector $connector,
+    ) {}
+
+    /**
+     * @throws NoSuchEntityException
+     * @throws AlgoliaException
+     */
+    public function matchAlgoliaSettings(IndexOptionsInterface $indexOptions, array $indexSettings): bool
```
cammonro: What do you think of simplifying this to just `matches`?
I think this reads pretty nicely:
```
if ($this->indexSettingsComparator->matches($indexOptions, $indexSettings)) { ... skip ... }
```

Service/IndexSettingsDiffChecker.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        $this->recursiveKSort($indexSettings);
+
+        $algoliaSettings = $this->connector->getSettings($indexOptions);
```
cammonro: Since we're introducing an additional GET request here - do you think the same settings might get called multiple times? If so maybe a cache might be called for? 

... but that means we need to invalidate as well. If everything runs through `IndexSettingsHandler` it can be the gatekeeper and ensure that `invalidate` gets called. Something to keep in mind as we incorporate the checks across the board.

Service/IndexSettingsHandler.php
cammonro, line 36: Could we modify the return value of `IndexSettingsHandler::setSettings` from `void` to `bool` to indicate whether the change was applied? This could be useful for callers to make decisions based on this operation. This would not break anything because any dependent code would not be expecting a return value and it would be silently ignored. 

I think this could be a valuable addition especially if the caller has to decide whether to perform something like `waitLastTask`.

cammonro (changes requested):
Nice work! This is a solid implementation. 

I've offered a few suggestions with one key concern related to how we're performing the comparison operations.

Helper/Entity/SuggestionHelper.php
damcou: I've tested all the configs I've got on my side and this is the only one I spotted yes.

Service/IndexSettingsDiffChecker.php
damcou: For me "checking the diff" was literally what the class is doing 😄 , but I'm ok with `IndexSettingsComparator` which is fine as well.

Service/IndexSettingsDiffChecker.php
damcou: Maybe this is a lack of technical language on my side but `canonicalize` makes less sense than `normalize` which seems easier to understand to me; so I'll go with this naming if you're ok with it.

cammonro: I would say canonicalizing is a more specific form of normalizing. Both involve transforming something into a "better" form - but canonicalize is saying this is the "official" thing to compare. Either is fine! Thanks for humoring me. 🙃

Service/IndexSettingsDiffChecker.php
cammonro: I'm probably just being an architectural snob and think "Comparator" sounds cooler. Your call!

Service/IndexSettingsDiffChecker.php
damcou: For now I didn't see any scenario where we fetched the same settings more than once in the process (it iterates over the different indices once), so I'd keep it simple for now (adding cache and invalidation seems a little bit overcomplicated for this). But I'll make sure to pay attention to this in the following tasks.

Service/IndexSettingsDiffChecker.php
damcou: Updated the `json_encode` process, do you have a precise unit test for this ?

Service/IndexSettingsDiffChecker.php
cammonro: Sounds good to me. Yes if that's the case - totally agree to keep it simple. Let's not optimize before we need it!

Service/IndexSettingsDiffChecker.php
cammonro: I don't think `json_encode` can serialize `NAN`/`INF` so if you test an object with something like that it should force a `JsonException`. If you want an example let me know and I can update the PR!

Service/IndexSettingsComparator.php
```diff
@@ -0,0 +1,59 @@
...
+     * @throws AlgoliaException
+     */
+    protected function getSettingsHash(array $settings): string
+    {
+        $this->normalize($settings);
+
+        try {
+            $jsonSettings = json_encode($settings, JSON_THROW_ON_ERROR);
+        } catch (\JsonException $e) {
+            throw new AlgoliaException('Invalid JSON : ' . $e->getMessage());
```
cammonro: Might not be a bad idea to pass the original exception to the constructor? 

```suggestion
            throw new AlgoliaException('Invalid JSON: ' . $e->getMessage(), 0, $e);
```

Service/IndexSettingsDiffChecker.php
damcou: Added a unit test ;)  Thanks for the suggestion !

Test/Unit/Service/IndexSettingsComparatorTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+    public function testInvalidJson(): void
+    {
+        $algoliaSettings = [INF];
+
+        $this->connector->expects($this->once())->method('getSettings')->willReturn($algoliaSettings);
+        // Must be different
+
+        $this->expectException(AlgoliaException::class);
+        $this->expectExceptionMessage("Invalid JSON: Inf and NaN cannot be JSON encoded");
```
cammonro: Looks good to me! 

Test string is highly specific - not sure if that will break in the future? This might be safer: 

```suggestion
        $this->expectExceptionMessageMatches('/Invalid JSON/');
```

If we implement [this suggestion](https://github.com/algolia/algoliasearch-magento-2/pull/1933#discussion_r3155131883) we can also test the exception chain:

```
$this->assertInstanceOf(\JsonException::class, $e->getPrevious());
```

But these are just grace notes.

cammonro (approved):
Looks great! Added a couple more suggestions but not blocking at all. Approved! 🚀

Service/IndexSettingsComparator.php
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * @throws NoSuchEntityException
+     * @throws AlgoliaException
+     */
+    public function matches(IndexOptionsInterface $indexOptions, array $indexSettings): bool
+    {
+        $algoliaSettings = $this->connector->getSettings($indexOptions);
+        $algoliaSettings = array_intersect_key($algoliaSettings, $indexSettings);
+
+        if ($this->getSettingsHash($indexSettings) === $this->getSettingsHash($algoliaSettings)) {
```
cammonro: Can this be simplified? 

`return $this->getSettingsHash($indexSettings) === $this->getSettingsHash($algoliaSettings)`

PR #1936, opened 2026-04-28
MAGE-1565 remove deprecated template beforecontent.phtml (3.18)

cammonro:
**Summary**

Removes the deprecated `beforecontent.phtml` template, originally deprecated in v3.16.

The template body has been empty since v3.16, so removing it has no runtime behavior change. Anyone still overriding the template path may see a logged Magento layout warning.


damcou (approved):
LGTM 🚀

PR #1938, opened 2026-04-29
MAGE-1569: conditional `SetSettings()` calls applied to `saveConfigurationToAlgolia()`

damcou:
This PR contains:
- `saveConfigurationToAlgolia()`and `ProductHelper::getSettings()` now perform `setSettings()`operations based on the results from the `IndexSettingsComparator`.
- `waitLastTask()` is now only call when `setSettings()`operations are performed
- Refactored the whole `IndicesConfigurator` class so it's now easier to read
- Refactoring Diagnostic logging as a result

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -319,17 +319,51 @@ class ProductHelper extends AbstractEntityHelper
     ): void {
         $indexSettings = $this->getIndexSettings($storeId);
 
-        $this->indexSettingsHandler->setSettings($indexOptions, $indexSettings);
+        if ($this->indexSettingsHandler->setSettings($indexOptions, $indexSettings)) {
+            $logEventName = 'Pushing settings for products indices.';
+            $this->logger->start($logEventName, true);
+            $this->logger->log('Index name: ' . $indexOptions->getIndexName());
+            $this->logger->log('Settings: ' . json_encode($indexSettings));
+            $this->logger->stop($logEventName, true);
 
-        $this->logger->log('Settings: ' . json_encode($indexSettings));
-        if ($saveToTmpIndicesToo) {
-            $this->indexSettingsHandler->setSettings(
-                $indexTmpOptions,
-                $indexSettings,
-                $indexOptions->getIndexName()
-            );
+            $this->algoliaConnector->waitLastTask($storeId);
 
-            $this->logger->log('Pushing the same settings to TMP index as well');
+            if ($saveToTmpIndicesToo) {
+                $this->indexSettingsHandler->setSettings($indexTmpOptions, $indexSettings, $indexOptions->getIndexName());
+
+                $logEventName = 'Pushing the same settings to TMP index as well';
+                $this->logger->start($logEventName, true);
+                $this->logger->log('Index name: ' . $indexOptions->getIndexName());
+                $this->logger->log('TMP Index name: ' . $indexTmpOptions->getIndexName());
+                $this->logger->log('Settings: ' . json_encode($indexSettings));
+                $this->logger->stop($logEventName, true);
+
+                $this->algoliaConnector->waitLastTask($storeId);
+            }
+
+            if ($saveToTmpIndicesToo) {
+                try {
+                    $this->algoliaConnector->copySynonyms($indexOptions, $indexTmpOptions);
+                    $this->algoliaConnector->waitLastTask($storeId);
+                    $this->logger->log('
+                        Copying synonyms from production index to "' . $indexTmpOptions->getIndexName() . '" to not erase them with the index move.
+                    ');
+                } catch (AlgoliaException $e) {
+                    $this->logger->error('Error encountered while copying synonyms: ' . $e->getMessage());
+                }
+
+                try {
+                    $this->algoliaConnector->copyQueryRules($indexOptions, $indexTmpOptions);
+                    $this->algoliaConnector->waitLastTask($storeId);
+                    $this->logger->log('
+                        Copying query rules from production index to "' . $indexTmpOptions->getIndexName() . '" to not erase them with the index move.
+                    ');
+                } catch (AlgoliaException $e) {
+                    if ($e->getCode() !== 404) {
+                        throw $e;
+                    }
+                }
+            }
         }
 
         $this->setFacetsQueryRules($indexOptions);
@@ -341,30 +375,6 @@ class ProductHelper extends AbstractEntityHelper
         }
 
         $this->replicaManager->syncReplicasToAlgolia($storeId, $indexSettings);
-
-        if ($saveToTmpIndicesToo) {
-            try {
-                $this->algoliaConnector->copySynonyms($indexOptions, $indexTmpOptions);
-                $this->algoliaConnector->waitLastTask($storeId);
-                $this->logger->log('
-                        Copying synonyms from production index to "' . $indexTmpOptions->getIndexName() . '" to not erase them with the index move.
-                    ');
-            } catch (AlgoliaException $e) {
-                $this->logger->error('Error encountered while copying synonyms: ' . $e->getMessage());
-            }
-
-            try {
-                $this->algoliaConnector->copyQueryRules($indexOptions, $indexTmpOptions);
-                $this->algoliaConnector->waitLastTask($storeId);
-                $this->logger->log('
-                        Copying query rules from production index to "' . $indexTmpOptions->getIndexName() . '" to not erase them with the index move.
-                    ');
-            } catch (AlgoliaException $e) {
-                if ($e->getCode() !== 404) {
-                    throw $e;
-                }
-            }
-        }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


Helper/Entity/ProductHelper.php
```diff
@@ -319,52 +319,62 @@ public function setSettings(
...
-        $this->algoliaConnector->waitLastTask($storeId);
-
-        if ($saveToTmpIndicesToo) {
-            $this->setFacetsQueryRules($indexTmpOptions);
             $this->algoliaConnector->waitLastTask($storeId);
-        }
 
-        $this->replicaManager->syncReplicasToAlgolia($storeId, $indexSettings);
+            if ($saveToTmpIndicesToo) {
+                $this->indexSettingsHandler->setSettings($indexTmpOptions, $indexSettings, $indexOptions->getIndexName());
```
cammonro: I'm realizing that our comparator is [not actually taking into account `$mergeSettingsFrom`](https://github.com/algolia/algoliasearch-magento-2/blob/471f9c0959461bf9dfbff41f7bf4372c497532f2/Service/IndexSettingsHandler.php#L43).

This could lead to an unintentional skip. We need to hoist the [`mergeSettings`](https://github.com/algolia/algoliasearch-magento-2/blob/471f9c0959461bf9dfbff41f7bf4372c497532f2/Service/AlgoliaConnector.php#L390) out of `AlgoliaConnector` into a separate composable class I think - because the merge needs to happen *before* the comparator sees it.

Helper/Entity/ProductHelper.php
```diff
@@ -319,52 +319,62 @@ public function setSettings(
...
+                $this->logger->start($logEventName, true);
+                $this->logger->log('Index name: ' . $indexOptions->getIndexName());
+                $this->logger->log('TMP Index name: ' . $indexTmpOptions->getIndexName());
+                $this->logger->log('Settings: ' . json_encode($indexSettings));
+                $this->logger->stop($logEventName, true);
 
-        if ($saveToTmpIndicesToo) {
-            try {
-                $this->algoliaConnector->copySynonyms($indexOptions, $indexTmpOptions);
                 $this->algoliaConnector->waitLastTask($storeId);
```
cammonro, line 342: Any way we can reduce the extra calls to `waitLastTask`? This can really bottleneck things.

Helper/Entity/ProductHelper.php
```diff
@@ -319,52 +319,62 @@ public function setSettings(
...
-            $this->setFacetsQueryRules($indexTmpOptions);
             $this->algoliaConnector->waitLastTask($storeId);
-        }
 
-        $this->replicaManager->syncReplicasToAlgolia($storeId, $indexSettings);
+            if ($saveToTmpIndicesToo) {
+                $this->indexSettingsHandler->setSettings($indexTmpOptions, $indexSettings, $indexOptions->getIndexName());
+
+                $logEventName = 'Pushing the same settings to TMP index as well';
+                $this->logger->start($logEventName, true);
```
cammonro: By reducing the logging scope we're no longer capturing the API calls (which is what we really want to measure) so the logger will not reveal any meaningful timing. We need to make sure we're wrapping the full settings flow in a logging event.

cammonro (changes requested):
Moving in the right direction here - but it looks like we have some scenarios that need addressing.

Helper/Entity/ProductHelper.php
damcou: Replying on JIRA.

Helper/Entity/ProductHelper.php
damcou: Ah yes you're right, I oversimplified it, gonna rollback those changes while keeping some readability improvements.

Helper/Entity/ProductHelper.php
damcou: This is indeed a little bit tricky, the only place where we use the `$mergeSettingsFrom` parameter is when we want to apply settings on the product temporary index (in `ProductHelper::setSettings()` ) and this is already depending on if we skip the products main index setSettings or not. 
Is there a scenario (other than a manual update in the Algolia dashboard) where settings would be different in the temp index and would need to be updated by the process even though we don't update the main index ?

cammonro: I think the need for the merge settings op is that there could be settings in Algolia (not in Magento) that might be lost during the copy to the temp index without a merge. We do have a task in our epic to revisit this settings copy logic - and maybe that's where these decisions belong - but regardless our method contracts in our comparator and handler are silently ignoring the parameter so we will need to refactor this somehow. As it stands it's technically broken.

cammonro: One other thought is that maybe we don't need to diff ever on the temp because it's always going to be newly created. The revisited copy operation (future) could solve this?

damcou: Replied on JIRA 😉

Helper/Entity/ProductHelper.php
```diff
@@ -319,52 +319,56 @@ public function setSettings(
...
-        $this->algoliaConnector->waitLastTask($storeId);
-
-        if ($saveToTmpIndicesToo) {
-            $this->setFacetsQueryRules($indexTmpOptions);
             $this->algoliaConnector->waitLastTask($storeId);
-        }
 
-        $this->replicaManager->syncReplicasToAlgolia($storeId, $indexSettings);
+            if ($saveToTmpIndicesToo) {
+                // For temp index, we need to call AlgoliaConnector::setSettings() directly with explicit settings merge
```
cammonro, line 329: I agree this is worth a callout - but can this comment expand on the reason for this explicit settings call and why we are not using the IndexSettingsHandler? e.g. No replicas on temp index and no diff to perform.

Service/IndexSettingsHandler.php
```diff
@@ -56,30 +52,31 @@ public function setSettings(
             $this->connector->setSettings(
                 $indexOptions,
                 $indexSettings,
-                false,
-                true,
-                $mergeSettingsFrom
+                false
```
cammonro, line 55: This is a sensible change - if merge settings behavior is complicating our handler needlessly then let's not even handle it. 👍

Model/IndicesConfigurator.php
```diff
@@ -233,19 +225,12 @@ protected function setExtraSettings(int $storeId, bool $saveToTmpIndicesToo): vo
...
-                    $this->logger->log('Extra settings: ' . json_encode($extraSettings));
-
-                    $this->algoliaConnector->setSettings(
-                        $indexOptions,
-                        $extraSettings,
-                        true,
-                        false
-                    );
-                    $this->algoliaConnector->waitLastTask($storeId);
+                    if ($this->indexSettingsHandler->setSettings($indexOptions, $extraSettings)) {
```
cammonro, line 230: We should align with our new direction on how we apply settings to the temp index. i.e. explicit / no handler

Test/Unit/Service/IndexSettingsHandlerTest.php
cammonro, line 1: It's great that you're adding these tests.I'm wondering if we should seize this opportunity to also add unit tests on `ProductHelper` for thoroughness. Not the whole class but just the paths that run through these scenarios we're addressing.

cammonro (changes requested):
I agree with your decisions here - just a few more observations on things I think we can tighten a bit.

Helper/Entity/ProductHelper.php
```diff
@@ -482,11 +486,12 @@ protected function getUnretrieveableAttributes($storeId = null)
 
     /**
      * @param IndexOptionsInterface $indexOptions
+     * @param bool $waitLastTask
      * @return void
      * @throws AlgoliaException
      * @throws NoSuchEntityException
      */
-    protected function setFacetsQueryRules(IndexOptionsInterface $indexOptions)
+    public function setFacetsQueryRules(IndexOptionsInterface $indexOptions, bool $waitLastTask = false): void
```
cammonro: I see we've opened the API here for the integration test. Are we sure we want to do this? Or can the test use reflection perhaps?

Service/IndexSettingsHandler.php
damcou: Ok so from what I understand, this is fine as it is ? (see JIRA ;) )

Model/IndicesConfigurator.php
damcou: Good catch !

Helper/Entity/ProductHelper.php
damcou: Yeah, I thought having a public method exposed in the helper to create facet query rules would make sense, but indeed, this is a little bit weird to use it only in test, gonna use `invokeMethod` instead 👌

Service/IndexSettingsHandler.php
cammonro: Indeed! 😁

cammonro (approved):
Bravo! Approuvé! 👏

PR #1939, opened 2026-05-02
Fix serialization of Closure error in Job Queue view

angelvilaplana:
**Summary**

This PR fixes a "Uncaught Exception: Serialization of 'Closure' is not allowed" error that occurs when viewing specific jobs in Indexing Queue.

Previously, the AbstractAction was saving the entire Job model into the backend session. In Magento 2, models often contain dependencies (via plugins or resource models) that include anonymous functions (Closures). 

Since PHP cannot serialize Closures, the SessionManager would crash during session_write_close(). Because session writing occurs at the end of the PHP lifecycle, the Job View page would load the main content successfully, but then throw the exception in the footer/bottom of the page, preventing a clean response.

image

<br>

Changes:
- Removed SessionManagerInterface dependency from View block and AbstractAction.
- Modified getCurrentJob() in the View block to load the model dynamically using the id parameter from the Request.

**Result**

The Job View page now loads successfully without triggering a session serialization exceptions.

Testing performed:

1. Navigated to Stores > Algolia Search > Indexing Queue
2. Clicked on "View" for an existing job.
3. Verified the page renders correctly and the job details are displayed without errors.

Commands run:

- `bin/magento setup:di:compile` (to verify constructor injection)
- `bin/magento cache:clean`

### DIFF
--- a/Block/Adminhtml/Job/View.php
+++ b/Block/Adminhtml/Job/View.php
@@ -2,28 +2,35 @@
 
 namespace Algolia\AlgoliaSearch\Block\Adminhtml\Job;
 
+use Algolia\AlgoliaSearch\Model\JobFactory;
+use Algolia\AlgoliaSearch\Model\ResourceModel\Job as JobResource;
 use Magento\Backend\Block\Widget\Button;
-use Magento\Framework\Session\SessionManagerInterface;
 use Magento\Framework\View\Element\Template;
 use Magento\Framework\View\Element\Template\Context;
 
 class View extends Template
 {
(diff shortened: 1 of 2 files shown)


cammonro:
Thank you @angelvilaplana for sharing your solution with us. Your diagnosis was spot on. `Model\Job` carries an injected `ObjectManagerInterface` for dynamic class loading, so storing it in the session could never serialize. We'll be incorporating your fix into our upcoming release!

angelvilaplana:
Thanks @cammonro! I'm glad I could help! 🙂

PR #1940, opened 2026-05-06
MAGE-1584: Add standardized unit tests generation

damcou:
This PR contains:
- Added missing instruction files for generating unit tests with Claude for types:  blocks, controllers, models, plugins, view models, others.
- Updated `/generate-unit-tests %FILE%` skill
- Created `/generate-all-unit-tests-for-type %TYPE%` skill for mass generation
- Added block generated unit tests
- Added plugin generated unit tests
- CI now runs Magento 2.4.8, PHP v8.4 and PHPUnit 10.
- Claude's instructions have been adjusted to generate unit tests compatible with PHPUnit 12 but backward compatible with PHPUnit 10 & 11.


.claude/skills/generate-all-unit-tests-for-type/SKILL.md
```diff
@@ -0,0 +1,128 @@
...
+```bash
+cd <magento_root> && vendor/bin/phpunit -c dev/tests/unit/phpunit.xml.dist \
+  app/code/Algolia/AlgoliaSearch/Test/Unit/<TypeTestDir>/
+```
+
+Where `<TypeTestDir>` mirrors the source directory (e.g. `Block/` for blocks, `Plugin/` for plugins).
+
+For types that span multiple directories (e.g. `model` covers `Model/Indexer/`, `Model/Backend/`, `Model/Source/`), run a single phpunit invocation passing all test directories.
+
+Report the full test output. If tests fail, list each failing test with its error message. Do **not** automatically fix failing tests — report them to the user and stop.
```
cammonro, line 128: This is a good safety measure. Can we expand the guidelines since we are dealing with legacy code? It's quite possible that some old edge cases were never fully thought through and blindly asserting "current behavior" turns a unit test into a bug that is now locked in by a passing test. (In other words - the old code could be wrong!)

My suggestion is that when the intended behavior of a path is unclear or looks possibly unintended, the agent must not silently assert it as correct and instead should flag it in some way at both the test level and in the final summary.

.claude/testing-core.md
```diff
@@ -13,7 +13,12 @@
 ### Test Strategy
 - Don't test Event Manager's dispatch methods
 - Use @dataProvider annotations when it's needed
+  - use /**
+        * @dataProvider myProvider
+        */
+  - not #[DataProvider('myProvider')]
```
cammonro: This seems contradictory with [other instructions](https://github.com/algolia/algoliasearch-magento-2/blob/567336145d40af601e607796e78ab9edf5703485/.claude/types/testing-services.md?plain=1#L35). Since we're moving to PHPUnit 12 we should standardize on the newer attribute based approach.

.claude/specifics/testing-Configuration.md
```diff
@@ -0,0 +1,48 @@
...
+## What to test
+
+### Methods with pure logic (directly testable)
+
+- **`areCategoriesInFacets(array $facets)`** — Pure: checks if `'categories'` is in the `'attribute'` column. Use a `@dataProvider`.
+- **`getUrlTrackedParameters()`** — Has one branch: appends `'page'` only when infinite scroll is disabled. Two test cases.
+
+### Methods that require a partial mock
+
+For methods that call `$this->getRequest()`, `$this->getCurrentCategory()`, etc., use `getMockBuilder` with `setMethods()` to override only the inherited methods, leaving the logic under test intact:
```
cammonro: PHPUnit 12 in Magento 2.4.9 will definitely mean a new strategy unfortunately. 

Can we target this skill for PHPUnit 10-12 as we seek to straddle Magento versions? We're [pretty strict here](https://github.com/algolia/algoliasearch-magento-2/blob/567336145d40af601e607796e78ab9edf5703485/.claude/testing-core.md?plain=1#L9-L11) and this is subject to rapid drift. I believe PHPUnit 10 can neatly bridge to PHPUnit 12 if we abandon PHPUnit 8 entirely.

The [addMethods](https://github.com/algolia/algoliasearch-magento-2/blob/567336145d40af601e607796e78ab9edf5703485/.claude/skills/generate-unit-tests/SKILL.md?plain=1#L61) is removed in PHPUnit 12 ([#5321](https://github.com/sebastianbergmann/phpunit/issues/5321)). `setMethods` was removed in 10.0.0 ([#3769](https://github.com/sebastianbergmann/phpunit/issues/3769)). 

`getMockBuilder()` remains in [12.5](https://docs.phpunit.de/en/12.5/test-doubles.html) but is now documented as an "edge-case" escape hatch behind `createMock()` / `createStub()` / `createPartialMock()` / `createMockForIntersectionOfInterfaces()`.

.claude/reference/testing/core.md
cammonro, line 1: I feel your skill is doing a nice job of being mostly compliant with the [official guidelines](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices#progressive-disclosure-patterns) but I think we need to tighten the supporting agent instructions placement. 

The guidelines don't seem to discuss shared docs like you've done (which seems a perfectly valid pattern to me) but maybe we can just organize using the recommended `reference` directory naming convention?
```
.claude/
├── reference/
│   └── testing/
│       ├── core.md
│       ├── types/
│       └── specifics/
└── skills/
    ├── generate-unit-tests/SKILL.md
    └── generate-all-unit-tests-for-type/SKILL.md

```

Test/TestCase.php
```diff
@@ -53,9 +60,15 @@ protected function setPrivateProperty(object $obj, string $prop, mixed $value):
      */
     protected function getPrivateProperty(object $obj, string $prop): mixed
     {
-        $reflection = new \ReflectionClass($obj);
-
-        return $reflection->getProperty($prop)->getValue($obj);
+        $ref = new \ReflectionClass($obj);
```
cammonro, line 61: This is a cool addition but rather than duplicate the code can we call a shared helper to obtain the correct ref?

Test/TestCase.php
```diff
@@ -70,8 +83,6 @@ protected function getPrivateProperty(object $obj, string $prop): mixed
     protected function mockProperty(object $object, string $propertyName, string $propertyClass): void
     {
         $mock = $this->createMock($propertyClass);
-        $reflection = new \ReflectionClass($object);
-        $property = $reflection->getProperty($propertyName);
-        $property->setValue($object, $mock);
+        $this->setPrivateProperty($object, $propertyName, $mock);
```
cammonro, line 85: Good catch!

.claude/reference/testing/core.md
cammonro: What do you think about adding a rule to extend `Algolia\AlgoliaSearch\Test\TestCase` to ensure reflection utilities are always available when needed? On a related note, I've eliminated the `*Testable` pattern [here](https://github.com/algolia/algoliasearch-magento-2/pull/1956) but it would be good if Claude can help with consistency.

.claude/skills/generate-all-unit-tests-for-type/SKILL.md
```diff
@@ -0,0 +1,128 @@
...
+
+Read both files **once** before processing any individual file. These rules apply to every file in the batch.
+
+## Step 6 — Generate tests for each file
+
+Process files one at a time, in order. For each file:
+
+1. **Check for class-specific rules**: extract the class name (e.g. `Configuration` from `Block/Configuration.php`) and check whether `.claude/specifics/testing-<ClassName>.md` exists. If it does, read it — its rules override the type guide for this file only.
+2. **Read the source file**: understand constructor dependencies, public methods, guard clauses, and any use of `addCommitCallback` or external service calls.
+3. **Check for an existing test file**: if one exists, read it first to avoid duplicating tests that are already written.
```
cammonro, line 95: I like the effort to prevent duplicate tests! One principle worth adding to `testing-core.md`: **test a shared behavior once, where it lives.** 

e.g. If several classes call the same helper, the helper's behavior should be verified in the helper's own test, not re-verified through each wrapper. 

A wrapper's test should only assert the logic the wrapper itself adds and should mock the helper. This both reduces test count and stops a single helper change from breaking ten wrapper tests. It also gives the agent a cleaner answer to "where does this assertion go."

.claude/reference/testing/types/testing-models.md
cammonro, line 1: Does this replace `types/types-indexers.md` ?

.claude/reference/testing/specifics/testing-ConfigHelper.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Skip these
+
+All methods that only call `$this->configInterface->getValue(...)` and return the result (possibly cast). These are just reading config — nothing to assert that isn't already mocked. Examples: `getApplicationID()`, `getAPIKey()`, `getSearchOnlyAPIKey()`, `getIndexPrefix()`, `isLoggingEnabled()`, etc.
+
+---
+
+## What to test
+
+### `isEnabledFrontEnd()`
```
cammonro: I'm concerned about this kind of specificity in our rules. A lot of this stuff seems derivable from the code the agent is already reading, and all of it rots the moment the code changes. The risk here is that that agent will trust it over the source and hallucinate a wrong outcome.

cammonro (commented):
Nice job! This is a genuinely strong piece of work. The layered design (core rules → per-type guides → class-specific notes, shared across two separate skills) is solid! 💪 

Several of the principles are excellent. Some things I liked:

- not testing the Event Manager dispatch
- skipping pure delegation
- never write a test that just returns what we mock
- testing observable behavior over implementation details
- reading an existing test file before generating 
- confirming before mass generation

I added detailed feedback but at a high level here are some things I'd like us to be able to address:

- The platform shift to PHPUnit 12
- Less specificity and more strategy - staleness can bite us if we're not careful
- The assumption that the legacy code is right - let's let the agent flag unusual edge case results!
- Find behaviors to test at their source - another way the agent can help
- Compliance with [Anthropic guidelines](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices) (you're practically there already IMO) 

(Additional comments in Jira)

.claude/skills/generate-all-unit-tests-for-type/SKILL.md
damcou: I'll try to phrase that 👍

.claude/testing-core.md
damcou: Indeed ! This was written before PHPUnit 12 switch, I'll change this

.claude/reference/testing/core.md
damcou: It was specified in the skill files but you're right, this can be located in `core.md`

.claude/reference/testing/types/testing-models.md
damcou: Agreed, this doesn't seems useful since there's already a part related to indexers there ... removing `types-indexers.md` to simplify

.claude/skills/generate-unit-tests/SKILL.md
```diff
@@ -48,20 +54,21 @@ Rules:
 - **Any additional tests per method should be driven by its cyclomatic complexity**
 - **Focus on user-observable behavior**: what the method returns, what exception it throws, which collaborator it calls with which arguments
 - **Do not test implementation details**: internal variable names, call counts on low-level helpers, private method logic
-- **Use `@dataProvider`** (or `#[DataProvider]` attribute) when testing the same behavior across multiple input variants
+- **Don't check helpers logic in the class itself** — if the class delegates to a helper, the helper's behavior should be verified in the helper's own test, not re-verified here.
+- **Use `#[DataProvider]` attribute when testing the same behavior across multiple input variants
```
cammonro, line 58: It looks like we have few more occurrences of the old `@dataProvider` attribution that we will need to clean up - here are some I found:
- `testing-services.md`
- `testing-others.md`
- `testing-Configuration.md`
- `testing-ProductWithoutChildren.md`

.claude/reference/testing/specifics/testing-Configuration.md
```diff
@@ -0,0 +1,52 @@
...
+
+`Configuration` extends the custom `Algolia` base block, which itself wraps Magento's `Template`. Most block infrastructure (`getRequest()`, `getCurrentCategory()`, `getCurrentProduct()`, etc.) is inherited via protected methods — these cannot be mocked by simply swapping constructor dependencies.
+
+---
+
+## What to test
+
+### Methods with pure logic (directly testable)
+
+- **`areCategoriesInFacets(array $facets)`** — Pure: checks if `'categories'` is in the `'attribute'` column. Use a `@dataProvider`.
```
cammonro: Old attribution method

.claude/reference/testing/specifics/testing-ProductWithoutChildren.md
```diff
@@ -0,0 +1,71 @@
...
+
+## What to test
+
+### `getFields()` (protected, use `invokeMethod()`)
+Three branches based on `taxHelper->getPriceDisplayType()`:
+- `DISPLAY_TYPE_EXCLUDING_TAX` → returns `['price' => false]`
+- `DISPLAY_TYPE_INCLUDING_TAX` → returns `['price' => true]`
+- Any other value → returns `['price' => false, 'price_with_tax' => true]`
+
+Use `@dataProvider` for the three cases.
```
cammonro: Old attribution method

.claude/reference/testing/types/testing-others.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+---
+
+## Validators (`Validator/`)
+
+Pure logic classes — take input, validate, return result or throw exception.
+
+**What to test:**
+- All distinct valid/invalid states (use `@dataProvider` when there are multiple variants of invalid input).
```
cammonro: Old attribution method

.claude/reference/testing/core.md
```diff
@@ -6,21 +6,31 @@
...
-- Use PHPUnit 10
-- PHP 8.3
-- Magento 2.4.8
+- Use PHPUnit 12
+- PHP 8.5
+- Magento 2.4.9
 
 ### Test Strategy
 - Don't test Event Manager's dispatch methods
 - Use @dataProvider annotations when it's needed
```
cammonro, line 15: Inconsistent description of the attribution method

Test/Unit/Logger/DiagnosticLoggerTest.php
```diff
@@ -7,7 +7,9 @@
 use Algolia\AlgoliaSearch\Logger\TimedLogger;
 use Algolia\AlgoliaSearch\Service\StoreNameFetcher;
 use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Attributes\AllowMockObjectsWithoutExpectations;
 
+#[AllowMockObjectsWithoutExpectations]
```
cammonro, line 12: I couldn't find instructions to the agent about this but is this a transitionary step to get us on PHPUnit 12? If so we should codify that in our references with a view towards better testing discipline. I don't want the agent to commit to old PHPUnit 8 practices where the line between [mocks and stubs are blurred](https://phpunit.expert/articles/testing-with-and-without-dependencies.html).

cammonro: Maybe something in `core.md` like `createMock()` without expectations is a defect to be migrated to `createStub()`, that the attribute is a temporary escape hatch. We could require a justifying comment, and favor method-scope over class-level application?

.claude/reference/testing/core.md
```diff
@@ -6,21 +6,31 @@
...
+- We still want to be backward compatible with PHPUnit 10 and PHPUnit 11. Do not use any classes that are not available for these versions.
+- Extend `\Algolia\AlgoliaSearch\Test\TestCase` (not PHPUnit's TestCase directly) so inherited helpers like `invokeMethod()`, `setPrivateProperty()`, and `getPrivateProperty()` are available
 
 ### Mocking Strategy
 - Instantiate the tested class in the setUp() method by mocking every object that needs to be passed in its constructor
 - Use nullable types for the testing class properties. ex:
     - protected ?ConfigHelper $configHelper;
       or
     - protected null|(CategoryResourceModel&MockObject) $categoryResource = null;
+- Prefer usage of `createStub` over `createMock`.
```
cammonro: I feel we need some more nuance here - our goal should be to align with the spirit of PHPUnit 12 not just silence errors. 
```suggestion
- Choose the double by the role the collaborator plays in the test:

  - Use createStub() when the collaborator only feeds canned data into the class under test (indirect input). Assert on what the class returns or does, never on the stub.
  - Use createMock() with expects() only when the call itself is the behavior under test (indirect output: dispatch, persist, log, an outbound API call). The interaction is the contract you are proving.
  - Default a collaborator to a stub. Promote it to a mock only when you are asserting the interaction. 
```

cammonro: We also should probably rework [our example](https://github.com/algolia/algoliasearch-magento-2/blob/a9227b2c12663022d871eba5f12b7f51ab028eb7/.claude/reference/testing/core.md?plain=1#L54-L56) to demonstrate the different scenarios. I worry this kind of contradiction might produce undesirable results with the agent.

.claude/skills/generate-unit-tests/SKILL.md
```diff
@@ -48,20 +54,21 @@ Rules:
...
-- Extend `\Algolia\AlgoliaSearch\Test\TestCase` (not PHPUnit's TestCase directly) so inherited helpers like `invokeMethod()`, `setPrivateProperty()`, and `getPrivateProperty()` are available
 - Declare strict types
 - Use the same namespace structure as the module (e.g. `Algolia\AlgoliaSearch\Test\Unit\Model\Indexer`)
 - Import all used classes at the top
 
 After writing the file, report:
 - The output file path
 - A one-line description of each test and what behavior it covers
+
+Some parts of the codebase are very old and cannot be blindly considered as valid. When behavior looks unclear or possibly unintended, flag it both in the test (as a comment) and in the final summary.
```
cammonro: I'm worried the agent might miss this or interpret loosely. Without being specific we'll get different comments from run to run.

This might be better surfaced in `core.md` to be shared by both skills? 
We could add a "characterization-test protocol" section with more detailed guidance - maybe something like: 

> Cover every reachable path that produces observably different behavior, but when a path's intended behavior is unclear or looks unintended, the agent must NOT silently assert it as correct. Instead comment with 
`// CHARACTERIZATION: encodes current behavior, needs confirmation` and surface it in the run summary for human confirmation. 

The advantage of something like this is that we can grep the code and it will remain the same across every agent run.

cammonro (changes requested):
You did a great job here tightening this up! 🙌 

I've provided further notes for hardening the approach. 

A few more edits I think are worth doing before merge, only because the skills are the generator and these shape every test it writes next (the `createStub` nuance, finishing the `@dataProvider` to `#[DataProvider]` sweep, and a line codifying how we treat `AllowMockObjectsWithoutExpectations` would be good) - as well as promoting a consistent characterization flag into `core.md`.

Beyond this I think we're good. We can iterate and improve as we monitor the results of the generator.

.claude/skills/generate-unit-tests/SKILL.md
damcou: Good catch ! Fixing this !

cammonro (approved):
I like the additions! Let's get this merged and we can continue to refine.

PR #1941, opened 2026-05-19
MAGE-1570: Apply new settings comparator to batch processing

damcou:
This PR contains:
- Add `IndexSettingsComparator` to batch processing for entities that are performing `setSettings` operation on full indexing (products and categories) : a settings comparison is now performed before adding `saveConfigurationToAlgolia` operations in the queue process (and direct indexing when queue is disabled). 
- When queue is enabled `Product/IndexBuilder` now checks settings each time records are pushed to ensure they are up-to-date with Algolia's dashboard. (Since `saveConfigurationToAlgolia` can be skipped at the beginning of the process).
- `IndicesConfigurator::saveConfigurationToAlgolia` now has a third optional parameter `$filteredEntities` which allows to only save configuration for given entities to avoid updating all settings indices each time. (to ensure backward compatibility with 3.18.0, if this parameter is not specified, legacy behavior still applies).
- As a result, product and category `IndexBuilder` only push settings for their related indices (same for `IndexMover` which is only related to products)
- `SaveSettings` Observer now also leverages this parameter:
   - saving config on products, instantsearch and images sections now only push product index settings
   - saving config on category section now only push category index settings
   - saving config on any other section still push settings on all indices
- Settings, Rules and Synonyms are now copied to temporary index with one single `operation` call 
- Added unit tests for `IndicesConfigurator`
- Updated integration test with the `$filteredEntities` param

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -310,6 +310,7 @@ class ProductHelper extends AbstractEntityHelper
      * @throws AlgoliaException
      * @throws LocalizedException
      * @throws NoSuchEntityException
+     * @throws \Throwable
      */
     public function setSettings(
         IndexOptionsInterface $indexOptions,
@@ -322,47 +323,13 @@ class ProductHelper extends AbstractEntityHelper
         if ($this->indexSettingsHandler->setSettings($indexOptions, $indexSettings)) {
             $this->logger->log('Index name: ' . $indexOptions->getIndexName());
             $this->logger->log('Settings: ' . json_encode($indexSettings));
-
             $this->algoliaConnector->waitLastTask($storeId);
+        }
 
-            if ($saveToTmpIndicesToo) {
-                // For temp index, we need to call AlgoliaConnector::setSettings() directly with explicit settings merge
-                // Main reasons:
-                //  - Temp indices never have replicas associated
-                //  - Settings rely directly on prod index so performing a diff with temporary online settings is useless
-                $this->algoliaConnector->setSettings(
-                    $indexTmpOptions,
-                    $indexSettings,
-                    false,
-                    true,
-                    $indexOptions->getIndexName()
-                );
-
-                $this->logger->log('Pushing the same settings to TMP index as well');
-                $this->algoliaConnector->waitLastTask($storeId);
-
-                try {
-                    $this->algoliaConnector->copySynonyms($indexOptions, $indexTmpOptions);
-                    $this->algoliaConnector->waitLastTask($storeId);
-                    $this->logger->log('
-                        Copying synonyms from production index to "' . $indexTmpOptions->getIndexName() . '" to not erase them with the index move.
-                    ');
-                } catch (AlgoliaException $e) {
-                    $this->logger->error('Error encountered while copying synonyms: ' . $e->getMessage());
-                }
-
-                try {
-                    $this->algoliaConnector->copyQueryRules($indexOptions, $indexTmpOptions);
-                    $this->algoliaConnector->waitLastTask($storeId);
-                    $this->logger->log('
-                        Copying query rules from production index to "' . $indexTmpOptions->getIndexName() . '" to not erase them with the index move.
-                    ');
-                } catch (AlgoliaException $e) {
-                    if ($e->getCode() !== 404) {
-                        throw $e;
-                    }
-                }
-            }
+        if ($saveToTmpIndicesToo && !$this->algoliaConnector->indexExists($indexTmpOptions->getIndexName(), $storeId)) {
+            $this->algoliaConnector->copyIndexConfig($indexOptions, $indexTmpOptions);
+            $this->logger->log('Copying the settings, synonyms and rules from production to "' . $indexTmpOptions->getIndexName() . '" index.');
+            $this->algoliaConnector->waitLastTask($storeId);
         }
 
         $this->setFacetsQueryRules($indexOptions);
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -326,7 +326,7 @@ class ProductHelper extends AbstractEntityHelper
             $this->algoliaConnector->waitLastTask($storeId);
         }
 
-        if ($saveToTmpIndicesToo && !$this->algoliaConnector->indexExists($indexTmpOptions->getIndexName(), $storeId)) {
+        if ($saveToTmpIndicesToo) {
             $this->algoliaConnector->copyIndexConfig($indexOptions, $indexTmpOptions);
             $this->logger->log('Copying the settings, synonyms and rules from production to "' . $indexTmpOptions->getIndexName() . '" index.');
             $this->algoliaConnector->waitLastTask($storeId);
--- a/Model/IndicesConfigurator.php
+++ b/Model/IndicesConfigurator.php
@@ -52,7 +52,11 @@ class IndicesConfigurator
      * @throws LocalizedException
      * @throws NoSuchEntityException
      */
-    public function saveConfigurationToAlgolia(int $storeId, bool $useTmpIndex = false): void
+    public function saveConfigurationToAlgolia(
+        int $storeId,
+        bool $useTmpIndex = false,
+        array $filteredEntities = [])
+    : void
     {
         $logEventName = 'Save configuration to Algolia for store: ' . $this->logger->getStoreName($storeId);
         $this->logger->start($logEventName, true);
@@ -70,14 +74,37 @@ class IndicesConfigurator
             return;
         }
 
+        if (count($filteredEntities) > 0) {
+            $this->logger->log('Filtered entities: ' . implode(',', $filteredEntities));
+
+            if (in_array('products', $filteredEntities)) {
+                $this->setProductsSettings($storeId, $useTmpIndex);
+            }
+            if (in_array('categories', $filteredEntities)) {
+                $this->setCategoriesSettings($storeId);
+            }
+        } else {
+            $this->setAllEntitiesSettings($storeId, $useTmpIndex);
+        }
+
+        $this->setExtraSettings($storeId, $useTmpIndex, $filteredEntities);
+
+        $this->logger->stop($logEventName, true);
+    }
+
+    /**
+     * @throws LocalizedException
+     * @throws NoSuchEntityException
+     * @throws DiagnosticsException
+     * @throws AlgoliaException
+     */
+    protected function setAllEntitiesSettings(int $storeId, bool $useTmpIndex = false): void
+    {
         $this->setCategoriesSettings($storeId);
         $this->setPagesSettings($storeId);
         $this->setQuerySuggestionsSettings($storeId);
         $this->setAdditionalSectionsSettings($storeId);
         $this->setProductsSettings($storeId, $useTmpIndex);
-        $this->setExtraSettings($storeId, $useTmpIndex);
-
-        $this->logger->stop($logEventName, true);
     }
 
     protected function logSettingsPush(IndexOptionsInterface $indexOptions, array $settings): void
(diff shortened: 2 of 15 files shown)


Service/Product/IndexBuilder.php
```diff
@@ -91,6 +95,11 @@ public function buildIndex(int $storeId, ?array $entityIds, ?array $options): vo
             return;
         }
 
+        // Before pushing the records, check settings and update them if needed (only when queue is active)
+        if ($this->queueHelper->isQueueActive($storeId)) {
+            $this->updateSettings($storeId);
```
cammonro: I'm having trouble understanding why we need to perform this update settings check. Shouldn't this already be handed by the call [here](https://github.com/algolia/algoliasearch-magento-2/blob/f5c7bd8a6c9d517226c295e58455a8c22c436308/Service/Product/BatchQueueProcessor.php#L65) which routes to `ProductHelper::setSettings` → `IndexSettingsHandler` → `IndexSettingsComparator` for non temp indexing? 

Do we have a use case? I feel this adds a lot of overhead - and even impacts delta indexing which should never need to perform a setting update. (`buildIndex(...)` is the shared entry point for *both* `buildIndexFull` and `buildIndexList`.)

Service/Product/BatchQueueProcessor.php
```diff
@@ -85,10 +88,18 @@ protected function getCollectionSize(int $storeId, Collection $collection): int
 
     protected function syncAlgoliaSettings(int $storeId, bool $useTmpIndex): void
     {
+        $indexOptions = $this->indexOptionsBuilder->buildEntityIndexOptions($storeId, $useTmpIndex);
+        $productSettings = $this->productHelper->getIndexSettings($storeId);
+
+        if (!$useTmpIndex && $this->indexSettingsComparator->matches($indexOptions, $productSettings)) {
+            return;
```
cammonro, line 99: This is a very useful addition - since it prevents creating a job we may never need. Do we need the `$useTmpIndex` exception case or can we skip in either event? Can we add a comment explaining why we need to check `$useTmpIndex`?

Service/Product/IndexBuilder.php
```diff
@@ -108,6 +117,18 @@ public function buildIndex(int $storeId, ?array $entityIds, ?array $options): vo
...
     }
 
+    /**
+     * @throws NoSuchEntityException
+     * @throws AlgoliaException
+     * @throws LocalizedException
+     */
+    protected function updateSettings(int $storeId): void
+    {
+        $indexOptions = $this->indexOptionsBuilder->buildEntityIndexOptions($storeId);
```
cammonro: This is going to target the prod index during temp-index runs - is that what we want?

Model/Observer/SaveSettings.php
```diff
@@ -8,10 +8,19 @@
 use Algolia\AlgoliaSearch\Model\IndicesConfigurator;
 use Magento\Framework\Event\Observer;
 use Magento\Framework\Event\ObserverInterface;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Exception\NoSuchEntityException;
 use Magento\Store\Model\StoreManagerInterface;
 
 class SaveSettings implements ObserverInterface
 {
+    public const EVENTS = [
```
cammonro, line 17: Looks like we're getting a jump on pre-filtering - I like this addition!

Model/IndicesConfigurator.php
```diff
@@ -70,14 +74,37 @@ public function saveConfigurationToAlgolia(int $storeId, bool $useTmpIndex = fal
...
         }
 
+        if (count($filteredEntities) > 0) {
+            $this->logger->log('Filtered entities: ' . implode(',', $filteredEntities));
+
+            if (in_array('products', $filteredEntities)) {
+                $this->setProductsSettings($storeId, $useTmpIndex);
+            }
+            if (in_array('categories', $filteredEntities)) {
+                $this->setCategoriesSettings($storeId);
```
cammonro, line 84: I like the filtering concept - but we've introduced an inconsistent API since we don't handle things like `pages` or `suggestions`- which would silently no-op. This is probably out of scope - but in for a penny in for a pound. Perhaps we can tag and scope it in another follow up ticket?

Service/Category/BatchQueueProcessor.php
```diff
@@ -82,8 +85,21 @@ protected function processSpecificCategories(array $categoryIds, int $categories
...
+
+        if (!$this->indexSettingsComparator->matches($indexOptions, $categoriesSettings)) {
+            /** @uses IndicesConfigurator::saveConfigurationToAlgolia() */
+            $this->queue->addToQueue(
+                IndicesConfigurator::class,
+                'saveConfigurationToAlgolia',
+                [
+                    'storeId' => $storeId,
+                    'useTmpIndex' => false,
+                    'filteredEntities' => ['categories']
```
cammonro, line 99: Since we're no longer pushing settings updates to other entities like `products` - I'm assuming there's no dependency here?

Model/IndicesConfigurator.php
```diff
@@ -203,9 +230,9 @@ protected function setProductsSettings(int $storeId, bool $useTmpIndex): void
     }
 
     /**
-     * @throws AlgoliaException|NoSuchEntityException
+     * @throws AlgoliaException|NoSuchEntityException|DiagnosticsException
      */
-    protected function setExtraSettings(int $storeId, bool $saveToTmpIndicesToo): void
+    protected function setExtraSettings(int $storeId, bool $saveToTmpIndicesToo, ?array $filteredEntities): void
```
cammonro: We might need a default null value here for PHP 8?

Ui/Component/Listing/Column/Data.php
```diff
@@ -44,7 +44,7 @@ protected function formatData(array $data, $depth = 0): string
             $stringKey = '- <strong>' . $key . '</strong>';
 
             if (is_array($value)) {
-                if ($key === 'entityIds') {
+                if ($key === 'entityIds' || $key === 'filteredEntities') {
```
cammonro, line 47: Nice way to provide visibility on the new filtering. 👍

Helper/Entity/ProductHelper.php
```diff
@@ -322,47 +323,13 @@ public function setSettings(
...
-                    $this->logger->log('
-                        Copying query rules from production index to "' . $indexTmpOptions->getIndexName() . '" to not erase them with the index move.
-                    ');
-                } catch (AlgoliaException $e) {
-                    if ($e->getCode() !== 404) {
-                        throw $e;
-                    }
-                }
-            }
+        if ($saveToTmpIndicesToo && !$this->algoliaConnector->indexExists($indexTmpOptions->getIndexName(), $storeId)) {
```
cammonro: It's good we're checking this but if an aborted prior reindex (timeout, crash, hand-stopped worker) leaves tmp around with stale settings/synonyms/rules - it won't get updated. Maybe we should call a `deleteIndex` op here to ensure a clean slate?

Helper/Entity/ProductHelper.php
```diff
@@ -322,47 +323,13 @@ public function setSettings(
...
+            $this->algoliaConnector->copyIndexConfig($indexOptions, $indexTmpOptions);
```
cammonro, line 330: Nice refactor - eliminates several wait tasks!

cammonro (changes requested):
@damcou There are some nice improvements here but I do have reservations about the direction we're going with the added `setSettings` calls. (Please see comments)

Really like the new entity filtering mechanism and job creation guard and how we've eliminated a lot of wait tasks. 🤘

Model/IndicesConfigurator.php
damcou: You're right, I've added `pages`, `suggestions` and `additional_sections` checks to be complete.

Service/Category/BatchQueueProcessor.php
damcou: That's a good question, for me, updating categories settings could potentially affect products indexing but not really products settings, do you see a usecase where we should push `products` settings as well ?

Model/IndicesConfigurator.php
damcou: Good catch !

Helper/Entity/ProductHelper.php
damcou: I have trouble to see where such delete operation would be safely added , I think there are many cases to consider and I wouldn't like to delete a tmp index when we shouldn't... I think this is worth a ticket by itself if you agree ? In the meantime, it doesn't hurt too much to remove the `indexExists` check.

Service/Product/IndexBuilder.php
damcou: anwser in JIRA

Service/Product/IndexBuilder.php
damcou: Removed

Test/Unit/Model/IndicesConfiguratorTest.php
```diff
@@ -0,0 +1,234 @@
...
+        $this->allowPassingGuards();
+
+        $this->configurator->expects($this->once())->method('setProductsSettings');
+        $this->configurator->expects($this->once())->method('setCategoriesSettings');
+        $this->configurator->expects($this->never())->method('setAllEntitiesSettings');
+
+        $this->configurator->saveConfigurationToAlgolia($this->storeId, false, ['products', 'categories']);
+    }
+
+    public function testSkipsUnrecognizedEntitiesInFilterBranch(): void
```
cammonro, line 205: I think this test should probably be renamed now that we include all filters?

Helper/Entity/ProductHelper.php
```diff
@@ -322,47 +323,13 @@ public function setSettings(
...
-                        Copying query rules from production index to "' . $indexTmpOptions->getIndexName() . '" to not erase them with the index move.
-                    ');
-                } catch (AlgoliaException $e) {
-                    if ($e->getCode() !== 404) {
-                        throw $e;
-                    }
-                }
-            }
+        if ($saveToTmpIndicesToo) {
+            $this->algoliaConnector->copyIndexConfig($indexOptions, $indexTmpOptions);
```
cammonro, line 330: Way simpler! ❤️  Très magnifique!

Service/Product/BatchQueueProcessor.php
```diff
@@ -85,10 +88,22 @@ protected function getCollectionSize(int $storeId, Collection $collection): int
 
     protected function syncAlgoliaSettings(int $storeId, bool $useTmpIndex): void
     {
+        $indexOptions = $this->indexOptionsBuilder->buildEntityIndexOptions($storeId, $useTmpIndex);
+        $productSettings = $this->productHelper->getIndexSettings($storeId);
+
+        // $useTmpIndex needs to be checked here because we need to ensure proper creation of the tmp index at this point
+        // (tmp index needs to be initialized in the setSettings operation even if the settings are matching ALgolia dashboard)
```
cammonro, line 95: I can call you Betty and Betty when you call me you can call me AL-golia (Paul Simon reference 😛 )

Helper/Entity/ProductHelper.php
cammonro: You know - I think the way you've now implemented is fine - because index **copy** operations do not trigger no-op de-prioritization like `setSettings` (based on my understanding). So I think it's worth a ticket to investigate the merits of performing a preemptive delete for clean state in general - but from the perspective of avoid no-ops this may now be a non-issue.

cammonro (approved):
This looks great. Nice job. 👏  A few small non-blocking nits I noted but otherwise ... Approved!

Service/Product/BatchQueueProcessor.php
damcou: I don't have the ref :D (keeping it as it is since it seems to be a nice joke ! )

PR #1942, opened 2026-05-19
MAGE-1540 Hoist unit tests to AbstractStoreCommand scope

cammonro:
### Summary

Adds unit test class `Test/Unit/Console/Command/AbstractStoreCommandTest` for `Console/Command/AbstractStoreCommand`, in preparation for the ingestion-package CLI commands that extend it on PR https://github.com/algolia/algoliasearch-ingestion-magento-2/pull/11.

Hoisting it here keeps the parent's contract regression-tested in the package that ships it, and prevents the ingestion package from depending on internal behavior of an upstream class without coverage.

> [!NOTE]
> This PR also merges the latest GA release of 3.18.0 (same as what was done for [3.18.1](https://github.com/algolia/algoliasearch-magento-2/pull/1945))

### Results

image


Test/Unit/Console/Command/AbstractStoreCommandTest.php
```diff
@@ -0,0 +1,79 @@
...
+use Algolia\AlgoliaSearch\Console\Command\AbstractStoreCommand;
+use Algolia\AlgoliaSearch\Service\StoreNameFetcher;
+use Algolia\AlgoliaSearch\Test\TestCase;
+use Magento\Framework\App\Area;
+use Magento\Framework\App\State;
+use Magento\Framework\Exception\LocalizedException;
+use PHPUnit\Framework\MockObject\MockObject;
+use Symfony\Component\Console\Output\BufferedOutput;
+
+class AbstractStoreCommandTest extends TestCase
```
damcou, line 14: There's an existing `ClearQueueCommandTest` class, should it extend this one ?

damcou (approved):
Reviewed AbstractStoreCommandTest class only, I had a small question but LGTM ✅ 🚀

Test/Unit/Console/Command/AbstractStoreCommandTest.php
cammonro: We certainly could but I was trying not to slip on scope too much. I created a separate Jira task to unify our console unit tests under Symfony `CommandTester` - used in the new ingestion module ([example](https://github.com/algolia/algoliasearch-ingestion-magento-2/blob/90e518e07a5e4f6e7dfaab346663f8137fa55370/Test/Unit/Console/Command/Ingestion/IngestionInitCommandTest.php#L47)) - which I think is a better pattern and reuses an already well defined mechanism.

PR #1946, opened 2026-05-28
MAGE-1573: Collect task ids and wait once during save settings (also MAGE-1574 and MAGE-1575)

damcou:
This PR contains:
- Removed all `waitLastTask` occurrences in `saveConfigurationToAlgolia`
- Instead we now collect all task ids and wait for each of them at the end of the process
- As a result, AlgoliaConnector has been improved with a `collectTaskIdToWaitFor` method which adds the last task id of a given index into an array and a `waitForAllCollectedTaskIds` method which iterates over this array and wait for all task ids listed this way
- Removed unnecessary `getSettings` occurrences in the additional sections config loo. 
- Updated related unit tests
- Updated related integration tests

**Note**: This PR covers MAGE-1573, MAGE-1574 and MAGE-1575

### DIFF
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -329,7 +329,7 @@ class ProductHelper extends AbstractEntityHelper
         if ($saveToTmpIndicesToo) {
             $this->algoliaConnector->copyIndexConfig($indexOptions, $indexTmpOptions);
             $this->logger->log('Copying the settings, synonyms and rules from production to "' . $indexTmpOptions->getIndexName() . '" index.');
-            $this->algoliaConnector->collectTaskIdToWaitFor($indexTmpOptions);
+            $this->algoliaConnector->collectTaskIdToWaitFor($indexOptions);
         }
 
         $this->setFacetsQueryRules($indexOptions);
@@ -456,12 +456,12 @@ class ProductHelper extends AbstractEntityHelper
 
     /**
      * @param IndexOptionsInterface $indexOptions
-     * @param bool $waitLastTask
+     * @param bool $collectTaskId
      * @return void
      * @throws AlgoliaException
      * @throws NoSuchEntityException
      */
-    protected function setFacetsQueryRules(IndexOptionsInterface $indexOptions, bool $waitLastTask = false): void
+    protected function setFacetsQueryRules(IndexOptionsInterface $indexOptions, bool $collectTaskId = false): void
     {
         $this->clearFacetsQueryRules($indexOptions);
 
@@ -500,7 +500,7 @@ class ProductHelper extends AbstractEntityHelper
 
             $this->algoliaConnector->saveRules($indexOptions, $rules, true);
 
-            if ($waitLastTask) {
+            if ($collectTaskId) {
                 $this->algoliaConnector->collectTaskIdToWaitFor($indexOptions);
             }
         }
--- a/Helper/Entity/ProductHelper.php
+++ b/Helper/Entity/ProductHelper.php
@@ -323,13 +323,13 @@ class ProductHelper extends AbstractEntityHelper
         if ($this->indexSettingsHandler->setSettings($indexOptions, $indexSettings)) {
             $this->logger->log('Index name: ' . $indexOptions->getIndexName());
             $this->logger->log('Settings: ' . json_encode($indexSettings));
-            $this->algoliaConnector->waitLastTask($storeId);
+            $this->algoliaConnector->collectTaskIdToWaitFor($indexOptions);
         }
 
         if ($saveToTmpIndicesToo) {
             $this->algoliaConnector->copyIndexConfig($indexOptions, $indexTmpOptions);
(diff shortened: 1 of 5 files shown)


Helper/Entity/ProductHelper.php
```diff
@@ -323,13 +323,13 @@ public function setSettings(
...
             $this->logger->log('Settings: ' . json_encode($indexSettings));
-            $this->algoliaConnector->waitLastTask($storeId);
+            $this->algoliaConnector->collectTaskIdToWaitFor($indexOptions);
         }
 
         if ($saveToTmpIndicesToo) {
             $this->algoliaConnector->copyIndexConfig($indexOptions, $indexTmpOptions);
             $this->logger->log('Copying the settings, synonyms and rules from production to "' . $indexTmpOptions->getIndexName() . '" index.');
-            $this->algoliaConnector->waitLastTask($storeId);
+            $this->algoliaConnector->collectTaskIdToWaitFor($indexTmpOptions);
```
cammonro: Are we collecting the right wait task? We're using the `fromIndex` [here](https://github.com/algolia/algoliasearch-magento-2/blob/05257f2cee51b4b1e6983b0663e281b5cb77219e/Service/AlgoliaConnector.php#L592).

Service/AlgoliaConnector.php
```diff
@@ -699,6 +701,26 @@ public function waitLastTask(?int $storeId = null, ?string $lastUsedIndexName =
         $this->getClient($storeId)->waitForTask($lastUsedIndexName, $lastTaskId);
     }
 
+    public function collectTaskIdToWaitFor(IndexOptionsInterface $indexOptions): void
```
cammonro, line 704: I like the idea of adding this harvester to the singleton but I'm wondering if we need to scope by store ID in case of operations across multi app?

Helper/Entity/ProductHelper.php
```diff
@@ -501,7 +501,7 @@ protected function setFacetsQueryRules(IndexOptionsInterface $indexOptions, bool
             $this->algoliaConnector->saveRules($indexOptions, $rules, true);
 
             if ($waitLastTask) {
-                $this->algoliaConnector->waitLastTask($indexOptions->getStoreId());
+                $this->algoliaConnector->collectTaskIdToWaitFor($indexOptions);
```
cammonro: Since we're not really waiting anymore (just collecting) the `$waitLastTask` boolean meaning has changed a bit. Maybe we can just remove it and let the caller manage its own wait?

cammonro (changes requested):
Looking good so far! I like the new harvester. I had a few question/concerns if you wouldn't mind taking a look.

Helper/Entity/ProductHelper.php
damcou: good catch 👍

Service/AlgoliaConnector.php
damcou: Added a store scope; good idea !

Helper/Entity/ProductHelper.php
damcou: I changed the naming to` $collectTaskId` to be more meaningful, for some reason this is still needed for the integration test.

cammonro: Since we're only using it for our integration testing can't we just call the wait from the test rather than make it part of the API?

cammonro (commented):
Looking good! But I'm thinking we should maybe just change our integration tests. What do you think?

Helper/Entity/ProductHelper.php
damcou: Done ✅

Test/Integration/Indexing/Config/MultiStoreConfigTest.php
cammonro: Since the second params are now ignored - can we clean these up for clarity?

Test/Integration/Indexing/Config/MultiStoreConfigTest.php
```diff
@@ -102,6 +102,7 @@ public function testMultiStoreIndicesCreation()
         $this->algoliaConnector->waitForAllCollectedTaskIds($defaultIndexOptions->getStoreId());
 
         $this->invokeMethod($productHelper, 'setFacetsQueryRules', [$fixtureIndexOptions, true]);
+        $this->algoliaConnector->collectTaskIdToWaitFor($fixtureIndexOptions);
```
cammonro, line 105: Previously we were collecting for both fixtures - is this no longer needed?

Test/Integration/Indexing/Config/MultiStoreConfigTest.php
damcou: oopsy, I forgot, sorry

Test/Integration/Indexing/Config/MultiStoreConfigTest.php
damcou: No because the purpose of the test is precisely to test that only the second store has Facet Query Rules

cammonro: Got it! Thanks for explaining.

cammonro (approved):
Let's go! 🚀

PR #1947, opened 2026-06-02
MAGE-1576: SaveSettings store filtering when indexing is disabled

damcou:
This PR contains:
- `SaveSettings` Observer inside store filtering when indexing is disabled 
- Added possibility to add a configurable separator in the Algolia logging so it's (hopefully) easier to read
- Added related unit tests class

### DIFF
--- a/Logger/DiagnosticsLogger.php
+++ b/Logger/DiagnosticsLogger.php
@@ -17,6 +17,7 @@ class DiagnosticsLogger
     /** @var array */
     protected const ALGOLIA_TAGS = ['group' => 'algolia'];
     protected const PROFILE_LOG_MESSAGES_DEFAULT = false;
+    protected const SEPARATOR_DEFAULT = false;
     protected const DEFAULT_NAMESPACE_DEPTH = 2;
 
     protected bool $isLoggerEnabled = false;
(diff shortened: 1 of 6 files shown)


Logger/DiagnosticsLogger.php
```diff
@@ -17,6 +17,7 @@ class DiagnosticsLogger
     /** @var array */
     protected const ALGOLIA_TAGS = ['group' => 'algolia'];
     protected const PROFILE_LOG_MESSAGES_DEFAULT = false;
+    protected const SEPARATOR_DEFAULT = false;
```
cammonro, line 20: Since this is related to `TimedLogger` it seems like the default constant should belong there? 

What if we made this an integer instead that could give you the option adjust the spacing? (with default = 0)

Logger/TimedLogger.php
```diff
@@ -16,24 +16,32 @@ public function __construct(
     )
     {}
 
-    public function start($action): void
+    public function start(string $action, bool $separator): void
```
cammonro, line 19: It would be nice if we could apply the defaults here as well.

Logger/TimedLogger.php
```diff
@@ -16,24 +16,32 @@ public function __construct(
...
+    public function stop(string $action, bool $separator): void
     {
         if (false === isset($this->timers[$action])) {
             throw new DiagnosticsException(__('Algolia Logger => non existing action'));
         }
 
         $this->log('<<<<< END ' . $action . ' (' . $this->formatTime($this->timers[$action], microtime(true)) . ')');
+
+        if ($separator) {
+            $this->log('');
```
cammonro, line 42: This will still prefix and add `[] []` - which may be totally fine - but if we used something like `Monolog\Formatter\LineFormatter` I believe we could have completely empty lines - not sure if that would be useful.

Model/Observer/SaveSettings.php
```diff
@@ -42,6 +42,10 @@ public function execute(Observer $observer)
          try {
             $storeIds = array_keys($this->storeManager->getStores());
             foreach ($storeIds as $storeId) {
+                if (!$this->helper->isIndexingEnabled($storeId)) {
```
cammonro, line 45: Yay!

cammonro (approved):
✅  Approved with some non blocking suggestions.

Like the added unit tests!

PR #1949, opened 2026-06-04
MAGE-1581: SetSettings optimizations for 3.17.4-dev

damcou:
This PR contains porting of all changes related to `SetSettings` optimizations from 3.18.1 including:
- https://github.com/algolia/algoliasearch-magento-2/pull/1933
- https://github.com/algolia/algoliasearch-magento-2/pull/1938
- https://github.com/algolia/algoliasearch-magento-2/pull/1941
- https://github.com/algolia/algoliasearch-magento-2/pull/1946
- https://github.com/algolia/algoliasearch-magento-2/pull/1947


Model/IndicesConfigurator.php
```diff
@@ -164,26 +206,29 @@ protected function setQuerySuggestionsSettings(int $storeId): void
      */
     protected function setAdditionalSectionsSettings(int $storeId): void
     {
-        $logEventName = 'Pushing settings for additional section indices.';
+        $logEventName = 'Pushing settings for query suggestions indices.';
         $this->logger->start($logEventName, true);
 
         $protectedSections = ['products', 'categories', 'pages', 'suggestions'];
-        foreach ($this->configHelper->getAutocompleteSections() as $section) {
+        $configSections = $this->configHelper->getAutocompleteSections();
```
cammonro: Since [we introduced store scoping on this config](https://github.com/algolia/algoliasearch-magento-2/blob/fea99e9c46e806ecc3d5131128821f4423e7a23a/Model/IndicesConfigurator.php#L205) in `3.18.1` - it seems we should include it here with the patch also no?

cammonro (approved):
Looks good to me @damcou ! Just one part of the patch that did not backport (because of the shift from the deprecated wrapper it looks like). Should just be a one line fix without committing to the dedicated config helper.

Model/IndicesConfigurator.php
damcou: good catch ! Indeed, I forgot this one ! 🙄

PR #1950, opened 2026-06-08
MAGE-1606: update version files (3.17.4) 

damcou:
Prepping for 3.17.4


README.md
cammonro: Since `3.17.4` is released after `3.18.x` - we should update the EOL matrix especially for [`3.16.x`](https://github.com/algolia/algoliasearch-magento-2/blob/a2c5c7f1a2638d441c28b4576643cd3b3d29a408/README.md?plain=1#L59)

cammonro (changes requested):

cammonro (approved):
Codacy alerts persist but these appear to be related to diff failures on their end - not our code. LGTM!

PR #1956, opened 2026-06-12
MAGE-1618 (also MAGE-1621) Reorganize Service/Index* class family (no behavior change)

cammonro:
## Summary

This PR is a pure structural refactor. 

The settings related service family has been growing (`IndexSettingsHandler`, `IndexSettingsComparator`, and soon `IndexSettingsPreserver`), as has the broader index related family (`IndexOptionsBuilder`, `IndexNameFetcher`). Grouping them under `Service/Index/` and `Service/Index/Settings/` is a low cost cleanup that fits a minor version and gives the upcoming index setting related feature a natural home.

The `IndexSettings*` and `Index*` class name prefixes are kept intentionally, so the names stay self documenting in IDE tabs, stack traces, PHPStan output, and DI error messages without needing directory context.

## What changed

Four classes moved to new namespaces (tracked by git as renames, so history is preserved):

| Old FQCN | New FQCN |
| --- | --- |
| `Algolia\AlgoliaSearch\Service\IndexOptionsBuilder` | `Algolia\AlgoliaSearch\Service\Index\IndexOptionsBuilder` |
| `Algolia\AlgoliaSearch\Service\IndexNameFetcher` | `Algolia\AlgoliaSearch\Service\Index\IndexNameFetcher` |
| `Algolia\AlgoliaSearch\Service\IndexSettingsHandler` | `Algolia\AlgoliaSearch\Service\Index\Settings\IndexSettingsHandler` |
| `Algolia\AlgoliaSearch\Service\IndexSettingsComparator` | `Algolia\AlgoliaSearch\Service\Index\Settings\IndexSettingsComparator` |

`Service/AlgoliaConnector.php` stays where it is. It is the Algolia client wrapper, not index orchestration, so the reorg does not apply to it.

### Commits

1. **Reorganize index related service classes.** Moves the four classes, updates their namespaces, and repoints every in package consumer (and the unit and integration tests) at the new FQCNs. `Service/AlgoliaConnector` previously referenced `IndexNameFetcher` and `IndexOptionsBuilder` by bare same namespace name, so it gains explicit `use` imports for the new locations. No `etc/di.xml` change is needed: these classes are wired by autowiring, and the only DI declaration in that area targets the non moving `Service\Product\IndexOptionsBuilder\Proxy`.
2. **Add deprecation stubs.** Each old path now holds an empty subclass of the new class, marked `@deprecated since 3.19` and slated for removal in 3.20.

Why empty subclasses rather than `class_alias()`: a real subclass keeps DI `<preference>` and `<type>` declarations, plugins, and customers' own `extends` chains resolving against the old FQCN as a genuine, Magento compatible class.

## Backward compatibility

The stubs make every old FQCN continue to resolve. Verified at runtime that each old class loads and `is_subclass_of` its new counterpart:

```
OK  Service\IndexOptionsBuilder      -> Service\Index\IndexOptionsBuilder
OK  Service\IndexNameFetcher         -> Service\Index\IndexNameFetcher
OK  Service\IndexSettingsHandler     -> Service\Index\Settings\IndexSettingsHandler
OK  Service\IndexSettingsComparator  -> Service\Index\Settings\IndexSettingsComparator
```

### Cross package audit

Grepped the four sibling repos for references to the moving classes. All usages are plain references with no `<preference>` or `<type>` override against the old FQCN, so the stubs will cover them for now:

- `algoliasearch-adapter-magento-2`: `IndexOptionsBuilder` (base and `Product`), `IndexNameFetcher`
- `algoliasearch-inventory-magento-2`: `IndexSettingsHandler`, `IndexOptionsBuilder`, `IndexNameFetcher`
- `algoliasearch-custom-algolia-magento-2`, `algoliasearch-client-php`: no references

However, we'll need to upgrade the first two packages prior to release 3.20.

The in package entity subclasses (`Service\Product\IndexOptionsBuilder` and the Category, Page, Suggestion, AdditionalSection variants) now extend the new base directly.

## Addendum: retire the `*Testable` unit-test pattern (MAGE-1621)

While here, this PR also removes the `*Testable` subclass pattern from unit tests. These throwaway subclasses re-declared protected methods as public just to test them, which duplicated a capability the base `Algolia\AlgoliaSearch\Test\TestCase` already provides via `invokeMethod()`.

The trigger was `IndexSettingsHandlerTestable` (the moved handler's test stub). Rather than relocate it with the reorg, it was deleted, and the remaining occurrences were cleaned up at the same time so the pattern does not linger.

**Removed (4 files), with their consuming tests migrated to `invokeMethod()`:**
- `Test/Unit/Service/Index/Settings/IndexSettingsHandlerTestable.php` (`IndexSettingsHandlerTest`)
- `Test/Unit/Controller/Adminhtml/IndexingManager/ReindexTestable.php` (`ReindexTest`)
- `Test/Unit/Service/Insights/EventProcessorTestable.php` (`EventProcessorTest`)
- `Test/Unit/Service/Product/FacetBuilderTestable.php` (`FacetBuilderTest`)

## Testing

- Full unit suite passes: `OK (343 tests, 814 assertions)`.
- `php -l` clean on every changed and added file.
- `magento2-lint` (PHP-CS-Fixer) clean on all moved classes and stubs.
- `composer validate` passes (only the pre existing `version` field notice).
image

## Follow ups

- A follow up PR will introduce `Service\Index\Settings\IndexSettingsPreserver` and the `attributesForFaceting` preservation behavior.
- **3.20 cleanup:** remove the four deprecation stub files once dependents have migrated.


damcou (approved):
LGTM ! (More information on JIRA)

PR #1958, opened 2026-06-15
Release/3.18.1 -> main

damcou:
### Updates
- Ensured compatibility of the extension with Magento 2.4.9 and PHP 8.5
- Added improvements on `setSettings` operations sent during indexing and configuration saving:
  - Introduced `IndexSettingsComparator` service to compare settings with the Algolia Dashboard.
  - Added diff check for both configuration save and batch processing.
  - Added collect of all task IDs during `SaveSettings` and wait only once for all operations to complete at the end.
  - Added store filtering during `SaveSettings`.
  - Updated Unit and Integration tests to reflect those changes.


cammonro (approved):
🚀

PR #1920, opened 2026-03-20
Copy settings and synonyms on pages full reindex

PromInc:
**Summary**

When doing a full reindex on Pages, the settings and synonyms configured in Algolia do not copy over to the new index - thus they are lost.

**Result**

Settings and Synonyms should not be wiped of the Algolia `_pages` index

**Steps to Recreate**
1. Set custom settings (ie. custom ranking) in Algolia for a `_pages` index
2. Create synonyms in Algolia for a `_pages` index
3. 4. Trigger a reindex from Magento `php bin/magento indexer:reindex algolia_pages`
5. Wait till complete
6. Observe in Algolia that settings and synonyms have been deleted

PR #1960, opened 2026-06-23
MAGE-1625: Unified temporary index management for entities

damcou:
This PR contains:
- Unified way to manage temporary index management for suggestions, pages and additional sections
- Fixed pages indexing issue raised in #1920 (thanks @PromInc )
- Fixed suggestions indexing issue said in #1930 (thanks @its-leoeff511)


Service/Page/IndexBuilder.php
```diff
@@ -99,16 +102,5 @@ public function buildIndex($storeId, ?array $entityIds, ?array $options): void
...
-
-        if ($isFullReindex) {
-            $tempIndexOptions = $this->indexOptionsBuilder->buildEntityIndexOptions($storeId, true);
-
-            $this->algoliaConnector->copyQueryRules($indexOptions, $tempIndexOptions);
-            $this->algoliaConnector->moveIndex($tempIndexOptions, $indexOptions);
-        }
-        $this->algoliaConnector->setSettings(
-            $indexOptions,
-            $this->pageHelper->getIndexSettings($storeId)
```
cammonro, line 111: We seemed to have removed Magento updating the settings at time of full reindex. Right now it seems that only what is on prod will be preserved. Is this intentional?

Service/AdditionalSection/IndexBuilder.php
```diff
@@ -78,13 +78,7 @@ public function buildIndex(int $storeId, ?array $entityIds, ?array $options): vo
                 $this->saveObjects($chunk, $tempIndexOptions);
             }
 
-            $this->algoliaConnector->copyQueryRules($indexOptions, $tempIndexOptions);
-            $this->algoliaConnector->moveIndex($tempIndexOptions, $indexOptions);
-
-            $this->algoliaConnector->setSettings(
```
cammonro, line 84: Same question as pages settings updated via Magento.

Test/Integration/Indexing/Page/PagesIndexingTest.php
```diff
@@ -88,7 +88,6 @@ public function testDefaultIndexableAttributes()
             'content',
             'algoliaLastUpdateAtCET',
             '_highlightResult',
-            '_snippetResult',
```
cammonro, line 91: Does this have to do with the dropped pages setting update?

Service/Page/IndexBuilder.php
```diff
@@ -85,6 +85,9 @@ public function buildIndex($storeId, ?array $entityIds, ?array $options): void
                     continue;
                 }
             }
+            if ($isFullReindex) {
+                $this->moveTemporaryIndex($indexOptions, $toIndexOptions);
```
cammonro: Do we want to nest this under `if (isset($pages['toIndex']) && count($pages['toIndex'])) {`? If all pages are deleted from Magento then the index will not update. It's an edge case but something to consider.

cammonro (commented):
The refactor looks solid but we seem to have changed the settings behavior a little?

Service/Page/IndexBuilder.php
damcou: You're right, after thinking about it a bit, I think we should continue to push our settings during the process, I've decided to add `saveConfigurationToAlgolia` as we do for categories and products which leverages the `IndexSettingsComparator` service

Service/AdditionalSection/IndexBuilder.php
damcou: Since additional sections is a specific use case which works with enforced index names and we can't just simply add `saveConfigurationToAlgolia` for each section, I decided to restore the `setSettings` here.

Test/Integration/Indexing/Page/PagesIndexingTest.php
damcou: Since we now push the Magento settings, I could re-add it.

Service/Page/IndexBuilder.php
damcou: Updated !

Service/AdditionalSection/IndexBuilder.php
cammonro: Won't this risk a no-op? Perhaps we should leverage `IndexSettingsHandler` here?

cammonro (changes requested):
This looks good to me except the potential no op on additional sections settings. I don't believe we're applying the comparator there?

Service/AdditionalSection/IndexBuilder.php
damcou: Updated !

cammonro (approved):
Approved! Let's go 🚀

PR #1961, opened 2026-06-24
MAGE-1618 Preserve ingestion-owned index settings on reindex

cammonro:
**Summary**

A problem was identified where the Magento integration clobbers `attributesForFaceting` entries that are managed _outside_ of the extension (notably the internal underscore-prefixed entries written by the Algolia ingestion pipeline for Collections) every time it pushes index settings during a reindex or from the admin. This PR makes those remotely managed entries survive a Magento integration settings write.

This is PR 2 - stacked on the `Service/Index/` reorganization that already merged via #1956.

Noteworthy changes:

- **Adds `IndexSettingsPreserver`** - a pure transformation keyed by a `settingKey => PCRE pattern` rule map. For each configured key present in the payload Magento is about to push, remote entries matching the pattern are merged back in (remote wins). The default rule preserves leading-underscore entries on `attributesForFaceting`, both bare (`_foo`) and inside a `FacetBuilder` decorator (`searchable(_foo)`, `filterOnly(_foo)`). A locally proposed entry that collides with a protected pattern is dropped and logged as a warning rather than overwriting the remote value.
- **Restructures the settings write path** - `IndexSettingsHandler::setSettings` now fetches the remote settings once, runs preservation, then threads that same payload into the comparator. `IndexSettingsComparator::matches` gained an optional pre-fetched `$algoliaSettings` argument (falls back to fetching when null), so the remote settings are read exactly once per write instead of twice. Preservation runs before the no-op comparison so that diff detection sees the **final payload** and a write whose only difference was a preserved entry is correctly skipped. (This prevents a regression of the no-op de-prioritization that was addressed in 3.18.1 and 3.17.4)
- **Added unit tests**

**Result**

Full unit test suite green:

image

Full reindex with ingestion enabled:
image
image

Collections remain stable:
image
Data in tact:
image
image

Front end renders the Collection:

image

### DIFF
--- a/Service/Index/Settings/IndexSettingsPreserver.php
+++ b/Service/Index/Settings/IndexSettingsPreserver.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Service\Index\Settings;
+
+use Algolia\AlgoliaSearch\Logger\AlgoliaLogger;
+
+/**
+ * Merges remote setting entries that are managed outside of Magento (for example the internal
+ * `attributesForFaceting` entries written by the Algolia ingestion pipeline) back into the payload
+ * Magento is about to push, so they are not clobbered on every reindex.
+ *
+ * Configured by a map of `settingKey => PCRE pattern`. For each configured key present in the proposed
+ * payload, remote entries matching the pattern are preserved (remote wins).
+ */
+class IndexSettingsPreserver
+{
+    /**
+     * Default preservation rules: setting key => PCRE pattern applied to each entry.
+     *
+     * The `attributesForFaceting` pattern matches a leading underscore, both bare (`_foo`) and inside a
+     * `FacetBuilder` decorator (`searchable(_foo)`, `filterOnly(_foo)`).
+     *
+     * @var array<string, string>
+     */
+    public const DEFAULT_RULES = [
+        'attributesForFaceting' => '/(?:^|\()_/',
+    ];
+
+    /**
+     * @param array<string, string> $rules setting key => PCRE pattern
+     */
+    public function __construct(
+        protected AlgoliaLogger $logger,
+        protected array         $rules = self::DEFAULT_RULES,
+    ) {}
+
+    /**
+     * Merge remote entries matching the configured preservation rules into the proposed payload.
+     *
+     * Pure transformation: both the proposed payload and the remote payload are supplied by the caller,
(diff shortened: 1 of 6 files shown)


Service/Index/Settings/IndexSettingsPreserver.php
```diff
@@ -0,0 +1,86 @@
...
+            $proposedValues = is_array($proposed[$key]) ? $proposed[$key] : [];
+            $remoteValues = (isset($remote[$key]) && is_array($remote[$key])) ? $remote[$key] : [];
+
+            $preserved = array_filter(
+                $remoteValues,
+                fn($entry): bool => is_string($entry) && preg_match($pattern, $entry) === 1
+            );
+
+            // Remote wins: a protected entry should never originate from Magento, so drop any that does
+            // (and surface it) before merging the authoritative remote entries back in.
```
damcou, line 60: Do we agree that this is only for future proof purposes ? AFAIK for now, since we only use this for `_collections`, the idea is to get this value from Algolia and add it to the payload so it doesn't get lost in the process, right ? So for now there is no cases where we push values which we don't want to update on the dashboard ?

When we say "remote wins", for me it means that if we push "A" for a protected entry and remote says "B", it will be "B", but in our case, we just want the payload to include "B" because we don't push anything for this value.

damcou (approved):
Test are passing ✅ 
I just have a question on the overall usage of this new class, but nothing blocking (except maybe some Codacy complaints ?)  great job 🎉

Service/Index/Settings/IndexSettingsPreserver.php
cammonro: @damcou You are exactly right. The intention of this class is purely to _preserve_ what we explicitly name. The rules define what gets preserved and provides a way to broaden the preservation logic if it is needed. 

As of right now we are only targeting `attributesForFaceting` since this is a known use case. Direction on that front is that anything that starts with `_` is reserved by Algolia. In that instance, Algolia is the source of truth (not Magento) so "remote wins".

In reality, it is not possible to create an EAV through regular means with a field name that starts with an `_` but this is a guard "just in case" (perhaps some some other customization is writing fields like this). In that case it would not be able to overwrite Algolia's internally managed attribute.

cammonro:
@damcou Thank you for the review! Codacy complaints were false positives which I resolved. Callbacks are _not_ a vulnerability when they are hardcoded inline arrow functions or closures.

PR #1967, opened 2026-06-30
MAGE-1543 Add reference to new ingestion module

cammonro:
**Summary**

Adds a new subsection under "Optional Modules" for the [Ingestion Module](https://github.com/algolia/algoliasearch-ingestion-magento-2).


damcou (approved):
Let's go 🚀

PR #1971, opened 2026-07-09
MAGE-1648 Strip semanticSearch from mergeSettings (3.18.2)

cammonro:
## Summary

During a full reindex, `saveConfigurationToAlgolia` fails with a server-side `BadRequestException`:

```
Request body parsing: json: cannot unmarshal array into Go struct field
Alias.semanticSearch of type neuralengine.SemanticSearchSettings
```

This extension never sets `semanticSearch` itself. The value is echoed back from the live production index through the products temp-index merge. `semanticSearch` however is not played back on the standard `getSettings` path.  An empty `semanticSearch` object decodes in PHP to an empty array `[]`, and `json_encode([])` emits a JSON array, which the API rejects because `SemanticSearchSettings` must be an object.

Preconditions for this scenario: the live prod index has `semanticSearch` configured, and the merchant has non-empty products Extra Settings JSON (opening the `if ($extraSettings)` branch that triggers the merge).

This PR adds `'semanticSearch'` to the `$removals` array in`AlgoliaConnector::getSettingsToRemove()` so it is stripped from online settings before the merge round-trip.

This is a targeted patch but the merge machinery this bug lives in is to be removed entirely in a future refactor. This fix is intentionally throwaway but targeted unit tests are supplied for this version. 

The comprehensive `AlgoliaConnectorTest` suite lands with 3.19.

## Result

Unit tests:
image

NeuralSearch settings preservation confirmed via manual testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Bug Fixes
 - Fixed pages indexing issue where settings and synonyms added on the Algolia dashboard were wiped during the process - thank you @PromInc
 - Fixed suggestions indexing issue where the index could be completely emptied during the process - thank you @its-leoeff511
+- Prevented semanticSearch echo back to Algolia when extra settings are enabled 
 
 ## 3.18.1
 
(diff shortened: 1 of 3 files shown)


cammonro:
Codacy platform failure is a false positive.

damcou (approved):
LGTM 🚀

PR #1977, opened 2026-08-05
MAGE-1674: Change conditions for Cron/ProcessQueue

damcou:
This PR contains (for Cron/ProcessQueue.php):

- Removed the check linked to outdated indexer
- Replaced it with the same condition but for queue enablement
- Removed usage of monolithic `ConfigHelper`
- Added Unit tests for the class


cammonro (approved):
Nice! Let's ship it. 🚀

PR #1978, opened 2026-08-05
MAGE-1676 Fix Closure serialization error on the queue job view pages

cammonro:
## Summary

Opening a job detail page from the Indexing Queue or Queue Archive grids could fail with:

```
Uncaught Exception: Serialization of 'Closure' is not allowed
```

Both `Controller\Adminhtml\Queue\AbstractAction::initJob()` and its Queue Archive counterpart loaded the entity and then stashed the **fully hydrated model** in the backend session (`$this->backendSession->setData('current_job', $model)`) so the view block could pick it up. `Model\Job` carries an injected `ObjectManagerInterface` for its handler dispatch, so when Magento serializes the session at `session_write_close()`, it walks into a closure and fatals.

The fix removes the session round trip entirely. Each view block now loads the entity itself from the `id` request parameter, which is the only thing that ever needed to cross the request boundary:

- `Block\Adminhtml\Job\View` loads via `Model\JobFactory` + `Model\ResourceModel\Job`. (No repository exists at this time)
- `Block\Adminhtml\QueueArchive\View` loads via `Api\QueueArchiveRepositoryInterface`.
- Both cache the loaded entity in a local property, so repeated `getCurrentJob()` calls in the templates still hit the database once.
- `SessionManagerInterface` is dropped from both controllers and both blocks.

Also included in this PR:

- Documented the underlying invariant in `doc/ARCHITECTURE.md` (Architectural Invariants and Anti-Patterns) so this does not get reintroduced: no Magento model instance goes into the backend session, persist a scalar id and reload the entity where it is needed.
- Fixed `$queue->getProcssedAt()` to `getProcessedAt()` in the Queue Archive view template. The typo left the **Processed At** field silently blank on that page.
- Modernized both controllers and blocks to constructor property promotion, native return types, and removed the PHPDoc blocks that only restated the type declarations.
- Corrected the "Informations" label to "Information" in both view templates.
- Added `CHANGELOG.md` entries for 3.19.0, including the constructor signature changes under breaking changes.

Thanks to @angelvilaplana for the original fix in #1939 🙌 , which covered the Indexing Queue view. This PR keeps that authorship and extends the same treatment to the Queue Archive view.

**Breaking changes**

Constructor signatures changed. Anything extending these classes or instantiating them directly needs updating:

| Class | Change |
| --- | --- |
| `Controller\Adminhtml\Queue\AbstractAction` | no longer accepts `SessionManagerInterface` |
| `Controller\Adminhtml\QueueArchive\AbstractAction` | no longer accepts `SessionManagerInterface` |
| `Block\Adminhtml\Job\View` | requires `Model\JobFactory` and `Model\ResourceModel\Job` in place of `SessionManagerInterface` |
| `Block\Adminhtml\QueueArchive\View` | requires `Api\QueueArchiveRepositoryInterface` in place of `SessionManagerInterface` |

`Block\Adminhtml\QueueArchive\View::getCurrentJob()` now returns `QueueArchiveInterface` and throws `NoSuchEntityException` for an unknown id, rather than returning `null` from the session.

## Result

New unit tests cover the load path for both blocks:

- `Test/Unit/Block/Adminhtml/Job/ViewTest.php`
- `Test/Unit/Block/Adminhtml/QueueArchive/ViewTest.php`

They assert that `getCurrentJob()` loads the entity using the `id` request parameter and that the result is memoized so a second call does not re-query.

image

```bash
vendor/bin/phpunit -c dev/tests/unit/phpunit.xml.dist \
  vendor/algolia/algoliasearch-magento-2/Test/Unit/Block/Adminhtml
```

Manual verification in admin:

1. Stores > Configuration > Algolia Search: enable the indexing queue.
2. Trigger a reindex so jobs land in `algoliasearch_queue`.
3. Algolia Search > Indexing Queue, click into a job. The detail page renders instead of an error.
4. Algolia Search > Queue Archive, click into an archived job. The detail page renders and **Processed At** now shows a timestamp.
5. Confirm no "Serialization of 'Closure' is not allowed" entries appear in `var/log/`.

image

image

### DIFF
--- a/.claude/reference/testing/core.md
+++ b/.claude/reference/testing/core.md
@@ -1,6 +1,6 @@
 #Core Testing Guidelines
 
-## Informations
+## Information
 - The files we are testing are located in a Magento2 module
 - The test files are located in the Test/Unit directory, create all the tests in this parent's directory
 - Follow the same structure as in the module (ex: tests for the Service/Product/RecordBuilder.php should be located in the Test/Unit/Service/Product directory)
--- a/Block/Adminhtml/Job/View.php
+++ b/Block/Adminhtml/Job/View.php
@@ -2,23 +2,29 @@
 
 namespace Algolia\AlgoliaSearch\Block\Adminhtml\Job;
 
+use Algolia\AlgoliaSearch\Model\JobFactory;
+use Algolia\AlgoliaSearch\Model\ResourceModel\Job as JobResource;
 use Magento\Backend\Block\Widget\Button;
-use Magento\Framework\Session\SessionManagerInterface;
 use Magento\Framework\View\Element\Template;
 use Magento\Framework\View\Element\Template\Context;
 
 class View extends Template
 {
-    /** @var SessionManagerInterface */
-    protected $backendSession;
+    /** @var JobResource */
+    protected $jobResource;
+
+    /** @var JobFactory */
+    protected $jobFactory;
 
     public function __construct(
-        Context          $context,
-        SessionManagerInterface   $backendSession,
+        Context       $context,
+        JobResource   $jobResource,
+        JobFactory    $jobFactory,
         array $data = []
     ) {
         parent::__construct($context, $data);
-        $this->backendSession = $backendSession;
+        $this->jobResource = $jobResource;
+        $this->jobFactory = $jobFactory;
     }
 
     /**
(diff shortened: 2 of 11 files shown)


damcou (approved):
LGTM, tests are passing ✅

PR #1989, opened 2026-08-18
MAGE-1670: Update IndexSettingsComparator class with reconcileKeys logic

damcou:
This PR contains:
- Add a `reconcileKeys` method to the `IndexSettingsComparator` class which ensures that the comparison doesn't find "false differences" whenever a getSettings return an empty array for attributes such as `customRanking` or `unretrievableAttributes`
- Updated unit tests as a result

### DIFF
--- a/Service/Index/Settings/IndexSettingsComparator.php
+++ b/Service/Index/Settings/IndexSettingsComparator.php
@@ -28,9 +28,33 @@ class IndexSettingsComparator
             $algoliaSettings = $this->connector->getSettings($indexOptions);
         }
 
-        $algoliaSettings = array_intersect_key($algoliaSettings, $indexSettings);
+        [$algoliaSettings, $indexSettings] = $this->reconcileKeys($algoliaSettings, $indexSettings);
 
         return $this->getSettingsHash($indexSettings) === $this->getSettingsHash($algoliaSettings);
+    }
+
+    private function reconcileKeys(array $remote, array $local): array
+    {
+        // Existing behaviour: settings Algolia manages but the extension does not
+        // are not differences.
+        $remote = array_intersect_key($remote, $local);
+
+        // Algolia does not round-trip empty list settings. `customRanking` is
+        // omitted from getSettings whenever it is empty, and
+        // `unretrievableAttributes` is omitted until it has been set at least
+        // once. For these keys "absent remotely" and "empty locally" describe the
+        // same index state, so they must compare equal.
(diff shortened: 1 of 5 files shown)


Service/Index/Settings/IndexSettingsComparator.php
```diff
@@ -28,11 +28,35 @@ public function matches(
             $algoliaSettings = $this->connector->getSettings($indexOptions);
         }
 
-        $algoliaSettings = array_intersect_key($algoliaSettings, $indexSettings);
+        [$algoliaSettings, $indexSettings] = $this->reconcileKeys($algoliaSettings, $indexSettings);
 
         return $this->getSettingsHash($indexSettings) === $this->getSettingsHash($algoliaSettings);
     }
 
+    private function reconcileKeys(array $remote, array $local): array
```
cammonro: Let's make this `protected` for extension modification. My original proposal was too restrictive here.

Service/Index/Settings/IndexSettingsComparator.php
```diff
@@ -28,11 +28,35 @@ public function matches(
...
+        // omitted from getSettings whenever it is empty, and
+        // `unretrievableAttributes` is omitted until it has been set at least
+        // once. For these keys "absent remotely" and "empty locally" describe the
+        // same index state, so they must compare equal.
+        foreach ($local as $key => $value) {
+            if ($value === []) {
+                if (!array_key_exists($key, $remote)) {
+                    unset($local[$key]);
+                } else if ($remote[$key] === null)  {
+                    $local[$key] = null;
```
cammonro, line 52: Would like to understand this use case more. Do we know if this triggers a no-op? What happens if we unset the key altogether?

cammonro (commented):
Looking good! But have a question about the `null` scenario you've discovered.

Service/Index/Settings/IndexSettingsComparator.php
damcou: Reply on JIRA

cammonro: @damcou I get it now! Thanks for explaining. The logic is sound and the hash comparator can now work as expected. Our unit tests lock the contract. Excellent work. 😀

cammonro (approved):
Solid! Let's ship it. 🚀

cammonro:
Can we add an entry to `CHANGELOG.md` while we're at it?

PR #1990, opened 2026-08-25
MAGE-1669: Updated `IndexSettingsHandler::setSettings()`

damcou:
This PR contains:
- Updated `IndexSettingsHandler::setSettings()`: 
  - We now use multiple `IndexSettingsComparator::matches()` for settings both forwarded and non-forwarded to replicas to ensure no no-op operations are sent to the dashboard. (`getSettings()` is still called once though)
  - Removed `waitLastTask` occurrences in favor of `collectTaskIdToWaitFor` ones to ensure waiting at the end of the process.
- As a result, the `collectTaskIdToWaitFor` occurrences of the `ProductHelper::setSettings() `and `IndexConfigurator` methods are removed (task collection is now completely managed by the `IndexSettingsHandler` itself)
- Updated `IndexSettingsHandler` unit tests.   
  - Removed state machine logic which was there to ensure that the removed `waitForLastTask()` logic was properly working 
  - Updated all tests to ensure proper number of `matches` and `collectTaskIdToWaitFor` triggered for each case.

### DIFF
--- a/Service/Index/Settings/IndexSettingsHandler.php
+++ b/Service/Index/Settings/IndexSettingsHandler.php
@@ -63,6 +63,7 @@ class IndexSettingsHandler
                 $indexSettings,
                 false
             );
+            $this->connector->collectTaskIdToWaitFor($indexOptions);
 
             return true;
         }
@@ -73,21 +74,27 @@ class IndexSettingsHandler
 
         // FORWARDED: $settings without excluded attributes
         if ($forward) {
-            $this->connector->setSettings(
-                $indexOptions,
-                $forward,
-                true
-            );
-            $this->connector->waitLastTask($indexOptions->getStoreId());
+            if (! $this->indexSettingsComparator->matches($indexOptions, $forward, $remoteSettings)) {
+                $this->connector->setSettings(
+                    $indexOptions,
+                    $forward,
+                    true
+                );
+                $this->connector->collectTaskIdToWaitFor($indexOptions);
+//                $this->connector->waitLastTask($indexOptions->getStoreId());
+            }
         }
 
         // NOT FORWARDED: array containing excluded attributes only
         if ($noForward) {
-            $this->connector->setSettings(
-                $indexOptions,
-                $noForward,
-                false
-            );
+            if (! $this->indexSettingsComparator->matches($indexOptions, $noForward, $remoteSettings)) {
+                $this->connector->setSettings(
+                    $indexOptions,
+                    $noForward,
+                    false
+                );
+                $this->connector->collectTaskIdToWaitFor($indexOptions);
+            }
         }
 
         return true;
--- a/Service/Index/Settings/IndexSettingsHandler.php
+++ b/Service/Index/Settings/IndexSettingsHandler.php
@@ -81,7 +81,6 @@ class IndexSettingsHandler
                     true
                 );
                 $this->connector->collectTaskIdToWaitFor($indexOptions);
-//                $this->connector->waitLastTask($indexOptions->getStoreId());
             }
         }
 
--- a/Service/Index/Settings/IndexSettingsHandler.php
+++ b/Service/Index/Settings/IndexSettingsHandler.php
@@ -73,27 +73,23 @@ class IndexSettingsHandler
         [$forward, $noForward] = $this->splitSettings($indexSettings);
 
         // FORWARDED: $settings without excluded attributes
-        if ($forward) {
-            if (! $this->indexSettingsComparator->matches($indexOptions, $forward, $remoteSettings)) {
-                $this->connector->setSettings(
-                    $indexOptions,
-                    $forward,
-                    true
-                );
-                $this->connector->collectTaskIdToWaitFor($indexOptions);
-            }
+        if ($forward && !$this->indexSettingsComparator->matches($indexOptions, $forward, $remoteSettings)) {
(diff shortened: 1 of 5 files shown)


Service/Index/Settings/IndexSettingsHandler.php
```diff
@@ -73,21 +74,26 @@ public function setSettings(IndexOptionsInterface $indexOptions, array $indexSet
 
         // FORWARDED: $settings without excluded attributes
         if ($forward) {
-            $this->connector->setSettings(
-                $indexOptions,
-                $forward,
-                true
-            );
-            $this->connector->waitLastTask($indexOptions->getStoreId());
+            if (! $this->indexSettingsComparator->matches($indexOptions, $forward, $remoteSettings)) {
```
cammonro: Just a suggestion but we could combine these nested ifs into a single statement. I think that's worth shedding a pair of brackets for readability.

Service/Index/Settings/IndexSettingsHandler.php
```diff
@@ -73,21 +74,26 @@ public function setSettings(IndexOptionsInterface $indexOptions, array $indexSet
...
         }
 
         // NOT FORWARDED: array containing excluded attributes only
         if ($noForward) {
-            $this->connector->setSettings(
-                $indexOptions,
-                $noForward,
-                false
-            );
+            if (! $this->indexSettingsComparator->matches($indexOptions, $noForward, $remoteSettings)) {
```
cammonro: These if statements could also be merged - same as above.

Test/Unit/Service/Index/Settings/IndexSettingsHandlerTest.php
```diff
@@ -3,6 +3,7 @@
 namespace Algolia\AlgoliaSearch\Test\Unit\Service\Index\Settings;
 
 use Algolia\AlgoliaSearch\Api\Data\IndexOptionsInterface;
+use Algolia\AlgoliaSearch\Controller\View\Index;
```
cammonro: Looks like a stray import?

Test/Unit/Service/Index/Settings/IndexSettingsHandlerTest.php
```diff
@@ -162,13 +123,18 @@ public function testSetSettingsWithForwardingEnabledOnlyExcludedSettings(): void
                 false
             );
 
+        // IndexSettingsComparator::matches() is called 3 times:
```
cammonro: This comment seems incorrect? Only 2 times?

Test/Unit/Service/Index/Settings/IndexSettingsHandlerTest.php
```diff
@@ -188,13 +154,18 @@ public function testSetSettingsWithForwardingEnabledOnlyForwardableSettings(): v
                 true
             );
 
+        // IndexSettingsComparator::matches() is called 3 times:
+        // - once for the full payload
```
cammonro, line 152: Another misleading comment...

Test/Unit/Helper/Entity/ProductHelperTest.php
cammonro: This test name no longer applies - we're actually asserting the opposite.

Test/Unit/Service/Index/Settings/IndexSettingsHandlerTest.php
```diff
@@ -44,9 +45,6 @@ protected function setUp(): void
         $this->logger = $this->createMock(AlgoliaLogger::class);
         $this->indexOptions = $this->createMock(IndexOptionsInterface::class);
 
-        // Configure the mock to use our state machine
```
cammonro, line 47: I guess we don't need this at all anymore do we? 🤔 Something to keep in mind when merging and resolving conflicts in the [PHPUnit 12 work](https://github.com/algolia/algoliasearch-magento-2/pull/1988#discussion_r3814562304) that refactored this...

cammonro (changes requested):
PR looks great to me. Great job! 

Had some minor nits but my biggest concern is I'm not sure we're testing all 4 scenarios in our unit tests? Namely what *changed* (not whether settings are empty) 

1. both halves changed
2. forward-only changed
3. no-forward-only changed
4. neither changed

To do this we'd have to mock the comparator to return different values.

Test/Unit/Service/Index/Settings/IndexSettingsHandlerTest.php
damcou: Nope, from my understanding, the state machine was there to handle the different scenarios with the waitLastTask() occurrences, which we don't use anymore.

Test/Unit/Helper/Entity/ProductHelperTest.php
```diff
@@ -119,11 +119,11 @@ public function testSkipsAlgoliaConnectorUpdateWhenSettingsUnchanged(): void
         $this->productHelper->setSettings($this->indexOptions, $this->indexTmpOptions, $this->storeId);
     }
 
-    public function testWaitsForLastTaskWhenSettingsChanged(): void
+    public function testWDoesNotCollectTaskIdWhenSettingsChanged(): void
```
cammonro: Small typo

cammonro (approved):
Found one small typo but this looks great. Love the data provider for simulating the change state!

PR #1991, opened 2026-08-27
MAGE-1670: Update IndexSettingsComparator class with reconcileKeys logic (3.18.2)

damcou:
This PR backports changes from https://github.com/algolia/algoliasearch-magento-2/pull/1989 for 3.18.2

### DIFF
--- a/Service/IndexSettingsComparator.php
+++ b/Service/IndexSettingsComparator.php
@@ -19,9 +19,33 @@ class IndexSettingsComparator
     public function matches(IndexOptionsInterface $indexOptions, array $indexSettings): bool
     {
         $algoliaSettings = $this->connector->getSettings($indexOptions);
-        $algoliaSettings = array_intersect_key($algoliaSettings, $indexSettings);
+        [$algoliaSettings, $indexSettings] = $this->reconcileKeys($algoliaSettings, $indexSettings);
 
         return $this->getSettingsHash($indexSettings) === $this->getSettingsHash($algoliaSettings);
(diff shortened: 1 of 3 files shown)


Service/IndexSettingsComparator.php
```diff
@@ -19,11 +19,35 @@ public function __construct(
     public function matches(IndexOptionsInterface $indexOptions, array $indexSettings): bool
```
cammonro, line 19: @damcou We'll need to change to change this API surface (add the optional third arg introduced in 3.19) to support both backports - will that be a separate PR?

Test/Unit/Service/IndexSettingsComparatorTest.php
```diff
@@ -209,4 +209,154 @@ public function testInvalidJson(): void
         // Must be different
         $this->assertFalse($this->indexSettingsComparator->matches($this->indexOptions, [INF]));
     }
+
+    // ── reconcileKeys() ──
+
+    /**
+     * @dataProvider reconcileKeysProvider
+     */
+    public function testReconcileKeys(array $remote, array $local, array $expectedRemote, array $expectedLocal): void
```
cammonro, line 218: Good call here to ship only the basic test shape and avoid the PHPUnit 12 overhead in 3.19.

cammonro (approved):
Looks good to me! Just one question on the `matches` API but that depends on your decision for the forward/noforward split backport. Not a blocker!

PR #1997, opened 2026-09-03
3.19.1 => main

damcou:
### Updates
- Updated `IndexSettingsHandler::setSettings`:
    - It now fetches the remote settings once, runs preservation, then threads that same payload into the comparator.
    - It now use multiple `IndexSettingsComparator::matches()` for settings both forwarded and non-forwarded to replicas to ensure no no-op operations are sent to the dashboard.
    - Removed `waitLastTask` occurrences in favor of `collectTaskIdToWaitFor` ones to ensure waiting at the end of the process.
    - Updated unit tests related to those classes.
- Updated all unit and integration tests to be compatible with PHPUnit 12.

### Bug fixes
- Updated `IndexSettingsComparator` to include a protected method `reconcileKeys` to ensure empty attributes such as `customRanking` or `unretrievableAttributes` are handled properly.


cammonro (approved):
🚀

PR #1998, opened 2026-09-14
MAGE-1704: Introduced Asynchronous Configuration save feature

damcou:
This PR contains:
- Introduced "Asynchronous configuration save" feature for multi-stores setup experiencing very long saving time.
- Updated related unit tests
- Updated version files and changelog


Helper/ConfigHelper.php
```diff
@@ -1088,6 +1089,15 @@ public function isAutoPriceIndexingEnabled(?int $storeId = null): bool
         );
     }
 
+    public function isAsyncConfigSaveEnabled(?int $storeId = null): bool
```
cammonro, line 1092: Good call to feature flag this especially for a patch - ideally we should start pushing this into a separate helper (`AdvancedConfigHelper`?) rather than the old monolith but that's a minor release change.

Model/IndicesConfigurator.php
```diff
@@ -55,7 +55,8 @@ public function __construct(
     public function saveConfigurationToAlgolia(
         int $storeId,
         bool $useTmpIndex = false,
-        array $filteredEntities = [])
+        array $filteredEntities = [],
+        bool $skipWait = false)
```
cammonro, line 59: We're touching an API surface here but we don't declare @api on this class and the default covers us. Perhaps in the future we can switch to an options object?

Model/IndicesConfigurator.php
```diff
@@ -98,6 +99,11 @@ public function saveConfigurationToAlgolia(
 
         $this->setExtraSettings($storeId, $useTmpIndex, $filteredEntities);
 
+        if ($this->configHelper->isAsyncConfigSaveEnabled($storeId) && $skipWait) {
+            $this->logger->stop($logEventName, true, true);
```
cammonro, line 104: The danger of dropping a `logger->stop` is real, to simplify the code should we move a single call above these conditionals? Especially since the `$logEventName` is for saving not waiting?

Model/IndicesConfigurator.php
```diff
@@ -98,6 +99,11 @@ public function saveConfigurationToAlgolia(
 
         $this->setExtraSettings($storeId, $useTmpIndex, $filteredEntities);
 
+        if ($this->configHelper->isAsyncConfigSaveEnabled($storeId) && $skipWait) {
```
cammonro, line 102: If we hit drift issues we'll need a diagnostic in `algolia.log`  - can we add something like:

```
$this->logger->log('Skipping wait for task completion (asynchronous configuration save enabled).'
```

cammonro (approved):
This is a solid immediate fix for the 504 gateway timeouts on long save operations. Feature flagging is the right approach so it's totally opt-in. Added some comments (I think a diagnostic log here is important) but I am approving to unblock.

Let's consider this a short term fix because the potential for settings drift remains and can give false confidence. Will follow up in Jira with some additional thoughts on possible follow up items.

Model/IndicesConfigurator.php
damcou: Exactly, I tried to make sure that the minimum of occurrences would need a change with the default set to `false` . An option object is a good idea, but would require quite a lot of changes indeed.

Model/IndicesConfigurator.php
damcou: Reply on JIRA

